PR #70, opened 2014-03-23
WIP: Python buffer (PEP 3118) implementation

jakebolewski:
This is very much a work in progress branch to explore replacing PyCall's numpy dependency with Python Buffers / memory views (https://github.com/stevengj/PyCall.jl/issues/38, and https://github.com/JuliaLang/IJulia.jl/issues/49).  Link to PEP proposal: http://legacy.python.org/dev/peps/pep-3118/

Note: to get something working I've ignored for the time being important things like proper reference counting, and python initialization.  I have also removed including the numpy code to make testing easier.

I want to get some feedback and to bring up some questions / problems I can foresee in implementing the rest of this.  This pull request currently only works for Python v. > 3.2 as the layout of the Buffer structure changed in 3.3.  Python initialization is deferred until runtime, so is it best to defer PyBuffer type creation until runtime?.  Now with faster startup times, is there a reason not to initialize libpython when first loading the module?  Another wrinkle is that Buffer support has only been backported to 2.7.  The other option is the keep the pointer opaque and do a copy of the data upon Buffer creation.

Currently the buffer implementation matches the functionality of the numpy interface (most of the code is largely the same  with few modifications) under the assumption that the buffers are contiguous. Python's buffer interface allows for non-contiguous buffers so this will have to be implemented.  Sould there be a distinction between the two PyArray wrapper types?  There have been recent discussions about Julia's Array type hierarchy reflecting the underlying data layout.

The biggest advantage of utilizing python's buffer interface is that it supports heterogeneous structure types.  This will allow for conversions between numpy's record arrays and julia array's of immutable types.  Parsing the buffer format specification is going to take a bit more work.  The main question I have is how to best generate the types for these buffers.  For instance upon being passed a numpy record array specifying some structure, do we create a requisite type on the fly or look to the global environment and try to match an existing type?  I would be interested in any thoughts anyone might have as for how this would work.

The last thing to do would be to implement the buffer interface for Julia's array types to support Python -> Julia interfaces.  What is the best way to do this?

### DIFF
--- a/test/test_buffer.jl
+++ b/test/test_buffer.jl
@@ -113,7 +113,7 @@ a[1] = 3
 a[2] = 2
 a[3] = 1
 @test sprint() do io; show(io, npa); end == "PyObject array([3, 2, 1])"
-aa = np.frombuffer(a)
+aa = np.asarray(a)
 @test np.shape(aa) == size(a)
 
 npa = np.ones((10,10), dtype="float32")
@@ -128,9 +128,8 @@ a = PyArray(npa)
 @test sprint() do io; show(io, a); end == sprint() do io
 				              show(io, ones(Float32, (10,10)))
 				      	  end
-#TODO:
-#aa = np.frombuffer(a)
-#@test np.shape(aa) == size(a)
+aa = np.asarray(a)
+@test np.shape(aa) == size(a)
 
 npa = np.ones((10,10,20), dtype="float32")
 a = PyArray(npa)
@@ -140,9 +139,9 @@ a = PyArray(npa)
 @test size(similar(a)) == size(a)
 @test eltype(similar(a)) == eltype(a)
 @test summary(a) == "10x10x20 Float32 PyArray"
-#TODO:
-#aa = np.frombuffer(a)
-#@test np.shape(aa) == size(a)
+
+aa = np.asarray(a)
+@test np.shape(aa) == size(a)
 
 # Test fail because of show uses unimplemented subarray methods 
 #@test sprint() do io; show(io, a); end == sprint() do io


src/buffer.jl
```diff
@@ -0,0 +1,432 @@
+# Support for python buffer protocol buffers
+
+# for versions > 3.2 (TODO: obj field not in struct for python versions <= 3.2)
```
stevengj: @jakebolewski, why do you say that the `obj` field is not in the struct for earlier versions?

I just looked in the Python 2.7.6 `Include/object.h` header file, and `Py_buffer` is defined as
```c
/* Py3k buffer interface */
typedef struct bufferinfo {
    void *buf;
    PyObject *obj;        /* owned reference */
    Py_ssize_t len;
    Py_ssize_t itemsize;  /* This is Py_ssize_t so it can be                    
                             pointed to by strides in simple case.*/
    int readonly;
    int ndim;
    char *format;
    Py_ssize_t *shape;
    Py_ssize_t *strides;
    Py_ssize_t *suboffsets;
    Py_ssize_t smalltable[2];  /* static store for shape and strides of         
                                  mono-dimensional buffers. */
    void *internal;
} Py_buffer;
```
which seems identical except for the `smalltable` field (which is missing in Python 3.4).  However, we can just add some padding to the structure for compatibility, since we don't actually access those fields ourselves.

jakebolewski: @stevengj I had a look, that comment is just wrong.  Perhaps I wrote that initially based on the documentation ->
[Python 2.7](https://docs.python.org/2/c-api/buffer.html#bufferobjects) vs [Python 3.4](https://docs.python.org/3.4/c-api/buffer.html#bufferobjects) where the obj field is documentated in the latter.

It was a bit speculative at the time, but I really should finish this PR.  Bidirectional zero copy buffers would be a really nice feature.

src/buffer.jl
```diff
@@ -0,0 +1,426 @@
...
+end
+
+release!(b::PyBuffer) = begin
+    if b.buf.obj != C_NULL
+        ccall((@pysym :PyBuffer_Release), Void, (Ptr{PyBuffer},), &b)
+    end
+end
+
+Base.ndims(b::PyBuffer)  = int(b.buf.ndim)
+Base.length(b::PyBuffer) = b.buf.ndim >= 1 ? div(b.buf.len, b.buf.itemsize) : 0
```
stevengj: Why is the length zero for `ndim == 0` ... aren't zero-dimensional arrays normally length 1?   Why do you have this check here?

src/buffer.jl
```diff
@@ -0,0 +1,426 @@
...
+Base.sizeof(b::PyBuffer) = b.buf.len
+
+Base.size(b::PyBuffer) = begin
+    if b.buf.ndim == 0
+        return (0,)
```
stevengj: as above

src/buffer.jl
stevengj: Also, is `shape` is NULL then `itemsize` should be disregarded and assumed to be `1`, according to the docs.

src/buffer.jl
```diff
@@ -0,0 +1,426 @@
...
+    if b.buf.ndim == 1
+        return (div(b.buf.len, b.buf.itemsize),)
+    end
+    @assert b.buf.shape != C_NULL
+    return tuple(Int[unsafe_load(b.buf.shape, i) for i=1:b.buf.ndim]...) 
+end 
+
+Base.strides(b::PyBuffer) = begin
+    if b.buf.ndim == 0 || b.buf.ndim == 1
+        return (1,)
```
stevengj: This seems wrong: 1-dimensional buffers are not necessarily stride-1, according to the docs.

jakebolewski: Yes this logic is incomplete, `strides` can also be negative.

src/buffer.jl
```diff
@@ -0,0 +1,426 @@
...
+end
+
+pyfmt(b::PyBuffer) = b.buf.format == C_NULL ? bytestring("") : bytestring(b.buf.format)
+
+sizeof_pyfmt(fmt::ByteString) = ccall((@pysym :PyBuffer_SizeFromFormat), Cint,
+                                      (Ptr{Cchar},), &fmt)
+
+#########################################################################
+
+pygetbuffer(o::PyObject, flags::Cint) = begin
```
stevengj: Why not just a `PyBuffer(o, flags)` constructor?

stevengj:
@jakebolewski, I merged a modified subset of this PR in order to fix stevengj/PyPlot.jl#118 — basically, my `PyObject(::IO)` wrappers weren't working because they need to use the buffer interface to get access to the raw bytes in order to implement `write`.

It would be good to have an updated PR which adds the other stuff I omitted, in particular your new NumPy-free `PyArray` (probably in a separate `pyarray.jl` file).

PallHaraldsson:
"I merged a modified subset of this PR in order to fix JuliaPy/PyPlot.jl#118"

[Because of this and old comments] I only scanned this issue, and should it still be open?

Is it my correct understanding, that even without, PyCall is pretty good (at least the multidimensional aspects)? This issue would just be icing-on-the cake? Not really faster, just loosing a dependency? [I see (unrelated) bugs fixed all the time, in PyCall, I think, and README seems to support, that it mostly works, rather than the opposite.]

In the source code:

function NpyArray{T<:NPY_TYPES}(a::StridedArray{T}, revdims::Bool) #not needing default =false (as I think not exported)
[..]

PyReverseDims{T<:NPY_TYPES}(a::StridedArray{T}) = NpyArray(a, true)
PyReverseDims(a::BitArray) = PyReverseDims(Array(a))

[doc]
PyReverseDims(a::AbstractArray) #this stray line is puzzling to me.. I thought a function body needed, can't do similar myself..

[PyReverseDims of course comes at an unavoidable performance cost], unless you chose not to flip, that is often ok; unclear from my reading of the code, if there's a cost for 1D arrays, and someone uses PyReverseDims out of habit.]

PR #85, opened 2014-08-01
RFC: Reverse Array dims in no-copy conversion from Julia to Numpy

kmsquire:
I've been working with image data and `PyPlot.imshow`, and finding that I have to munge the data more than I would like in order to get it to display:

``` julia
PyPlot.imshow(permutedims(img.data, (3,2,1))
```

Digging into this, I found that reversing the dims and strides in the `PyObject` constructor let me display the image (more) directly with `PyPlot.imshow(img.data)`.

Now it's likely that this change would cause issues with other arrays being passed to Numpy, so I don't expect this to be merged.  But I'm wondering if a version which reverses the dims/strides can be made available somehow?

### DIFF
--- a/src/numpy.jl
+++ b/src/numpy.jl
@@ -175,14 +175,20 @@ const npy_typestrs = Dict( "b1"=>Bool,
 #########################################################################
 # no-copy conversion of Julia arrays to NumPy arrays.
 
-function PyObject{T<:NPY_TYPES}(a::StridedArray{T})
+# In some cases, an array is already in row-major order (e.g., RGB
+# images) and it is useful to reverse the order of dimensions passed
+# to PyArray_New.  In these cases, set revdims=true.
+
+function PyObject{T<:NPY_TYPES}(a::StridedArray{T}, revdims::Bool=false)
     try
         @npyinitialize
+        size_a = revdims ? reverse(size(a)) : size(a)
+        strides_a = revdims ? reverse(strides(a)) : strides(a)
         p = @pycheck ccall(npy_api[:PyArray_New], PyPtr,
               (PyPtr,Cint,Ptr{Int},Cint, Ptr{Int},Ptr{T}, Cint,Cint,PyPtr),
               npy_api[:PyArray_Type],
-              ndims(a), Int[size(a)...], npy_type(T),
-              Int[strides(a)...] * sizeof(eltype(a)), a, sizeof(eltype(a)),
+              ndims(a), Int[size_a...], npy_type(T),
+              Int[strides_a...] * sizeof(eltype(a)), a, sizeof(eltype(a)),
               NPY_ARRAY_ALIGNED | NPY_ARRAY_WRITEABLE,
               C_NULL)
         return PyObject(p, a)


stevengj:
Hi @kmsquire, sorry I lost track of this.  This looks good to merge given a rebase and a test.

kmsquire:
Me too.  ;-)  Rebased.

kmsquire:
I'll add a test in a little while.

src/numpy.jl
```diff
@@ -175,14 +175,20 @@ const npy_typestrs = Dict( "b1"=>Bool,
 #########################################################################
 # no-copy conversion of Julia arrays to NumPy arrays.
 
-function PyObject{T<:NPY_TYPES}(a::StridedArray{T})
+# In some cases, an array is already in row-major order (e.g., RGB
```
stevengj: I would say rather than "Julia arrays are in column-major order, but in some cases it is useful to pass them to Python as row-major arrays simply by reversing the dimensions. For example, although NumPy works with both row-major and column-major data, some Python libraries like OpenCV seem to require row-major data (the default in NumPy).  In such cases, set `revdims=true`."

And then add a similar comment to the README.md file.

stevengj:
Ping, any chance of a test?  (The AppVeyor failure looks like an unrelated network problem.)

stevengj:
Probably just something like
```jl
if PyCall.npy_initialized
    let A = rand(2,3,4), o = PyObject(A), A2 = convert(PyAny, o)
        @test A == permutedims(A2, [3,2,1])
    end
end
```

src/numpy.jl
```diff
@@ -175,14 +175,20 @@ const npy_typestrs = Dict( "b1"=>Bool,
...
+# images) and it is useful to reverse the order of dimensions passed
+# to PyArray_New.  In these cases, set revdims=true.
+
+function PyObject{T<:NPY_TYPES}(a::StridedArray{T}, revdims::Bool=false)
     try
```
dmrd: This should not silently fall back on failure to array2py(a) if revdims=true.  It produces inconsistent behavior.  Is there some way array2py could take revdims as an extra arg as well?

stevengj: Good catch.   For now, it can throw an error if `revdims=true` but in the future `array2py` could take an extra arg.

dmrd:
Reproduction: https://gist.github.com/dmrd/b47c89c045279973258f997eeff1e863

Can we add BitArray--> Array{uint8} conversion as a default?  Numpy doesn't seem to do bit arrays.

stevengj:
@dmrd, a converter for `BitArray`, probably to `Array{Bool}`, is a good idea, but should be filed as a separate issue or PR.   The patch could probably be as simple as a line `PyObject(a::BitArray) = PyObject(Array{Bool}(a))` plus a test case.

stevengj:
Closing in favor of #265

kmsquire:
Thanks, @steveng, and sorry not to be more responsive.

stevengj:
No problem, thanks for doing most of the work on this!

PR #110, opened 2015-01-04
Use a bare module with no imports for ＠pyimport

simonster:
This avoids warnings about conflicting imports


simonster:
Hmm, I thought this would work since the `Module` constructor just calls `jl_f_new_module`, which [doesn't do much else](https://github.com/JuliaLang/julia/blob/master/src/module.c#L38-L44), and it does indeed suppress the warning, but it seems that there's code that expects the module to have a parent, and I'm not sure how to set that from Julia (yet).

simonster:
This is pretty ugly, but it looks like it works. A better long-term approach might be to extend the `Module` constructor to allow creation of `baremodule`s.

stevengj:
Thanks.

stevengj:
(Gotta fix that Julia 0.3 failure, but that seems unrelated to your patch.)

PR #118, opened 2015-03-04
Minor tweaks to badges and tracking master instructions

IainNZ:
I thought since PackageEvaluator is working nicely with PyCall, it'd be nice to show PyCall is working great on release.

My next PkgEval project is to start testing the various packages that depend on PyCall, which are mostly excluded right now.

I've also thrown in a tweak to the REQUIRE - PyCall is defacto 0.3 only anyway due to Dates, thought it was better to be explicit.


stevengj:
Thanks!

PR #124, opened 2015-03-17
Fix for JuliaLang/julia#10328

simonster:
More breakage on 0.4...


stevengj:
I'm getting tired of these continual spelling changes.

PR #134, opened 2015-04-24
Fix name collisions by explicitly qualifying Libdl

mweastwood:
(additionally removed some code for v0.3 compatibility because that
should be handled by Compat now)

Necessary since JuliaLang/julia#10951


stevengj:
Thanks!

PR #146, opened 2015-05-23
Fix ambiguity warning

yuyichao:
I believe @timholy's PR JuliaLang/julia#11194 makes julia (correctly) warns on the methods ambiguity. Switching the order of the definition solves the problem. I know there's discussions about defering the warning but IMHO it is good to put them in the least-ambigious order anyway...


yuyichao:
@stevengj Ping.

Just to clarify, this fixes a correct warning after a julia bug was fixed whereas the discovery inspired by this warning (https://github.com/JuliaLang/julia/issues/11407) is another julia bug.

PR #147, opened 2015-05-23
Create baremodule directly with sufficiently new Julia

simonster:
With JuliaLang/julia#9609 we can avoid the hack from #110


stevengj:
Okay, but Travis is segfaulting with Python 2.7 and Julia nightly....

(I'd like to get rid of this whole module thing entirely, but that relies on overloading `.` being supported.)

stevengj:
Okay, restarted Travis and it seems fine now.

PR #150, opened 2015-06-02
Fix deprecated ntuple signature

yuyichao:
There are more dlopen issues though... Will comment on the dlhandle PR.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -29,7 +29,7 @@ else
     const unsafe_convert = Base.convert
 end
 # This is not the exact version but should be closed enough
-if VERSION >= v"0.4.0-dev+4922"
+if v"0.4.0-dev+4922" <= VERSION < v"0.4.0-dev+5199"
     typealias HandleT Union(Libdl.DLHandle, Ptr{Void})
     hdl_ptr(hdl::Libdl.DLHandle) = hdl.ptr
 else
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -104,7 +104,7 @@ end
 
 function pydecref(o::PyObject)
     if initialized::Bool # don't decref after pyfinalize!
-        ccall((@pysym :Py_DecRef), Void, (PyPtr,), o.o)
+        ccall(py_decref, Void, (PyPtr,), o.o)
     end
     o.o = convert(PyPtr, C_NULL)
     o
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -209,9 +209,10 @@ if VERSION < v"0.4.0-dev+4319" # prior to 0.4 tuple-type changes
         if len != length(tt)
             throw(BoundsError())
         end
-        ntuple(len, i ->
-               convert(tt[i], PyObject(ccall((@pysym :PySequence_GetItem), PyPtr,
-                                             (PyPtr, Int), o, i-1))))
+        ntuple((i ->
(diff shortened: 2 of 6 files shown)


yuyichao:
https://github.com/JuliaLang/julia/pull/11486

yuyichao:
https://github.com/JuliaLang/julia/issues/11545

yuyichao:
OK, the second commit should workaround the Segfault.

I don't really like it (because of the inconsistency in dealing with symbols), and I guess you probably hate it more. However, this is the only way around it right now AFAICT.

yuyichao:
I'm convinced that it is quite hard to make the GC smart enough to handle this (See the linked issues in JuliaLang/julia#11545). So if JuliaLang/julia#11293 is to be kept, the solution in this PR is pretty much the only way out of it.

garrison:
@yuyichao It's looking like JuliaLang/julia#11293 is going to be reverted (my bad, completely).  Does this eliminate the need for the second commit here?

yuyichao:
@garrison Yes it would. (although it will not break it either)

yuyichao:
Update this PR for https://github.com/JuliaLang/julia/pull/11563
(basically replace the second commit with the revert of my previous commit)

yuyichao:
OK. the DLHandle PR is reverted now and this PR can be merged....

src/PyCall.jl
```diff
@@ -28,14 +28,6 @@ if VERSION >= v"0.4.0-dev+3710"
 else
     const unsafe_convert = Base.convert
 end
-# This is not the exact version but should be closed enough
-if VERSION >= v"0.4.0-dev+4922"
```
stevengj: It would be nicer to keep this for now, as `if  v"0.4.0-dev+XXX" >  VERSION >= v"0.4.0-dev+4922"` for the correct XXXX, so it will continue to work for people who have a 0.4 snapshot in the window where `DLHandle` was used.

src/conversions.jl
```diff
@@ -209,9 +209,10 @@ if VERSION < v"0.4.0-dev+4319" # prior to 0.4 tuple-type changes
         if len != length(tt)
             throw(BoundsError())
         end
-        ntuple(len, i ->
-               convert(tt[i], PyObject(ccall((@pysym :PySequence_GetItem), PyPtr,
-                                             (PyPtr, Int), o, i-1))))
+        ntuple((i ->
```
stevengj: why the extra parens?

src/PyCall.jl
yuyichao: ............... In this case should I add back my other commit (which I'm keeping on a backup branch right now) so that those people don't get a segfault on exit?

src/conversions.jl
yuyichao: Ahh. I was a little bit paranoid after getting many errors on `f(@compat sth, sth-else)` not working. I can remove this....

yuyichao:
OK. I've included the segfault fix (which doesn't conflict with anything else but make the DLHandle version not segfault on exit) and add a upper version check for `DLHandle`. Again, this is not the percise version but should be closed enough because I remember I compiled this version very soon after the revert...

yuyichao:
And sorry somehow I almost forgot about this pending PR. (which is why I didn't fix this two days earlier....)

stevengj:
Thanks!

PR #155, opened 2015-06-17
Use v0.4 Timer API

darwindarak:
This should address stevengj/PyPlot.jl#136 as well.

I don't know of a clean way of making `Dict{Symbol,Timer}()` compatible with the new timer API, so an if-statement will have to do for now.  `@compat` seems to only work on Timer type declarations, not the type itself.

### DIFF
--- a/src/gui.jl
+++ b/src/gui.jl
@@ -144,7 +144,11 @@ function Tk_eventloop(sec::Real=50e-3)
     install_doevent(doevent, sec)
 end
 # cache running event loops (so that we don't start any more than once)
-const eventloops = Dict{Symbol,Timer}()
+if VERSION < v"0.4.0-dev+5322"
+    const eventloops = Dict{Symbol,Compat.Timer2}()
+else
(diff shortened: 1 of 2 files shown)


kmsquire:
It would be good to add a minimum dependency on Compat 0.4.6 in REQUIRE.

darwindarak:
Right, that's kind of important.  Done!

src/gui.jl
```diff
@@ -62,8 +62,7 @@ end
 
 # call doevent(status) every sec seconds
 function install_doevent(doevent::Function, sec::Real)
-    timeout = Base.Timer(doevent)
-    Base.start_timer(timeout,sec,sec)
+    timeout = Base.Timer(doevent,sec,sec)
```
stevengj: This will break Julia 0.3, which only has `Timer(cb::Function)`.

stevengj: Oh, never mind, I see that Compat fixes this.

PR #156, opened 2015-06-17
Updated timer methods to 0.4-dev

paulmthompson:
Timer methods were recently changed so that start_timer call isn't used:

https://github.com/JuliaLang/julia/pull/11669

I actually don't know how to find the exact version where that change took place, but I think that it is "0.4.0-dev+5390"


paulmthompson:
looks like someone got to this as I was uploading and I missed it. Sorry for the duplicate!

PR #166, opened 2015-07-10
Added ability to reuse local dictionary across calls to pyeval

malmaud:
This makes it possible to call ``pyeval`` with class definitions and then later refer to those class definitions in a subsequent call to ``pyeval``.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -482,18 +482,21 @@ end
 
 #########################################################################
 
-const Py_eval_input = 258 # from Python.h
+const Py_single_input = 256  # from Python.h
+const Py_file_input = 257
+const Py_eval_input = 258
+
 const pyeval_fname = bytestring("PyCall.jl") # filename for pyeval
 
 # evaluate a python string, returning PyObject, given a dictionary
 # (string/symbol => value) of local variables to use in the expression
-function pyeval_(s::AbstractString, locals::PyDict)
+function pyeval_(s::AbstractString, locals::PyDict, input_type)
     sb = bytestring(s) # use temp var to prevent gc before we are done with o
     sigatomic_begin()
     try
         o = PyObject(@pycheckn ccall((@pysym :Py_CompileString), PyPtr,
                                      (Ptr{Uint8}, Ptr{Uint8}, Cint),
-                                     sb, pyeval_fname, Py_eval_input))
+                                     sb, pyeval_fname, input_type))
         main = @pycheckni ccall((@pysym :PyImport_AddModule),
                                 PyPtr, (Ptr{Uint8},),
                                 bytestring("__main__"))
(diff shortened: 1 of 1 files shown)


stevengj:
Looks good, thanks.

I've been thinking that it might be nice to have a
```py"""
....
"""
```
string macro that corresponds to `pyeval` with `Py_file_input`.  Maybe with `$...` interpolation in the form of generated local variables.

malmaud:
Similar to an IPython %%julia magic cell, but in reverse? That would be quite cool.

utensil:
I haven't found any document or comment to explain how to use this functionality. Am I missing something?

stevengj:
@utensil, call `pyeval` with `Py_file_input` as the fourth parameter.

utensil:
@stevengj Thanks for the explanation. I did see the intent expressed in the function signature, and I'm aware of [Py_eval_input/Py_single_input/Py_file_input](https://docs.python.org/2/c-api/veryhigh.html). 

"reuse local dictionary across calls to pyeval" as how it literally sounds is what I needed. But this title, the description and the implementation, all together didn't make any sense to me, so I looking for the actual usage as sample code to verify, or some further explanations.

utensil:
I find some other way to implement what I need, but when combining with these flags like `Py_file_input` and `Py_single_input`, it gives different results:

image

utensil:
The example above might be related to the library I'm using, so I created a minimal reproduction of this.

image


```julia
vara = 1;varb = 2;
local_vars = PyDict(Dict(:vara=>vara, :varb=>varb))
PyCall.pyeval_("vara + varb", local_vars, PyCall.Py_eval_input) # or single or file
```

It seems that:

* `Py_eval_input` returns an object of 3
* `Py_single_input` prints 3 to STDOUT, and returns an object of None
* `Py_file_input` return an object of None

stevengj:
That's right.  Only `Py_eval_input` returns the result of a expression.   To return the result of the last expression like in IPython it seems we have to parse the string expression by expression, see #255

utensil:
Oh yes, #255 is the right issue for me to track.

stevengj:
f07f94db7c690073ee738379494341c4276cdea7 now adds the `py"..."` syntax, and `pyeval` is deprecated.   `$` interpolation syntax is now used to pass Julia variables (as converted values — this is not string interpolation, despite the resemblance).   If you use a multi-line string, then it uses `Py_file_input` (in the global `__main__` Python namespace), so you can define Python classes, functions, etcetera and refer to them in subsequent `py"..."` strings.  e.g.

```jl
julia> py"""
       def foo(x):
           return x + 1
       """

julia> x = 17
17

julia> py"foo"(x)
18

julia> py"foo($x)"
18
```

Issue #167, opened 2015-07-20
take better advantage of static compilation


stevengj:
From stevengj/PyPlot.jl#143, almost all the time to load PyPlot is eliminated if I put `pyinitialize()` at the end of PyCall.jl.

However, it goes back to 3.8 seconds (vs. 7.2 seconds without precompilation) if I put `pyinitialize()` in `__init__` (which is where it belongs if I'm going to execute it automatically, since it needs to initialize lots of things at runtime.

Is there a way to re-arrange things so that we get more benefit from precompilation while still being safe?

stevengj:
I've been thinking about it, and maybe the best approach is to pick a specific `libpython` in `Pkg.build` and stick with it, rather than detecting it at runtime.  Advantages:

* Should enable full static compilation.
* You only have to set `PYTHON` and `PYTHONPATH` once, and thereafter it will remember it.
* Should eliminate the need for JuliaLang/pyjulia#33 and similar hacker with trying to load the running libpython (as long as you use PyCall from the same python that it was built with).

Disadvantages:

* Changing your Python requires you to re-run `Pkg.build("PyCall")`.  (e.g. from Python 2 to Python 3, or to a different Python distro; should not be required from e.g. 2.7.10 to 2.7.11, because that does not change the libpython path.)
* The pyjulia module will only work with a single Python version on your system at a time.  (Unless multiple versions of PyCall are built into different paths, which I suppose could be done by someone like Debian via fiddling with environment variables.)

cc: @malmaud, @jakebolewski

malmaud:
+1, as long as it's clearly documented. Maybe export an intuitively-named function to delete the cache and rebuild against the current runtime environment? 

In the long term, I have been wondering if it might make sense to integrate an Anaconda python distribution with the Julia package manager, so the Python distribution is in a known location and configuration. Advanced users could override this to point to a different Python environment, similar to the situation with BLAS in Julia today. That might help with the constant IJulia problems people experience.

stevengj:
The rebuild has to be done when the PyCall module is not loaded yet, so I think we are stuck with `Pkg.build("PyCall")`.  On the bright side, that's how you do it for every other Julia package.  Certainly it will be documented prominently in the PyCall README.

jakebolewski:
The user benefit here in terms of performance trumps the downside of statically picking one libpython to use.  If you are doing serious work, switching python versions I feel is pretty rare.

stevengj:
[...]

PR #169, opened 2015-07-21
determine libpython at Pkg.build time, not at runtime

stevengj:
See discussion in #167.

With `Base.compile(:PyCall)`, the load time is currently down to about 2 seconds on my machine.  I'm not sure how to get it down further, although it should be possible.

cc: @malmaud, @jakebolewski


malmaud:
I tested locally and it decreases load time from ~4s to ~2s.

stevengj:
I just pushed a few precompilations that get it down to 1.3 seconds on my machine.  (I'm guessing the improvements on Windows, compared to the old PyCall, may be even greater.)

stevengj:
Oh crap, I checked in `deps.jl`, not `build.jl`.

PR #172, opened 2015-07-24
give a descriptive error message if deps.jl does not exist

tkelman:


src/PyCall.jl
```diff
@@ -30,6 +30,8 @@ end
 
 #########################################################################
 
+isfile(joinpath(dirname(@__FILE__), "..", "deps", "deps.jl")) ||
+    error("PyCall not properly installed. Please run Pkg.build(\"PyCall\")")
 include(joinpath("..", "deps", "deps.jl")) # generated by Pkg.build("PyCall")
```
stevengj: Should this line be using `dirname(@__FILE__)` too?  Is there some reason a relative path wouldn't work here?

tkelman: maybe it should, I can make that change if you'd prefer

stevengj: Thanks, it would be good to be consistent.

PR #182, opened 2015-08-28
Add appveyor testing

davidanthoff:
This adds CI testing on Windows via appveyor. I tried it out on my fork, and all tests pass for julia 0.3. There are errors and problems with the nightlies, but those are probably just things that actually need to be fixed.

One problem is that build errors (``Pkg.build("PyCall")``) do not seem to trigger a test failure. Probably a more generic problem, but something that hits here. I'll continue to figure out what do do about that.

Once this is merged @stevengj needs to go to www.appveyor.com, create an account if he doesn't have one and add this project. At that point he can also add a badge to the readme.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,81 @@
+environment:
+  matrix:
+
+# Released versions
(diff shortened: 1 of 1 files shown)


stevengj:
Thanks!

PR #191, opened 2015-09-13
WIP towards make PyCall use Conda if system python is not found

dhoegh:
This is a WIP towards figuring out how Conda can be used as a replacement of system python and it is parallel work to https://github.com/JuliaLang/IJulia.jl/pull/357.
I still haven't figured out the API for how PyPlot/PyCall/IJulia should interact with Conda and if they should require Conda or if the user should add Conda.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -9,6 +9,7 @@
 isfile("deps.jl") && rm("deps.jl")
 
 using Compat
+import Conda
 
 PYTHONIOENCODING = get(ENV, "PYTHONIOENCODING", nothing)
 PYTHONHOME = get(ENV, "PYTHONHOME", nothing)
@@ -121,7 +122,22 @@ end
 # need to be able to get the version before Python is initialized
 Py_GetVersion(libpy) = bytestring(ccall(Libdl.dlsym(libpy, :Py_GetVersion), Ptr{UInt8}, ()))
 
-const python = get(ENV, "PYTHON", "python")
+use_conda = false
+const python = try
+    let py = get(ENV, "PYTHON", "python"), vers = convert(VersionNumber, pyconfigvar(py,"VERSION","0.0"))
+        if vers < v"2.7"
+            error("Python version $vers < 2.7 is not supported")
+        end
+        py
+    end
+catch e1
+    info( "No system-wide Python was found; got the following error:\n",
+    "$e1\nusing the Python distribution in the Conda package")
+    use_conda = true
+    Conda.add("numpy")
+    abspath(Conda.PYTHONDIR, "python" * (@windows? ".exe" : ""))
+end
+
 const (libpython, libpy_name) = find_libpython(python)
 const programname = pysys(python, "executable")
 
@@ -181,6 +197,8 @@ open("deps.jl", "w") do f
           const pyprogramname = $pystring("$(escape_string(programname))")
           const pyversion_build = $(repr(pyversion))
           const PYTHONHOME = $pystring("$(escape_string(get(ENV, "PYTHONHOME", "")))")
+          "Returns a boolean of whether the Python distribtion in the Conda package is used"
+          const use_conda = $use_conda
 
           const PyUnicode_AsUTF8String = :$PyUnicode_AsUTF8String
           const PyUnicode_DecodeUTF8 = :$PyUnicode_DecodeUTF8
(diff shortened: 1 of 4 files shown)


stevengj:
The difficulty with this is that many people using PyCall will be Python users who want to call Python packages that they've installed by their familiar way in Python.   If PyCall uses Conda's Python, they will be surprised because it won't find any of the packages they've installed with their regular Python.

dhoegh:
Yes that is the reasoning about not requiring Conda, but if Conda is installed then it prefers the python distribution in Conda. This means the changes will not affect people having a system wide python and do not have Conda installed.

dhoegh:
I have updated the PR. The miniconda python distribution in Conda only takes up 123 MB when nothing is installed which acceptable. I have added a `use_conda` constant for the purpose of automatically installation of python dependencies when Conda is used. This is shown in https://github.com/stevengj/PyPlot.jl/pull/170.

test/runtests.jl
```diff
@@ -1,4 +1,7 @@
 using Base.Test, PyCall, Compat
+import Conda
+
+PyCall.use_conda && Conda.add("numpy")
```
stevengj: Probably `Conda.add("numpy")` should be done in `Pkg.build`.

deps/build.jl
```diff
@@ -121,7 +122,17 @@ end
 # need to be able to get the version before Python is initialized
 Py_GetVersion(libpy) = bytestring(ccall(Libdl.dlsym(libpy, :Py_GetVersion), Ptr{UInt8}, ()))
 
-const python = get(ENV, "PYTHON", "python")
+use_conda = false
+const python = try 
+    pyconfigvar(get(ENV, "PYTHON", "python"),"VERSION","")
```
stevengj: Probaby call `get(ENV, "PYTHON", "python")` only once, and check for Python >= 2.7.   i.e.
```jl
let py = get(ENV, "PYTHON", "python"), vers = VersionNumber(pyconfigvar(py,"VERSION","0.0"))
    if vers < v"2.7"
        error("Python version $vers < 2.7 is not supported")
    end
end
```

deps/build.jl
```diff
@@ -121,7 +122,17 @@ end
...
+    get(ENV, "PYTHON", "python")
+catch e1
+    info( "No system-wide Python was found; got the following error:\n",
+    "$e1\nusing the Python distribution in the Conda package")
+    use_conda = true
+    abspath(Conda.PREFIX, "python" * (@windows? ".exe" : ""))
```
stevengj: Shouldn't that be `Conda.BINDIR`?

stevengj:
Would be good to have Travis testing too.   Maybe just add Python 2.6 to the Travis tests, if that is allowed by Travis.  With the version check I suggested above, that will force Pkg.build to use Conda.

deps/build.jl
dhoegh: No, Conda.BINDIR points to the binaries of installed packages and not the python binaries.

dhoegh:
I have addressed your comments. Maybe you could cancel the old appveyor build to reduce the load on appveyor.

stevengj:
cancelled the build.

stevengj:
Travis is failing with `could not spawn `/home/travis/.julia/v0.3/Conda/deps/usr/python -c "import distutils.sysconfig; print(distutils.sysconfig.get_config_var('VERSION'))"`: no such file or directory (ENOENT)`

Probably, `python` is in `Conda.BINDIR` on non-Windows systems.

(I don't understand why they couldn't just use the same paths on Unix and Windows.)

deps/build.jl
dhoegh: I am tired of windows/linux differences. On linux it should be `abspath(Conda.BINDIR, "python"))` on windows it should be `abspath(Conda.PREFIX, "python".exe"))`. This almost dictates another constant in Conda called `PYTHONDIR` else we should adjust PyCall in this special case.

dhoegh:
I have made a PR to `Conda.jl` to add `PYTHONDIR`. I will comeback tomorrow and to finish.

Luthaf:
> (I don't understand why they couldn't just use the same paths on Unix and Windows.)

Because they reproduce the way Python is packaged on Windows. But I have no idea why it was done so for Python in the first place.

tkelman:
endless frustration caused by someone deciding to be different on different platforms decades ago

dhoegh:
Everything is passing now.

stevengj:
Thanks!

PR #195, opened 2015-09-18
fixed typo in pyincref

jmxpearson:
This is a typo, right? Now matches pydecref.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -82,7 +82,7 @@ function pydecref(o::PyObject)
 end
 
 function pyincref(o::PyObject)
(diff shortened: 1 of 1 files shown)


stevengj:
No, it should be fine as is, because of this `convert` definition: https://github.com/stevengj/PyCall.jl/blob/804e065479362f0f3f8a413f8d9ddeb201fccfbc/src/PyCall.jl#L115

jmxpearson:
I see. Sorry and thanks!

stevengj:
If anything, `pydecref` should be changed to drop the `.o` in the `ccall` argument, for consistency's sake.

PR #201, opened 2015-10-02
Update conversions.jl

sbromberger:
Fixes

```
WARNING: BigInt(s::AbstractString) is deprecated, use parse(BigInt,s) instead.
 in depwarn at deprecated.jl:73
 in call at deprecated.jl:50
 in convert at /Users/bromberger1/.julia/v0.4/PyCall/src/conversions.jl:712
```

PR #202, opened 2015-10-02
Update conversions.jl

sbromberger:
See #201.


stevengj:
You know, you can do a forced push (after squashing commits) to update an existing PR, rather than creating a new PR...

sbromberger:
Understood, but I don't know how to squash commits when the files were edited within the github.com editor.

stevengj:
Can you add a test? Just convert(BigInt, PyObject(1234)) == 1234 will test this function.

sbromberger:
Sure, but I won't be able to squash the commits...

PR #217, opened 2015-11-16
Support Python 3.5 (and its @ operator - the "dedicated infix …

PallHaraldsson:
…operator for matrix multiplication")

http://legacy.python.org/dev/peps/pep-0465/

[TL;DR Feel free to ignore.. I do not want to take up time, just thought 3.5 might have been overlooked/not known.. If not drop some other 3.x?]

I suppose the @ operator already works (maybe tests are needed?).. I'm not pushing for supporting it or anything else new in 3.5, mostly wandering if anything needs supporting and/or if in general latest release needs/should to be tested against".

https://www.python.org/downloads/release/python-336/
"Release Date: 2014-10-12

This is a security-fix source-only release. The last binary release was 3.3.5."

https://www.python.org/downloads/release/python-326/
"Release Date: 2014-10-12

This is a security-fix source-only release. The last binary release was 3.2.5."

Maybe you do not want too test against to many versions?

It's unclear to me if Python 3.2, 3.3 are EOLed (can either/both be dropped?); if they need testing and or 3.4 (just skip that one?):

https://www.python.org/downloads/release/python-336/
"Release Date: 2014-10-12

This is a security-fix source-only release. The last binary release was 3.3.5."

Not important (to me), just an example I found:

https://wiki.openstack.org/wiki/Python3#Python_2:_Python_2.6_support_dropped.2C_Python_2.7_only
"OpenStack Liberty targets Python 2.7 and 3.4.
[..]
Python 3.3 support is being dropped since OpenStack Liberty."

I'm not even sure how many use any 3.x with Julia but it seems we want to support/test with 2.7 for a while more..:

http://blog.startifact.com/posts/the-call-of-python-28.html
"Python 2.7 is still massively popular: the end of life date for Python 2.7 was changed by Guido to 2020 (it was 2015). In the same change he felt he should repeat there will be no Python 2.8"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
     - 2.7
     - 3.2
     - 3.3
+    - 3.5
 compiler:
   - clang
 env:


PallHaraldsson:
I'm new to this.. Maybe I should have added also to appveyor.yml?

PallHaraldsson:
Not sure if important [I know Anaconda uses 3.5 and I see:] 

http://conda.pydata.org/docs/
"python 2.7, 3.3, 3.4, or 3.5"

stevengj:
The Appveyor failure is not due to this.  Something else caused it to fail recently and I haven't tracked down the source yet.

stevengj:
(To access the `@` operator from PyCall, just call the `__matmul__` method of an object as described in PEP 465.)

PallHaraldsson:
Since it seems no problem to have added 3.5 (adding load on Travis(?) CI; not dropping some other version in exchange), maybe you would want to also add 3.4 "for completeness" (it seems at least that version is still supported(?)).

OR you can just ignore this message, I'm not really worried about this, this is just for testing and 3.4 I guess works (since older and newer versions are tested..) and/or you may want to NOT support 3.4 anyway.


[Anyway I'll delete my branch and will not send PR for 3.4  (or for @ operator, didn't really think it needed supporting, or even that it was that useful to the Julia side (maybe for pyjulia), just wanted to make sure. Anyway thanks for pointing out __matmul__).

"The Appveyor failure is not due to this" - I wasn't aware, and not really up to speed on this Appveyor thing (you do not have to educate me), or concerned with that except if appveyor.yml should somehow be in sync with travis.yml: E.g. if python35 should also be there, I actually see python34 (but no python32..) that "contradicts"(?) travis.yml.]

PR #222, opened 2016-01-05
Python 2.6 is not supported, then travis testing should be skipped? [ci skip]

PallHaraldsson:
According to README at least

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-    - 2.6
     - 2.7
     - 3.2
     - 3.3


PallHaraldsson:
A. Unless I'm mistaken 2.6 doesn't work, according to source code also, so no need for load on travis? Unless I didn't read right and Conda made an exception. No need to answer, but I would be curious.

Anaconda (and Miniconda) are at 2.7 (not sure if some old 2.6 was meant to work and test for, then unclear why regular Python 2.6 wouldn't also work):

https://www.continuum.io/downloads


B. I see 3.4 isn't tested (but is in pyjulia, while there 3.5 isn't), but assume no need to as 3.5 test would cover.. I guess pyjulia doesn't really need to be in sync (or vise versa) with 3.5 tested there as all the heavy lifting seems to be here. I'm going to be telling people pyjulia now works with 0.4 (my understanding) with 2.7 (not worried about 3.5 working..).

Monday, December 7, 2015
"Python 3.5.1 and Python 3.4.4rc1 are now available for download.
[..]
Python 2.7.11 was also released today, making this a Python release-day hat trick!"

PallHaraldsson:
In case this https://github.com/JuliaLang/pyjulia/pull/48 is not seen (because of bogus failed check) or ignored.. letting know here.

PallHaraldsson:
[Alternatively to stop 3.6 test:] One more thing, if/since the 2.6 travis check is currently running, would there be a reason to allow Python 2.6 compatibility?

I assume not, and that PyCall actually doesn't work with 2.6, why there is a [non-redundant] check in the code, then there is some missing unit-test that should have broke for 2.6?

stevengj:
The 2.6 test is intentional.  The build script should detect that Python 2.6 is present and use Conda instead, so the point of this test is actually to test the Conda Python.

PR #223, opened 2016-01-10
Remove deprecated FileOffset

yuyichao:
Ref JuliaLang/julia#14590


yuyichao:
As Jameson pointed out, `sizeof(FileOffset)` is always `8` on all the platforms we support. (And I pushed a test commit to check and it is indeed the case) so I've removed the version check and always use `Int64` on all versions.

Issue #226, opened 2016-01-26
Error when filtering PyDict


ranjanan:
Consider the following code segment:
```julia
d = PyDict{Int,Int}()
d[1] = 1
d[2] = 2
d[3] = 3
function f(x,y)
  y == 1
end
filter!(f, d)
```
It gives the following error:
```
ERROR: UndefVarError: items not defined
 in filter! at /Users/ranjan/.julia/v0.4/PyCall/src/conversions.jl:562
```
The source code for `fitler!` is [here](https://github.com/stevengj/PyCall.jl/blob/master/src/conversions.jl#L537). Now removing `items` makes the function work. (I've sent PR #227 just in case this is the right fix. ) There's a comment above as to why we shouldn't do that. 
Could you show me an example where just having `(k,v)` wouldn't work? I cannot see `PyDict_Next` used for `delete!`. 
In such a case, how do we fix this?

PR #227, opened 2016-02-03
Fix dict filter by removing `items`, Fix tests for v0.3 compat

ranjanan:
Fix the function `filter!()` by removing `items(d)`

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -538,7 +538,7 @@ function filter!(f::Function, d::PyDict)
     # We must use items(d) here rather than (k,v) in d,
     # because PyDict_Next does not permit changing the set of keys
     # during iteration.
-    for (k,v) in items(d)
+    for (k,v) in d
         if !f(k,v)
             delete!(d,k)
         end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -38,8 +38,8 @@ end
 @test roundtripeq((1,(3.2,"hello"),true)) && roundtripeq(())
 @test roundtripeq(Int32)
 @test roundtripeq(@compat Dict(1 => "hello", 2 => "goodbye")) && roundtripeq(Dict())
-@test roundtripeq(UInt8[1,3,4,5])
-@test convert(Vector{UInt8}, "hello") == "hello".data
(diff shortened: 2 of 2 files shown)


ranjanan:
This should fix #226

ranjanan:
@tkelman Could you please help with code review?

ranjanan:
The Appveyor build fails because of [this](https://github.com/stevengj/PyCall.jl/blob/master/test/runtests.jl#L41). There needs to be an `@compat` there for all the `UInt8`'s. But I'm curious about why Appveyor checks against v0.3 while Travis checks against v0.4

tkelman:
The [.travis.yml file here](https://github.com/stevengj/PyCall.jl/blob/master/.travis.yml) is still using the PPA for installing Julia, which does not make it very easy to install 0.3 any more. You can't use multiple `language:` settings, and it's a little more effort to install specific versions of Python than it is specific versions of Julia. The Travis file could be reformatted to download the generic Linux binaries (and osx dmg installers if osx is running) instead of the PPA to restore testing against 0.3.

ranjanan:
I see. Maybe I should modify Travis to do pull the v0.3 binaries and send a separate PR. Also, the above commit fixes #228

stevengj:
Doing `for (k,v) in d` uses the dictionary iterator, which calls `PyDict_Next`.  The [documentation for the latter function](https://docs.python.org/3.2/c-api/dict.html) explicitly says "The dictionary `p` should not be mutated during iteration"

Hence, I don't think this change is a good idea; there must be a more reliable way to fix it.

stevengj:
Two options that I can see:

* Replace `items(d)` (which must have gotten removed from Julia at some point) by e.g. `collect(Tuple{PyObject,PyObject}, d)`.  This unfortunately makes a copy of the data.
* Do `for (k,v) in d` or similar, but instead of deleting the keys right away, `push!` them to a temporary array of keys to be deleted.   This is a bit more memory-efficient for large dictionaries.

stevengj:
Fixed by 73c3db71e8e1eca4ef1e950e5aaabe65f544b159

PR #236, opened 2016-02-18
Add docs for some exported functions

ranjanan:
This adds some basic documentation for some of the exported APIs.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -60,11 +60,14 @@ typealias PyPtr Ptr{PyObject_struct} # type for PythonObject* in ccall
 # Wrapper around Python's C PyObject* type, with hooks to Python reference
 # counting and conversion routines to/from C and Julia types.
 """
-This is a reference to a Python Object. Julia objects may be converted to Python Objects using `PyObject(juliavar)` and vice-versa using `convert(T, o::PyObject)`.
+`PyObject(juliavar)`
 
-Given o::PyObject, `o[:attribute]` is equivalent to `o.attribute` in Python, with automatic type conversion.
+This converts a julia variable to a PyObject, which is a reference to a Python object. 
+You can convert back to native julia types using `convert(T, o::PyObject)`, or using `PyAny(o)`.
 
-Given o::PyObject, `get(o, key)` is equivalent to `o[key]` in Python, with automatic type conversion.
+Given `o::PyObject`, `o[:attribute]` is equivalent to `o.attribute` in Python, with automatic type conversion.
+
+Given `o::PyObject`, `get(o, key)` is equivalent to `o[key]` in Python, with automatic type conversion.
 """
 type PyObject
     o::PyPtr # the actual PyObject*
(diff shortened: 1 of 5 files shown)


src/PyCall.jl
```diff
@@ -59,7 +59,13 @@ typealias PyPtr Ptr{PyObject_struct} # type for PythonObject* in ccall
 #########################################################################
 # Wrapper around Python's C PyObject* type, with hooks to Python reference
 # counting and conversion routines to/from C and Julia types.
+"""
+This is a reference to a Python Object. Julia objects may be converted to Python Objects using `PyObject(juliavar)` and vice-versa using `convert(T, o::PyObject)`.
 
+Given o::PyObject, `o[:attribute]` is equivalent to `o.attribute` in Python, with automatic type conversion.
```
stevengj: missing backquotes around `o::PyObject`

src/PyCall.jl
```diff
@@ -59,7 +59,13 @@ typealias PyPtr Ptr{PyObject_struct} # type for PythonObject* in ccall
 #########################################################################
 # Wrapper around Python's C PyObject* type, with hooks to Python reference
 # counting and conversion routines to/from C and Julia types.
+"""
+This is a reference to a Python Object. Julia objects may be converted to Python Objects using `PyObject(juliavar)` and vice-versa using `convert(T, o::PyObject)`.
```
stevengj: "Object" should not be capitalized.

They can be converted to native julia types `T` with `convert(T, o)` or `T(o)`.

src/PyCall.jl
```diff
@@ -264,6 +270,13 @@ keys(o::PyObject) = Symbol[m[1] for m in pycall(inspect["getmembers"],
 # we skip wrapping Julia reserved words (which cannot be type members)
 const reserved = Set{ASCIIString}(["while", "if", "for", "try", "return", "break", "continue", "function", "macro", "quote", "let", "local", "global", "const", "abstract", "typealias", "type", "bitstype", "immutable", "ccall", "do", "module", "baremodule", "using", "import", "export", "importall", "pymember", "false", "true", "Tuple"])
 
+"""
+pywrap(o::PyObject) returns a wrapper `w` that is an anonymous module which provides (read) access to converted versions of o's members as w.member.
```
stevengj: The  convention for Julia documentation is typically to put the function signature on the first line, by itself, with 4 spaces before it so that it is formatted as a code block:
```
"""
    pywrap(o::PyObject)

Return a wrapper `w` that is ...
"""

File: src/PyCall.jl [hunk 4]
```diff
@@ -264,6 +270,13 @@ keys(o::PyObject) = Symbol[m[1] for m in pycall(inspect["getmembers"],
 # we skip wrapping Julia reserved words (which cannot be type members)
 const reserved = Set{ASCIIString}(["while", "if", "for", "try", "return", "break", "continue", "function", "macro", "quote", "let", "local", "global", "const", "abstract", "typealias", "type", "bitstype", "immutable", "ccall", "do", "module", "baremodule", "using", "import", "export", "importall", "pymember", "false", "true", "Tuple"])
 
+"""
+pywrap(o::PyObject) returns a wrapper `w` that is an anonymous module which provides (read) access to converted versions of o's members as w.member. 
+
+For example, `@pyimport module as name` is equivalent to `name = pywrap(pyimport("module"))`
```
stevengj: Technically, it is equivalent to `const name = ...`.

src/PyCall.jl
```diff
@@ -285,7 +298,9 @@ end
 pyimport(name::AbstractString) =
     PyObject(@pycheckn ccall((@pysym :PyImport_ImportModule), PyPtr,
                              (Cstring,), name))
-
+"""
```
stevengj: I would attach the docstring to `pyimport(name::AbstractString)`, since that is the most common method to call; I may remove the `Symbol` version at some point, so perhaps it is better not to document that.

stevengj:
Thanks for working on this!

ranjanan:
I've incorporated your suggestions in that commit. Is there anything else you'd like me to add or modify?

src/PyCall.jl
```diff
@@ -59,7 +59,16 @@ typealias PyPtr Ptr{PyObject_struct} # type for PythonObject* in ccall
 #########################################################################
 # Wrapper around Python's C PyObject* type, with hooks to Python reference
 # counting and conversion routines to/from C and Julia types.
+"""
+`PyObject(juliavar)`
```
stevengj: The standard doc style is to indent the first line by four spaces, and not use backticks.  This will make it a code block, not just a portion of the line in code style.  i.e.
```
    PyObject(juliavar)

This converts...
```

stevengj:
Hi @ranjanan, sorry this fell off the queue for me.  Just a small formatting fix and it should be good to merge.

stevengj:
Thanks!  In the future, it is also a good idea to "squash" your commits into a single commit (via `git rebase`), but since this is just documentation patches I can merge it as-is.

(Squashing commits is especially important for code so that you don't have non-functional commits in the history.)

ranjanan:
Sure, thanks! Will keep that in mind.

PR #237, opened 2016-02-18
Converting Julia types into Python classes

cstjean:
Just the basics work.

```julia
using PyCall

type Bird
    song
end

function sing(b::Bird)
    b.song
end
function sing(b::Bird, n)
    prod([b.song for i in 1:n])
end

PyCall.def_py_methods(Bird, ("sing", sing))   # `sing` could be replaced by an anonymous function

py_bird = PyObject(Bird("hoio "))
py_bird[:sing]() # -> hoio
py_bird[:sing](3) # -> hoio hoio hoio
```
I'll work on the `@pydef` [interface](https://github.com/stevengj/PyCall.jl/issues/235#issuecomment-185801732). The neat thing about the PR so far is that we get multiple dispatch for free, and it would be very natural to do a macroexpansion of this:

```julia
@pydef type IO <: somepythontype
     readline = jl_IO_readline
     multiply = (io, x, y)->x*y
     __get__ = (io, field)->getfield(io, symbol(field))
     __set__ = io_setter
end
```

How do you like it? And do you mind `import MacroTools`?


cstjean:
I suppose it depends on the use case. If people start using it to interact with Python GUI or web frameworks, then it's worth making a really nice interface.

cstjean:
This PR's code is ready for review. See `test/io_2.jl` for a demo - all tests pass with the `@pydef` definition. I'd like to add a section to the README sometimes next week...

Inheritance works when inheriting from my own simple classes, but unfortunately it segfaults when the base class is more complicated. For instance:

```julia
using PyCall

@pyimport Tkinter

type Example end

@pydef type Example <: Tkinter.Frame
end
```
Could you please have a careful look at [the pyjlwrap_type call](https://github.com/stevengj/PyCall.jl/pull/237/files#diff-d2b8a921db0313b6372f90f1aecd211cR168)? I'll keep debugging and trying to isolate the issue, but if you see something obviously wrong, it would help. `pyjlwrap_type` has this bit of magic I don't quite get:

```
                     t.tp_base = ccall(:jl_value_ptr, Ptr{Void},
                                       (Ptr{PyTypeObject},),
                                       &jlWrapType)
```

and my code sets `t.tp_base` to something else.

cstjean:
I'm sorry, I don't think I can make inheritance of user-defined classes work at this time. It's non-trivial: Cython [explicitly forbids it](http://docs.cython.org/src/userguide/extension_types.html#subclassing), and as far as I can tell, Numba [never supported it either](https://github.com/numba/numba/issues/1336). 

That limits the usefulness of this PR. It's great at defining the wrapper for `IO`, but all the Python frameworks I've looked at require inheritance (including `scikit-learn`, which was my target). 

The current code segfaults during `PyType_Ready` when inheriting from complex classes like `Tkinter.Frame`, and I still don't know why. But even if it didn't, we don't currently account for the space taken by the base type. [The C extension tutorial](https://docs.python.org/2/extending/newtypes.html#subclassing-other-types) defines the inherited class' object structure as

```C
typedef struct {
    PyListObject list;
    int state;
} Shoddy;
```

for a type inheriting from `list`. "Its PyObject* pointer can be safely cast to both PyListObject* and Shoddy*". PyCall currently assumes this layout:

```julia
immutable Py_jlWrap
    ob_refcnt::Int
    ob_type::PyPtr

    jl_value::Any
end
```

We'll need to get the base class's `basicsize` and store `jl_value` afterwards. I'm not comfortable enough with `ccall` and CPython to make that happen.

stevengj:
@cstjean, that bit of magic is just a hack to get the raw pointer to a Julia object, in this case to declare a subtype of `jlWrapType`.

Actually maybe you should leave `tp_base` alone and use [tp_bases](https://docs.python.org/2/c-api/typeobj.html#c.PyTypeObject.tp_bases)

stevengj:
Or, hmm, no, it sounds like `tp_bases` might be set up automatically by libpython.

stevengj:
It doesn't sound infeasible to get the `PyTypeObject*` for a Python type, extract `tp_basicsize`, and do pointer-twiddling to look past that in a Julia derived-type object to get the `jl_value`.

cstjean:
> It doesn't sound infeasible to get the PyTypeObject* for a Python type, extract tp_basicsize, and do pointer-twiddling to look past that in a Julia derived-type object to get the jl_value.

If this is the only issue, I agree that it's solvable, but I still don't know why `PyType_ready` segfaults, and I suspect that it's because `tp_base` is primarily designed for C-defined types, and not complex Python-defined classes (though it works if the class is simple). Re. multiple-inheritance, the docs say about `tp_base`:

> At this level, only single inheritance is supported; multiple inheritance require dynamically creating a type object by calling the metatype.

I think that's what we'll need to do sooner or later. Do you have any experience with that?

I'm willing to try some more if you have any idea. I couldn't find any example of inheriting from a Python class, and I don't think that's a coincidence! It would help a lot.

cstjean:
Since PyCall can already pass Julia functions, it might be much simpler to call [type](https://jeffknupp.com/blog/2013/12/28/improve-your-python-metaclasses-and-dynamic-classes-with-type/) to create a metaclass instead of fighting with the C api.

stevengj:
Closing in favor of #250.

PR #250, opened 2016-03-22
Defining Python classes through `type`

cstjean:
Follow-up to #237. Using `type` to create the metaclass was much simpler than using the C API. Everything works, including multiple inheritance and special Python methods (`__init__`, `__len__`, etc.) The interface changed slightly, because we don't use `jlWrap` anymore:

```julia
PyCall.@pydef type PyIO
    __init__(self, io::IO; istextio=false) = begin
        self[:io] = io
        self[:istextio] = istextio
        nothing
    end
    closed.get(self) = !isopen(self[:io])
    fileno(self) = fd(self[:io])
    ....
end

let buf = IOBuffer(false, true), obuf = PyIO(buf)
    ...
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,6 +337,63 @@ For example, calling `f(x, function=g)` will fail because `function` is
 a reserved word in Julia. In such cases, you can use the lower-level
 Julia syntax `f(x; :function=>g)`.
 
+### Defining Python Classes
+
+`@pydef` creates a Python class whose methods are implemented in Julia.
+For instance,
+    
+    @pyimport numpy.polynomial as P
+
+    @pydef type Doubler <: P.Polynomial
+       __init__(self, x=10) = (self[:x] = x) 
+       my_method(self, arg1=5) = arg1 + 20  # the right-hand-side is Julia code
+       x2.get(self) = self[:x] * 2
+       x2.set!(self, new_x::Int) = (self[:x] = new_x / 2)
+    end
+
+    Doubler()[:x2]
+
+is equivalent to
+    
+    class Doubler(numpy.polynomial.Polynomial):
+       def __init__(self, x=10):
+          self.x = x
+
+       def my_method(self, arg1):
+          return arg1 + 20
+
+       @property
+       def x2(self): return self.x * 2
+
+       @x2.setter
+       def x2(self, new_val):
+           self.x = new_val / 2
+
+The methods' arguments and return values are automatically converted. All Python
+special methods are supported (__len__, __add__, etc.)
+
+`@pydef` allows for multiple-inheritance of Python types:
+
+    @pydef type SomeType <: (BaseClass1, BaseClass2)
+        ...
+    end
+
+Here's another example using [Tkinter](https://wiki.python.org/moin/TkInter)
+
+    using PyCall
+    @pyimport Tkinter as tk
+
+    @pydef type SampleApp <: tk.Tk
+	__init__(self, args...; kwargs...) = begin
+	    tk.Tk[:__init__](self, args...; kwargs...)
+	    self[:label] = tk.Label(text="Hello, world!")
+	    self[:label][:pack](padx=10, pady=10)
+	end
+    end
+
+    app = SampleApp()
+    app[:mainloop]()
+
 ### GUI Event Loops
 
 For Python packages that have a graphical user interface (GUI),
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,3 +2,4 @@ julia 0.4
 Compat 0.7.9
 Dates
 Conda 0.1.6
+MacroTools 0.2
--- a/src/io.jl
+++ b/src/io.jl
@@ -18,341 +18,93 @@ function ioraise(e)
     end
 end
 
-function jl_IO_close(self_::PyPtr, noarg_::PyPtr)
-    try
-        io = unsafe_pyjlwrap_to_objref(self_)::IO
-        close(io)
-        ccall((@pysym :Py_IncRef), Void, (PyPtr,), pynothing::PyPtr)
-        return pynothing::PyPtr
-    catch e
+macro with_ioraise(expr)
+    :(try
+        $(esc(expr))
+      catch e
         ioraise(e)
-    end
-    return convert(PyPtr, C_NULL)
+      end)
 end
 
-# "closed" must be an attribute "for backwards compatibility"
-function jl_IO_closed(self_::PyPtr, closure::Ptr{Void})
-    try
-        io = unsafe_pyjlwrap_to_objref(self_)::IO
-        # Note: isopen is not defined for all IO subclasses in Julia.
-        #       Should we do something different than throwing a MethodError,
-        #       if isopen is not available?
-        return pyincref(PyObject(!isopen(io))).o
-    catch e
-        ioraise(e)
+function jl_io_readline(io::IO, nb)
+    d = readline(io)
+    if length(d) > nb
+        resize!(d, nb)
     end
-    return convert(PyPtr, C_NULL)
+    return d
 end
 
-function jl_IO_fileno(self_::PyPtr, noarg_::PyPtr)
-    try
-        io = unsafe_pyjlwrap_to_objref(self_)::IO
-        return pyincref(PyObject(fd(io))).o
-    catch e
-        ioraise(e)
-    end
-    return convert(PyPtr, C_NULL)
-end
-
-function jl_IO_flush(self_::PyPtr, noarg_::PyPtr)
-    try
-        io = unsafe_pyjlwrap_to_objref(self_)::IO
-        if method_exists(flush, Tuple{typeof(io)})
-            flush(io)
-        end
-        ccall((@pysym :Py_IncRef), Void, (PyPtr,), pynothing::PyPtr)
-        return pynothing::PyPtr
-    catch e
-        ioraise(e)
-    end
-    return convert(PyPtr, C_NULL)
-end
-
-function jl_IO_isatty(self_::PyPtr, noarg_::PyPtr)
-    try
-        io = unsafe_pyjlwrap_to_objref(self_)::IO
-        return pyincref(PyObject(isa(io, Base.TTY))).o
-    catch e
-        ioraise(e)
-    end
-    return convert(PyPtr, C_NULL)
-end
-
-function jl_IO_readable(self_::PyPtr, noarg_::PyPtr)
-    try
-        io = unsafe_pyjlwrap_to_objref(self_)::IO
-        return pyincref(PyObject(isreadable(io))).o
-    catch e
-        ioraise(e)
-    end
-    return convert(PyPtr, C_NULL)
-end
-
-function jl_IO_writable(self_::PyPtr, noarg_::PyPtr)
-    try
-        io = unsafe_pyjlwrap_to_objref(self_)::IO
-        return pyincref(PyObject(iswritable(io))).o
-    catch e
-        ioraise(e)
-    end
-    return convert(PyPtr, C_NULL)
-end
-
-function jl_IO_readline(self_::PyPtr, args_::PyPtr)
-    try
-        io = unsafe_pyjlwrap_to_objref(self_)::IO
-
-        nargs = ccall((@pysym :PySequence_Size), Int, (PyPtr,), args_)
-        nb = typemax(Int) # max #bytes to return
-        if nargs == 1
-            nb = convert(Int, PyObject(ccall((@pysym :PySequence_GetItem),
-                                              PyPtr, (PyPtr, Int), args_, 0)))
-            if nb < 0
-                nb = typemax(Int)
-            end
-        elseif nargs > 1
-            throw(ArgumentError("readline cannot accept $nargs arguments"))
-        end
-
+function jl_io_readlines(io::IO, nb)
+    ret = Any[]  # should it be some other type?
+    nread = 0
+    while nread < nb && !eof(io)
         d = readline(io)
-        if length(d) > nb
-            resize!(d, nb)
-        end
-        return pyincref(PyObject(d)).o
-    catch e
-        ioraise(e)
+        nread += length(d)
+        push!(ret, PyObject(d))
     end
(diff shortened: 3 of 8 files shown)


stevengj:
@cstjean, this looks great!  Sorry I haven't kept up with this.

Is there any reason not to just use this for the `jl_IO` and `jl_TextIO` classes currently defined in `src/io.jl`?   It would be nicer to eat our own dogfood here.

cstjean:
Sure. I can define `PyObject(io::IO) = PyIO(io)` to maintain the same API. Do we care about the Julia-to-Python conversion of IO objects? It would require special-casing in `convert`, and seems like a really marginal use case.

cstjean:
I had to disable the `roundtripeq` tests, but otherwise it seems ok. The tests run locally.

stevengj:
I think it would be nice to preserve round-trip equality of Julia objects where possible, including IO objects.

Maybe we can define a more general convention where, if a type has a `_jlwrap_` member, and that member has `is_pyjlwrap(o["_jlwrap_"]) == true`, then conversion returns that object.

src/io.jl
```diff
@@ -18,341 +18,93 @@ function ioraise(e)
...
-# "encoding" attribute for TextIOBase
-function jl_IO_encoding(self_::PyPtr, closure::Ptr{Void})
-    try
-        return pyincref(PyObject("UTF-8")).o
-    catch e
-        ioraise(e)
-    end
-    return convert(PyPtr, C_NULL)
-end
+jl_readall(io::IO, istextio) = istextio ? readall(io) : readbytes(io)
```
stevengj: Probably we should just push one-liners like this into the `@pydef`

src/io.jl
```diff
@@ -18,341 +18,93 @@ function ioraise(e)
...
-                              t.tp_getset = pointer(jl_IO_getset)
-                              t.tp_repr = pyio_repr_ptr
-                          end)
+        @eval @pydef type PyIO
+            __init__(self, io::IO; istextio=false) = begin
+                self[:io] = pyjlwrap_new(io) # avoid recursion
+                self[:istextio] = istextio
+                nothing
+            end
+            close(self) = @with_ioraise(close(self[:io]))
```
stevengj: Since we know the type of the `io` member, probably we should do `unsafe_pyjlwrap_to_objref(self["io"])::IO` instead of `self[:io]` to avoid the overhead of the runtime type detection and to add compile-time type information. (Probably via a helper function `pyio_jl(self)` for this.)

cstjean:
> Maybe we can define a more general convention where, if a type has a _jlwrap_ member

I just tried that. It causes problems with the Python methods, because `self` is now auto-converted to `::IO`, and I can't write the constructor correctly:

```julia
            __init__(self, io::IO; istextio=false) = begin
                self[:_jlwrap_] = pyjlwrap_new(io) # avoid recursion
                self[:istextio] = istextio
                nothing
            end
```

Also, even if `roundtripeq(STDIO)` is true, `roundtripeq(TextIO(...))` still won't work because we don't have anywhere to store `istextio` in Julia.

Do you have a situation in mind where auto-converting to Julia will be useful? I feel like this is a separate problem, and we'll need a different interface to solve it.

stevengj:
Okay, I agree that we can try to handle this separately if it ever comes up in practice; I don't see a big need for round-tripping `IO` objects at the moment.

README.md
```diff
@@ -331,6 +331,61 @@ and also by providing more type information to the Julia compiler.
...
   `w.pymember("member")` (for the raw `PyObject`).
 
+### Defining Python Classes
+
+`@pydef` creates a Python class. For instance,
+    
+    @pyimport numpy.polynomial as P
+
+    @pydef type Doubler <: P.Polynomial
+       # __init__ must return `nothing`
```
stevengj: Would be nice if the `@pydef` macro enforced this automatically.   i.e. we special-case the `__init__` function by making `__init__` point to a function whose body is `(__user's init__(args...); nothing)`.

README.md
```diff
@@ -331,6 +331,61 @@ and also by providing more type information to the Julia compiler.
...
+
+    Doubler()[:x2]
+
+is equivalent to
+    
+    class Doubler(numpy.polynomial.Polynomial):
+       def __init__(self, x=10):
+          self.x = x
+
+       def pymethod1(self, arg1):
```
stevengj: `py_method` for consistency with the definition above?  Maybe call it `my_method1` instead, since `py_method` might give the wrong impression that there is something "pythonic" about this function which is actually in pure Julia.

README.md
```diff
@@ -331,6 +331,61 @@ and also by providing more type information to the Julia compiler.
   instead use `w.pymember(:member)` (for the `PyAny` conversion) or
   `w.pymember("member")` (for the raw `PyObject`).
 
+### Defining Python Classes
+
+`@pydef` creates a Python class. For instance,
```
stevengj: "... creates a Python class whose methods are implemented in Julia."

REQUIRE
```diff
@@ -2,3 +2,4 @@ julia 0.4
 Compat 0.7.1
 Dates
 Conda 0.1.6
+MacroTools 0.2
```
stevengj: add a newline at the end...

src/io.jl
```diff
@@ -18,341 +18,92 @@ function ioraise(e)
...
-                          end)
-        global const jl_TextIOType =
-            pyjlwrap_type("PyCall.jl_TextIO",
-                          t -> begin
-                              t.tp_getattro = @pyglobal(:PyObject_GenericGetAttr)
-                              t.tp_methods = pointer(jl_TextIO_methods)
-                              t.tp_getset = pointer(jl_IO_getset)
-                              t.tp_repr = pyio_repr_ptr
-                          end)
+        @eval @pydef type PyIO
```
stevengj: It would be nice to have a lower-level macro so that we don't have to use `@eval`, i.e. a `@pydef` macro that simply returns the Python type object without defining any Julia constant.   In the absence of a better name, maybe `@pydef_`.   Then we can do:
```jl
const PyIO = PyNULL()
function pyio_initialize()
    if PyIO.o == C_NULL
        copy!(PyIO, @pydef_ .....)
end
```
(I'd like to move away from all of these `global const Foo = ....` definitions that I currently have, since they screw up type inference in precompiled code.)

cstjean:
Do you have any suggestion for a better example? `Doubler` is lame, and the inheritance is pointless.

src/pyinit.jl
```diff
@@ -110,6 +100,15 @@ function __init__()
...
                                                  jl_Function_call_ptr)
 
+    # jl_FunctionType is a class, and when assigning it to an object
+    #    obj[:foo] = some_julia_function
+    # it won't behave like a regular Python method because it's not a Python
+    # function (in particular, `self` won't be passed to it). The solution is:
+    #    obj[:foo] = jlfun2pyfun(some_julia_function)
+    global const jlfun2pyfun =
+        pyeval("""lambda f: lambda *args, **kwargs: f(*args, **kwargs)""")
+    global const pyproperty = pybuiltin(:property)
```
stevengj: For global constant Python objects (any `PyObject` instance), to get proper type inference you should do `const pyproperty = PyNULL()` at the top-level scope followed by `copy!(pyproperty, pybuiltin("property"))` in `__init__`

src/pyclass.jl
```diff
@@ -0,0 +1,164 @@
...
+            @assert @capture(lhs,py_f_(args__)) "Malformed left-hand-side: $lhs"
+            if isa(py_f, Symbol)
+                # Method definition
+                # We save the gensym to support multiple dispatch
+                #    readlines(io) = ...
+                #    readlines(io, nlines) = ...
+                # otherwise the first and second `readlines` get different
+                # gensyms, and one of the two gets shadowed by the other.
+                jl_fun_name = get!(method_syms, py_f, gensym(py_f))
+                push!(function_defs, :(function $jl_fun_name($(args...))
```
stevengj: Right here, you could have a check: if `py_f == :init`, then append a `nothing` line to the function body.

stevengj:
This is really awesome code, by the way; I'm impressed!

cstjean:
I think I've fixed everything. Thank you for the kind words!

`@pydef_object` works, but it's still defining the class methods inside `pyio_initialize`, when they could be defined at the top-level. We could fix it by creating `@pydef_precompiled`, which would expand into the `PyNULL` trick. Is that worth the effort? Maybe I can leave it for a future PR. I wish that Julia had Common Lisp's `eval-when`, that would solve the issue.

src/pyclass.jl
```diff
@@ -160,9 +168,17 @@ Multiple dispatch works, too:
     x2.set!(self, new_x::Int) = ...
     x2.set!(self, new_x::Float64) = ...
 """
-macro pydef(type_expr)
-    type_name, base_classes, methods_, getter_dict, setter_dict, function_defs =
-        parse_pydef(type_expr)
+macro pydef(class_expr)
+    class_name, _, _ = parse_pydef_toplevel(class_expr)
+    :($(esc(class_name)) = @pydef_object($(esc(class_expr))))
```
stevengj: Probably should be `const class_name = ...`, since types in Julia are usually `const`

src/pyclass.jl
```diff
@@ -160,9 +168,17 @@ Multiple dispatch works, too:
...
+end
+
+
+""" `@pydef_object` is like `@pydef`, but non-mutating - it returns the
```
stevengj: Rather than "non-mutating", I would say "is side-effect free" (modifying a global variable is a side effect, whereas mutation usually refers to changes to the actual parameters).

stevengj:
Let's leave further precompilation improvements to a future PR.

In practice, if a `@pydef` refers to any large, complicated Julia functions, those should generally be separate top-level functions anyway, and as such can be precompiled.  Run-time compilation of little one-line methods like we are using in `io.jl` should be cheap.

stevengj:
The lame example is fine.   A simple-minded class is actually a good thing, since people can focus on the syntax rather than on understanding what the class does.

src/pyclass.jl
```diff
@@ -0,0 +1,196 @@
...
+    end
+
+Multiple dispatch works, too:
+
+    x2.set!(self, new_x::Int) = ...
+    x2.set!(self, new_x::Float64) = ...
+"""
+macro pydef(class_expr)
+    class_name, _, _ = parse_pydef_toplevel(class_expr)
+    :($(esc(class_name)) = @pydef_object($(esc(class_expr))))
```
stevengj: should be `const class_name = `?

cstjean:
Fixed.

> if a @pydef refers to any large, complicated Julia functions, those should generally be separate top-level functions anyway, and as such can be precompiled.

I've thought about supporting `function mymethod(self)` syntax, it would be more Python-like.

stevengj:
Supporting the `function` keyword inside `@pydef` would be good for completeness, but it can be a separate PR.

stevengj:
Should be good to merge after a rebase, assuming tests are green.

stevengj:
Just posted an announcement to julia-users; I think people will be excited by this.  https://groups.google.com/forum/#!topic/julia-users/75q3HInd1Jo

staticfloat:
What is the best way to accept kwargs in the same way that Python does?  Right now, I'm doing this:

```julia
using PyCall

macro default(varname, default_val)
    varname_direct = symbol(varname, :_direct)
    esc(quote
        if $varname === nothing && $varname_direct === nothing
            $varname = $default_val
        elseif $varname === nothing
            $varname = $varname_direct
        end
    end)
end

@pydef type Foo
    __init__(self) = begin
    end
    bar(self, var1_direct=nothing; var1=nothing) = begin
        @default var1 false
        return var1
    end
end
```

Where `@default` is a macro to merge together the two values `var1_direct` and `var1` such that a user can do the following from Python, where mixing usage of the keyword argument and non-keyword argument is more common:
```
julia> pyeval("f.bar()", f = Foo())
false

julia> pyeval("f.bar(True)", f = Foo())
true

julia> pyeval("f.bar(False)", f = Foo())
false

julia> pyeval("f.bar(var1=True)", f = Foo())
true
```

cstjean:
You could do this:

```julia
@pydef type Foo
    __init__(self) = begin
    end
    bar(self; var1=nothing) = self[:bar](var1)
    bar(self, var1) = begin
        return var1
    end
end
```

We should automatically generate those methods in the `@pydef` expansion. I bet that many frameworks call methods in both ways, and it will cause problems for PyCall users.

stevengj:
I don't know about automatically generating both variants; my feeling is the methods of `@pydef` should be ordinary Julia functions as much as possible, with Julia function semantics.   As you point out, mimicking the Python semantics with Julia functions is pretty trivial.

cstjean:
On second thought, if a framework calls a class's method in both ways, it's relying on both a specific keyword ordering and specific keyword argument names. Sounds unlikely, I agree that it's not worth supporting. @staticfloat out of curiosity, what's your use case?

staticfloat:
I'm writing a data provider for TensorFlow, although I don't _need_ this, it's just that when I write python, I occasionally mix positional and kwargs as Python doesn't make a big difference between them.  Example is things like an `stft` function, where I have `sftf(data, nfft=256, window=hamming)`; sometimes I may say `stft(x, 256, hann)`, and sometimes I may say `stft(x, window=hann)` if I decide I want to use the default `nfft` parameter.  Because this is so easy in Python, I worry that library authors will mix-and-match, and my PyCall objects will cause random crashes.  Because Python makes such little distinction between positional and keyword arguments, I would argue in favor of transparently mapping them in PyCall, but the solution that I have right now works just fine.

staticfloat:
On a second note, how do I define a function that I want to call from Julia in order to modify the python object?  I could, of course, use `pyeval()` but if I want to, say, invoke the `bar()` method directly from Julia, is there an easy way to do that?

cstjean:
> On a second note, how do I define a function that I want to call from Julia in order to modify the python object?

```julia
foo = Foo()
foo[:bar](45)
foo[:new_field] = zeros(10,10)
```

You can store arbitrary Julia objects that way (or through `__init__`)

PR #267, opened 2016-05-07
Fix depwarn on 0.5

yuyichao:
`String`, `Symbol`, `precision(BigFloat)`
Also fix an incorrect use of `unsafe_convert`.

Requires a new tag of `Compat`.
Also includes https://github.com/stevengj/PyCall.jl/pull/260

Requires https://github.com/JuliaLang/Compat.jl/pull/192

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.4
-Compat 0.7.14
+Compat 0.7.15
 Conda 0.1.6
 MacroTools 0.2
--- a/src/pytype.jl
+++ b/src/pytype.jl
@@ -261,7 +261,7 @@ type PyTypeObject
     # Julia-specific fields, after the end of the Python structure:
 
     # save the tp_name Julia string so that it is not garbage-collected
-    tp_name_save::ASCIIString
+    tp_name_save # This is a gc slot that is never read from
 
     function PyTypeObject(name::AbstractString, basicsize::Integer, init::Function)
         # figure out Py_TPFLAGS_DEFAULT, depending on Python version
@@ -280,7 +280,8 @@ type PyTypeObject
              Py_TPFLAGS_HAVE_CLASS |
              Py_TPFLAGS_HAVE_STACKLESS_EXTENSION |
              Py_TPFLAGS_HAVE_INDEX)
-        name_save = bytestring(name)
+        # This and the `unsafe_convert` below emulate a `ccall`
+        name_save = Base.cconvert(Ptr{UInt8}, name)
         t = new(0,C_NULL,0,
                 unsafe_convert(Ptr{UInt8}, name_save),
                 convert(Int, basicsize), 0,
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -24,6 +24,7 @@ import Base: sigatomic_begin, sigatomic_end
 
 ## Compatibility import for v0.4, v0.5
 using Compat
+import Compat.String
 import Base.unsafe_convert
 
 #########################################################################
@@ -286,7 +287,7 @@ keys(o::PyObject) = Symbol[m[1] for m in pycall(inspect["getmembers"],
 # and providing access to o's members (converted to PyAny) as w.member.
 
 # we skip wrapping Julia reserved words (which cannot be type members)
-const reserved = Set{ASCIIString}(["while", "if", "for", "try", "return", "break", "continue", "function", "macro", "quote", "let", "local", "global", "const", "abstract", "typealias", "type", "bitstype", "immutable", "ccall", "do", "module", "baremodule", "using", "import", "export", "importall", "pymember", "false", "true", "Tuple"])
+const reserved = Set{String}(["while", "if", "for", "try", "return", "break", "continue", "function", "macro", "quote", "let", "local", "global", "const", "abstract", "typealias", "type", "bitstype", "immutable", "ccall", "do", "module", "baremodule", "using", "import", "export", "importall", "pymember", "false", "true", "Tuple"])
 
 """
     pywrap(o::PyObject)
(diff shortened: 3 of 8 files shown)


REQUIRE
```diff
@@ -1,4 +1,4 @@
 julia 0.4
-Compat 0.7.14
+Compat 0.7.14+
```
stevengj: Presumably we should wait until the necessary version of Compat is tagged?

src/pytype.jl
```diff
@@ -261,7 +261,7 @@ type PyTypeObject
     # Julia-specific fields, after the end of the Python structure:
 
     # save the tp_name Julia string so that it is not garbage-collected
-    tp_name_save::ASCIIString
+    tp_name_save # This is a pointer field that is never read from
```
stevengj: This is not correct.  It is not a raw pointer field, it is a Julia object, stored to prevent garbage-collection.

yuyichao: That is exactly what a pointer field mean.

src/pytype.jl
```diff
@@ -280,7 +280,7 @@ type PyTypeObject
              Py_TPFLAGS_HAVE_CLASS |
              Py_TPFLAGS_HAVE_STACKLESS_EXTENSION |
              Py_TPFLAGS_HAVE_INDEX)
-        name_save = bytestring(name)
+        name_save = Base.cconvert(Ptr{UInt8}, name)
```
stevengj: Should be `String(name)`.   The `name` argument is an `AbstractString`, and could potentially be some other string type like `UTF16String`, and needs to be converted to UTF-8.

REQUIRE
yuyichao: Correct. I'm just not sure what's the status of https://github.com/JuliaLang/Compat.jl/pull/192 .

src/pytype.jl
yuyichao: This is the `ccall` interface. The code here is exactly the same as if you write `ccall(..., (..., Ptr{UInt8}, ...,), ..., name, ...)`

src/pytype.jl
stevengj: The point is that it is a `jl_value_t*`, not a `Ptr{UInt8}`.

src/pytype.jl
```diff
@@ -280,7 +280,7 @@ type PyTypeObject
...
         t = new(0,C_NULL,0,
                 unsafe_convert(Ptr{UInt8}, name_save),
```
stevengj: This could be just `pointer(name_save)`

src/pytype.jl
yuyichao: Correct. This is also what `!T.pointerfree` means.

src/pytype.jl
yuyichao: Not if you want to support `AbstractString`.

src/pytype.jl
stevengj: I'm not sure what you're talking about.  CPython never looks at the `tp_name_save` field; this is is something I tacked on to the end of the struct so that the Julia-side object would cache a reference to the string to make sure it is not garbage collected for as long as I keep the Julia type object around (forever).

src/pytype.jl
stevengj: `name` is an `AbstractString`.  `name_save = String(name)` is a UTF-8 string.

yuyichao:
General comment about passing julia object to C.
If you want to pass an julia object `v` to C in the same format it would have been passed in `ccall` as an argument of type `T` with explicit control of the object lifetime, the correct way to do this is (with the exception if `T` is `Any` since it is handled specially),

```
root = Base.cconvert(T, v) # this object controls the lifetime
cobj = Base.unsafe_convert(T, v) # this is the C value
```

The name of the `pointer field` comes from `pointerfree`. I can rename it to GC root if it's more clear.

stevengj:
Oh, right.   The `cconvert` function is very confusingly named, because I keep expecting it to act like `convert`; it's weird that `cconvert(T, v)` does not produce an object of type `T`.   Maybe `gcroot` would be a more descriptive name.

yuyichao:
> Maybe `gcroot` would be a more descriptive name.

`gc_convert` ;-p (not sure if it actually worth changing the name though....)

Another note is that `pointer` should generally be avoided whenever possible. It is usually only necessary for interacting with C code (ok, I admit I've used it to workaround LLVM vectorization issue......) and in such case the `cconvert`, `unsafe_convert` is much safer (despite possibly confusing names.....) and should automatically happen for `ccall`.

yuyichao:
> It's weird that `cconvert(T, v)` does not produce an object of type `T`.

It's true that it can return whatever type it feels like as long as it keeps the result of `unsafe_convert` on the same type alive. This indeed doesn't sound like what a `*convert` should do (except that it is part of the full conversion (`cconvert` + `unsafe_convert`) to a C object) so I now see why it is confusing.......... (started in 0.4 time, I've never seen the old interface and have naturally accepted the new one....)

yuyichao:
Updated.

There is one API change [here](https://github.com/stevengj/PyCall.jl/pull/267/files#diff-9bc274dbfe972452e9acafe692c9fd38R705). I take [this](https://github.com/JuliaLang/Compat.jl/pull/197#discussion_r62839598) as it doesn't matter....

stevengj:
Yeah, using `ByteString` there on 0.5 should be fine.

stevengj:
Travis failure:
```
ERROR: LoadError: PyError (:PyObject_Call) <type 'exceptions.TypeError'>
TypeError('type() argument 1 must be string, not unicode',)
```
Probably the string conversion needs to be:
```jl
function PyObject(s::AbstractString)
    sb = String(s)
    if pyunicode_literals || !isascii(sb)
        PyObject(@pycheckn ccall(@pysym(PyUnicode_DecodeUTF8),
                                 PyPtr, (Ptr{UInt8}, Int, Ptr{UInt8}),
                                 sb, sizeof(sb), C_NULL))
    else
        PyObject(@pycheckn ccall(@pysym(PyString_FromStringAndSize),
                                 PyPtr, (Ptr{UInt8}, Int), sb, sizeof(sb)))
    end
end
```
and you can get rid of `PyObject(::String)`.

yuyichao:
............ python unicode coupled with julia string change ...........

Yeah, I only tested on python3.... I'll fix it soon.

yuyichao:
Done

yuyichao:
ping.

stevengj:
Thanks!

PR #272, opened 2016-05-17
Fix #271

### LINKED ISSUE #271: v1.5.0 tag broke Plots.jl inside IJulia
See https://github.com/tbreloff/Plots.jl/issues/258. It looks like Plots.jl renders figures in the notebook in a slightly different way to PyPlot. PyPlot still works fine but Plots is hitting on a problem with `flush()`, which was changed in https://github.com/stevengj/PyCall.jl/commit/453100f810fe51292a6d5117a89d6ee69191ca7d.

cstjean:
When I translated the IO code to `@pydef`, I neglected to bring in that check from the `jl_io_flush` function.

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -82,7 +82,14 @@ function pyio_initialize()
             closed.get(self) = @with_ioraise(!isopen(pyio_jl(self)))
             encoding.get(self) = "UTF-8"
             fileno(self) = @with_ioraise(fd(pyio_jl(self)))
-            flush(self) = @with_ioraise(flush(pyio_jl(self)))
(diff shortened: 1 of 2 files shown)


stevengj:
Add a test case using `PyObject(IOBuffer())[:flush]()`?

cstjean:
Like this?

stevengj:
Great, thanks.

stevengj:
I've added you as a collaborator; feel free to merge when the tests are green.

stevengj:
And feel free to tag as a new patch release (1.6.1, I think).

PR #279, opened 2016-05-24
eliminate "global const foo = bar" from __init__

stevengj:
This eliminates the `global const` assignments from `__init__`, since this is deprecated and doesn't play well with precompilation (JuliaLang/julia#12010).


cstjean:
I don't know enough about PyCall for a detailed review, but this looks OK. 

I wouldn't be surprised if some libraries refer `PyCall.pynothing`... I know it's not exported, but maybe it should be.

stevengj:
According to a grep of all registered packages, no one is using `pynothing` outside of PyCall itself.

PR #287, opened 2016-06-12
Fixed eltype bug.

rasmushenningsson:
eltype wasn't defined for ::Type{PyDict{K,V}}

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -429,7 +429,7 @@ keys{T}(::Type{T}, d::PyDict) = convert(Vector{T}, d.isdict ? PyObject(@pycheckn
 values{T}(::Type{T}, d::PyDict) = convert(Vector{T}, d.isdict ? PyObject(@pycheckn ccall((@pysym :PyDict_Values), PyPtr, (PyPtr,), d)) : pycall(d.o["values"], PyObject))
 
 similar{K,V}(d::PyDict{K,V}) = Dict{pyany_toany(K),pyany_toany(V)}()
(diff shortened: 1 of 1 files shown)


src/conversions.jl
```diff
@@ -429,7 +429,7 @@ keys{T}(::Type{T}, d::PyDict) = convert(Vector{T}, d.isdict ? PyObject(@pycheckn
 values{T}(::Type{T}, d::PyDict) = convert(Vector{T}, d.isdict ? PyObject(@pycheckn ccall((@pysym :PyDict_Values), PyPtr, (PyPtr,), d)) : pycall(d.o["values"], PyObject))
 
 similar{K,V}(d::PyDict{K,V}) = Dict{pyany_toany(K),pyany_toany(V)}()
-eltype{K,V}(a::PyDict{K,V}) = Pair{pyany_toany(K),pyany_toany(V)}
```
cstjean: Shouldn't you keep the existing definition? It covers `PyDict` instances, whereas the line you added covers `PyDict` types.

rasmushenningsson: There is a fallback definition in operators.jl:
eltype(x) = eltype(typeof(x))

Since it is not needed, I left it out, but I could of course add it in again.

stevengj:
Thanks, could you add a test?

rasmushenningsson:
Definitely. I'm currently traveling, but I'll do it as soon as possible.

musm:
bump

stevengj:
I'll add a test separately.

PR #292, opened 2016-07-02
Fix incorrect ccall argument type

yuyichao:
`Array` as argument type means passing the array by reference, not passing the pointer to a string. This is what's causing https://github.com/JuliaLang/julia/issues/15432 and https://github.com/stevengj/PyCall.jl/issues/254#issuecomment-206515726 (plus incorrect handling in the ABI).


stevengj:
Whoops, this was a typo!  Thanks!

PR #293, opened 2016-07-07
Allow using PyCall with in-process symbols

Keno:
This is the companion pull request to https://github.com/JuliaLang/pyjulia/pull/54 in order to allow PyCall to use the python symbols already present in the current process. This behavior is currently guarded by the `PYCALL_CURRENT_PROCESS` environment variable. It may be worth enabling the codepath even when that variable is not set, but my thinking was that since a good amount of special handling is required to make this work properly, it should only be enabled if explicitly requested.


deps/build.jl
```diff
@@ -122,27 +122,64 @@ end
...
+            ccall(:_NSGetExecutablePath, Cint, (Ptr{UInt8}, Ptr{UInt32}), C_NULL, size)
+            path = Vector{UInt8}(size[])
+            if ccall(:_NSGetExecutablePath, Cint, (Ptr{UInt8}, Ptr{UInt32}), path, size) != 0
+                error("Failed to determine executable path")
+            end
+            path = String(path[1:size[]-1])
+            islink(path) && (path = readlink(path))
+            path
+        else
+            error("Unsupported operating system")
```
tkelman: does `/proc/self/exe` work on freebsd etc?

Keno: See http://stackoverflow.com/questions/1023306/finding-current-executables-path-without-proc-self-exe

deps/build.jl
```diff
@@ -122,27 +122,64 @@ end
 # need to be able to get the version before Python is initialized
 Py_GetVersion(libpy) = unsafe_string(ccall(Libdl.dlsym(libpy, :Py_GetVersion), Ptr{UInt8}, ()))
 
-const python = try
-    let py = get(ENV, "PYTHON", isfile("PYTHON") ? readchomp("PYTHON") : "python"), vers = convert(VersionNumber, pyconfigvar(py,"VERSION","0.0"))
-        if vers < v"2.7"
-            error("Python version $vers < 2.7 is not supported")
+# If the PYCALL_CURRENT_PROCESS is set, look for python symbols in the current
+# process first (e.g. for use from pyjulia).
+if get(ENV, "PYCALL_CURRENT_PROCESS", nothing) !== nothing
```
tkelman: `haskey` ?

Keno: fair point

src/PyCall.jl
```diff
@@ -37,14 +37,32 @@ const depfile = joinpath(dirname(@__FILE__), "..", "deps", "deps.jl")
...
 
-macro pysym(func)
-    :(($func, libpython))
-end
-macro pyglobal(name)
-    :(cglobal(($name, libpython)))
-end
-macro pyglobalobj(name)
-    :(cglobal(($name, libpython), PyObject_struct))
+if libpython == "executable"
```
stevengj: Maybe use a string that is not a valid filename, e.g. `"\0executable"` ?

stevengj: Or just `const libpython = nothing`?

deps/build.jl
stevengj: Can't we just check whether the current process has the `Py_GetVersion` symbol, rather than using an environment variable?

Keno:
Ok, I have got pycall basically working with this. There are two remaining (and related questions). They are probably more pyjulia questions, but let's discuss them here, since they affect how this PR will evolve.

The first most immediate problem is that deps.jl should be separate for pyjulia from the one for julia itself. Right now, one overrides the other which is annoying. Relatedly, I imagine people might want to use pyjulia from different python installations (e.g. python2/python3). In that case, we'd probably want to use the regular `.ji` files if your python matches the one selected by pyjulia, but otherwise create a separate `.ji` directory elsewhere. The question is where. In JuliaLang/pyjulia#54 I currently have one global such directory, but that obviously doesn't work if people want to use pyjulia with multiple versions of python. Does python have a directory where it is customary to put such things? Should we just version the directories by python version as well as julia version (e.g. `pyjulia2.7-v0.5-static`). Other suggestions?
Further, may it make sense to put the deps.jl file in the same directory as the `.ji` file since they're essentially tied together?

stevengj:
Maybe in `Pkg.dir("PyCall", "deps", "pythonX.Y")`?  It's a lot easier to work with a subdirectory of `deps` than some other random location in the directory tree, and that way it stays in sync with the PyCall version being used.   (e.g. it will continue to work if someone makes a private clone of PyCall).

stevengj:
As for pyjulia's `.ji` file, I agree that this should also be in the same directory (under `deps/pyjulia/pythonX.Y`), I think.

Keno:
That sounds good to me. Should there be an environment variable that determines which subdirectory to use?

Keno:
Well, the problem is that it has to create `.ji` files for all packages, not just PyCall, so having the `.ji` files under PyCall is a little weird.

stevengj:
I hate using environment variables for something like this.  Can't `PyCall.jl` just check whether the current process has the `Py_GetVersion` symbol, as I suggested above, and if it does it can assume it is in `pyjulia` and call `Py_GetVersion` to determine the correct directory.

Keno:
Yes, it can. I was about to say you'd pay a cost at startup time, but that's only once during precompilation,  so I think it's fine.

tkelman:
> it has to create .ji files for all packages

Why is this?

Keno:
Because it has to do it for PyCall and all packages that depend on it and there's no way to tell julia to only do it for those packages.

tkelman:
I'd be a bit surprised if someone needed something that depends on PyCall from inside pyjulia.

Keno:
PyPlot?

tkelman:
Why would you go through Julia for that if you're embedding Julia inside Python?

Keno:
You wouldn't, but you might be using a package which has a dependency on it (quite a few packages do).

tkelman:
Should we try to do this in a way that the precompile system can recognize which configuration PyCall is in and just recompile the dependency ji's in place if the config changes?

Keno:
I'm open to any suggestions.

Keno:
I suppose since we're intercepting the precompile calls anyway, we could check if PyCall was loaded and if so change the output path. Devious but I somewhat like it. @vtjnash ?

tkelman:
I'd say something like a dummy include_dependency that switched between two choices depending whether we're in in-process or shared library mode. Probably set a module level constant too so it can be checked.

Keno:
@vtjnash is saying the general solution is overly complicated, but intercepting the precompile call and hardlinking should work fine.

Keno:
After ruminating on this some more, this is how I think I want this to work:

PyCall checks if Python symbols are present in the process and if so does not load any deps.jl, but figures out where the python library is, by looking at where the symbol came from (or alternatively using the python APIs to query this).

Then pyjulia does the following:

- If deps.jl is present and the versions declared therein match the currently running python process (same version, dynamically linked), pyjulia just does what it has always done, which is load PyCall without any special precompile flags or anything.
- If deps.jl is absent, or the versions don't match, it intercepts precompilation, sets up a separate cache directory and uses that. Since PyCall will prefer local symbols over deps.jl and we intercepted precompilation, this will create a working cache file. Slight caveat is that while precompiling PyCall only the separate cache directory can be on the load path, but thereafter both can be.

In the second case we also use the interception trick to figure out if PyCall was loaded and if not put the cache file in the original place (rather than the pyjulia private place) and simply symlink, thereby sharing cache files for packages that do not depend on PyCall.

stevengj:
That sounds workable.

Keno:
Ok, I've implemented that approach here and in pyjulia. Seems to work decently well. Needs a bit more testing and cleanup.

Keno:
@stevengj Are you ok with this generally? If so I'll rebase and clean a bit, after which this should be good to go.

stevengj:
LGTM.

Keno:
rebased.

tkelman:
be sure to bump at least minor when tagging this, there's a chance it may have unintended consequences for dependent packages (hopefully not but won't know until after it gets tagged)

stevengj:
Yup, it's a new feature anyway so it should get a minor bump.

PR #294, opened 2016-07-08
Work around Windows Activation Context problems

### LINKED ISSUE #87: pyimport results in Runtime Error
When answering [a qustion](http://stackoverflow.com/questions/25155205/ijulia-notebook-ipython-files) on stack overflow I noticed that if I import a part of IPython with `@pyimport` up pops a alert windows alerting of a runtime error. 
```
using PyCall
@pyimport IPython.nbformat.current as ipyt
```
The alert windows says 
```
Runtime Error!

Program C:\Users\admin\AppData\Local\Julia-0.3.0-rc2\bin\julia.exe

R6034
An application has made an attempt to load the C runtime library incorrectly.
Please contact the application's support team for more information.
```
Interestingly enough the import still works and I can use `ipyt` to access the notebook parsing function of IPython.

This is on Windows 8.1 Pro 64 bit.
```
versioninfo()

Julia Version 0.3.0-rc2+12
Commit 3ddbaa1 (2014-08-06 22:29 UTC)
Platform Info:
  System: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i5-2500K CPU @ 3.30GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: libopenblas
  LIBM: libopenlibm
  LLVM: libLLVM-3.3
```
```
Pkg.status()

20 required packages:
 - ApproxFun                     0.0.2
 - GLFW                          1.0.0-alpha.3
 - GLM                           0.4.0
 - GLPK                          0.2.11
 - GLPKMathProgInterface         0.1.5
 - Gadfly                        0.3.4+             master
 - HypothesisTests               0.2.7
 - IJulia                        0.1.12
 - Images                        0.2.47
 - Jewel                         0.6.4
 - JuMP                          0.5.5
 - MixedModels                   0.3.9
 - Optim                         0.3.1
 - PValueAdjust                  1.0.0
 - RDatasets                     0.1.1
 - Resampling                    0.0.0
 - Rmath                         0.0.0
 - SIUnits                       0.0.2
 - TypeCheck                     0.0.3
 - Winston                       0.11.2
59 additional packages:
 - ArrayViews                    0.4.6
 - BinDeps                       0.2.14
 - Cairo                         0.2.15
 - Calculus                      0.1.4
 - Cartesian                     0.2.2
 - Codecs                        0.1.2
 - Color                         0.2.11
 - Compose                       0.3.4+             master
 - Contour                       0.0.2
 - DataArrays                    0.2.0
 - DataFrames                    0.5.7
 - DataStructures                0.3.1
 - Datetime                      0.1.6
 - Distance                      0.4.0
 - Distributions                 0.5.4
 - DualNumbers                   0.1.0
 - ForwardDiff                   0.0.2
 - GLAbstraction                 0.0.0-             master (unregistered)
 - GLPlot                        0.0.0-             master (unregistered)
 - GLWindow                      0.0.0-             master (unregistered)
 - GZip                          0.2.13
 - Graphs                        0.4.3
 - HTTPClient                    0.1.4
 - Hexagons                      0.0.2
 - ImmutableArrays               0.0.6
 - IniFile                       0.2.2
 - Iterators                     0.1.6
 - JSON                          0.3.7
 - KernelDensity                 0.0.2
 - Lazy                          0.4.1
 - LibCURL                       0.1.3
 - LibExpat                      0.0.4
 - Loess                         0.0.2
 - MathProgBase                  0.2.4
 - ModernGL                      0.0.0-             master (unregistered)
 - Mustache                      0.0.3
 - NLopt                         0.1.1
 - Nettle                        0.1.4
 - NeuralNets                    0.0.0-             master (unregistered)
 - NumericExtensions             0.6.2
 - NumericFuns                   0.2.3
 - Options                       0.2.2
 - PDMats                        0.2.3
 - PyCall                        0.4.8
 - PyPlot                        1.2.10
 - Quaternions                   0.0.4
 - REPLCompletions               0.0.1
 - React                         0.1.3
 - Reexport                      0.0.1
 - ReverseDiffSparse             0.1.2
 - SortingAlgorithms             0.0.1
 - StatsBase                     0.6.3
 - TexExtensions                 0.0.1
 - Tk                            0.2.13
 - URIParser                     0.0.2
 - URLParse                      0.0.0
 - WinRPM                        0.1.2
 - ZMQ                           0.1.13
 - Zlib                          0.1.7
```

Keno:
For a detailed explanation of the problems, see https://github.com/zeromq/pyzmq/issues/861.
Fixes #87.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -334,14 +334,63 @@ end
 
 #########################################################################
 
+if is_windows()
+    # Many python extensions are linked against a very specific version of the
+    # MSVC runtime library. To load this library, libpython declares an
+    # appropriate manifest, but unfortunately most extensions do not.
+    # Libpython itself does extend its activation context to any extensions it
+    # loads, but some python libraries (e.g. libzmq), load such extensions
+    # through ctypes, which does not have this functionality. Work around
+    # this by manually activating the activation context before any call to
+    # PyImport_ImportModule, since extensions are most likely to be loaded
+    # during import.
+
+    immutable ACTIVATION_CONTEXT_BASIC_INFORMATION
+        handle::Ptr{Void}
+        dwFlags::UInt32
+    end
+    const ActivationContextBasicInformation = 1
+    const QUERY_ACTCTX_FLAG_ACTCTX_IS_ADDRESS = 0x10
+
+    function ActivatePyActCtx()
+        some_address_in_libpython = @pyglobal(:PyImport_ImportModule)
+        ActCtxBasicInfo = Ref{ACTIVATION_CONTEXT_BASIC_INFORMATION}()
+        succeeded = ccall(:QueryActCtxW,stdcall,Bool,
+            (UInt32,Ptr{Void},Ptr{Void},Culong,Ptr{Void},Csize_t,Ptr{Csize_t}),
+            QUERY_ACTCTX_FLAG_ACTCTX_IS_ADDRESS, some_address_in_libpython,
+            C_NULL, ActivationContextBasicInformation, ActCtxBasicInfo,
+            sizeof(ActCtxBasicInfo), C_NULL)
+        @assert succeeded
+        cookie = Ref{Ptr{Void}}()
+        succeeded = ccall(:ActivateActCtx,stdcall, Bool,
+                          (Ptr{Void}, Ptr{Ptr{Void}}),
+                          ActCtxBasicInfo[].handle, cookie)
+        @assert succeeded
+        cookie[]
+    end
+
+    function DeactivatePyActCtx(cookie)
+        succeeded = ccall(:DeactivateActCtx, stdcall, Bool,
+                         (UInt32, Ptr{Void}), 0, cookie)
+        @assert succeeded
+    end
+else
+    ActivatePyActCtx() = nothing
+    DeactivatePyActCtx(cookie) = nothing
+end
+
 """
     pyimport(s::AbstractString)
 
 Import the Python module `s` (a string or symbol) and return a pointer to it (a `PyObject`). Functions or other symbols in the module may then be looked up by s[name] where name is a string (for the raw PyObject) or symbol (for automatic type-conversion). Unlike the @pyimport macro, this does not define a Julia module and members cannot be accessed with `s.name`
 """
-pyimport(name::AbstractString) =
-    PyObject(@pycheckn ccall((@pysym :PyImport_ImportModule), PyPtr,
+function pyimport(name::AbstractString)
+    cookie = ActivatePyActCtx()
+    obj = PyObject(@pycheckn ccall((@pysym :PyImport_ImportModule), PyPtr,
                              (Cstring,), name))
+    DeactivatePyActCtx(cookie)
+    obj
+end
 pyimport(name::Symbol) = pyimport(string(name))
 
 # convert expressions like :math or :(scipy.special) into module name strings
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -351,20 +351,28 @@ if is_windows()
     end
     const ActivationContextBasicInformation = 1
     const QUERY_ACTCTX_FLAG_ACTCTX_IS_ADDRESS = 0x10
+    const PyActCtx = Ref{Ptr{Void}}(C_NULL)
+
+    function ComputePyActCtx()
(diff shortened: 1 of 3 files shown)


src/PyCall.jl
```diff
@@ -334,14 +334,63 @@ end
...
+    # through ctypes, which does not have this functionality. Work around
+    # this by manually activating the activation context before any call to
+    # PyImport_ImportModule, since extensions are most likely to be loaded
+    # during import.
+
+    immutable ACTIVATION_CONTEXT_BASIC_INFORMATION
+        handle::Ptr{Void}
+        dwFlags::UInt32
+    end
+    const ActivationContextBasicInformation = 1
```
stevengj: The [QueryActCtxW docs](https://msdn.microsoft.com/en-us/library/aa375704(v=vs.85).aspx) list this constant as "not available" ... ?

Keno: It's lying.

stevengj:
Thanks for doing this, @Keno.   I'm in favor of merging this.

Is `ActivatePyActCtx` expensive compared to a typical `pyimport`?  If I have to do a whole bunch of `pyimport` calls, would it be better to have a way to do this once?  I'm hoping it is cheap enough that I can not worry.

Keno:
> Is ActivatePyActCtx expensive compared to a typical pyimport? If I have to do a whole bunch of pyimport calls, would it be better to have a way to do this once? I'm hoping it is cheap enough that I can not worry.

We could cache the activation context lookup, which is the expensive part. Activating the context itself should be very cheap. Windows basically does it on every call to a dll that has a different context from the calling one.

stevengj:
Can you add a test?   Maybe change [this test](https://github.com/stevengj/PyCall.jl/blob/ee0897b94268d3d380bfc810b08fa666f01a8f88/test/runtests.jl#L274-L284) to use `pyzmq` instead of `conda-api`?

stevengj:
Caching the lookup sounds like a good idea if it is expensive.

Keno:
I'm not sure if it is expensive, but certainly more expensive than the actual activation. I'll add code to do the caching and adjust the test as you suggested.

src/pyinit.jl
```diff
@@ -47,6 +47,9 @@ function __init__()
             end
         end
         ccall((@pysym :Py_InitializeEx), Void, (Cint,), 0)
+
+        # Will get reinitialized properly on first use
+        is_windows() && (PyActCtx[] = C_NULL)
```
stevengj: Shouldn't this line be after the `end`, i.e. outside the `if !already_inited` block?

test/runtests.jl
```diff
@@ -274,11 +274,11 @@ end
 import Conda
 if PyCall.conda
     # install the tiny conda-api package as a test
```
stevengj: comment needs to be changed: `# import pyzmq to test PR #294`.

Will importing `pyzmq` cause problems if we run `Pkg.test("PyCall")` inside IJulia or in any other context where `ZMQ.jl` has already loaded and initialized `libzmq`?

src/pyinit.jl
Keno: Yes

Keno:
I tested on Linux and there was no problem with both pyzmq and ZMQ being loaded. I would test on Windows, but I can't get WinRPM to work. If it becomes a problem we can disable the tests if we find ZMQ already loaded.

src/pyinit.jl
```diff
@@ -47,6 +47,7 @@ function __init__()
             end
         end
         ccall((@pysym :Py_InitializeEx), Void, (Cint,), 0)
+
```
stevengj: blank line

src/pyinit.jl
```diff
@@ -114,4 +115,7 @@ function __init__()
             end
         end
     end
+
+    # Will get reinitialized properly on first use
+    is_windows() && (PyActCtx[] = C_NULL)
```
stevengj: It seems like this has to go at the top of `__init__`, not at the bottom.   (Note that `__init__` calls `pyimport` several times.)

Keno: Fair point.

Keno:
Good to go?

PR #297, opened 2016-07-22
0.3 support was dropped.

PallHaraldsson:
[skip ci]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 python:
-    - 2.6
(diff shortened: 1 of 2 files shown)


stevengj:
2.6 is not supported.  The Python 2.6 tests in Travis/AppVeyor are actually testing Conda (i.e. it notices that Python is too old, and installs the Conda Python 2.7 instead).

PallHaraldsson:
Yes.. forgot.. fixed [if comment ok.. And 2.6 or "26" actually not in appveyor.]

Issue #295, opened 2016-07-13
PyCall breaks due to lack of wstring()


i-apellaniz:
Installing `LegacyString` to be able to make use of `wstring` with `using LegacyString` doesn't work.

Error message:
```
ERROR: LoadError: LoadError: `wstring` has been moved to the package LegacyStrings.jl:
Run Pkg.add("LegacyStrings") to install LegacyStrings on Julia v0.5-;
Then do `using LegacyStrings` to get `wstring`.

 in include_from_node1(::String) at ./loading.jl:426 (repeats 2 times)
 in macro expansion; at ./none:2 [inlined]
 in anonymous at ./<missing>:?
 in eval(::Module, ::Any) at ./boot.jl:234
 in process_options(::Base.JLOptions) at ./client.jl:243
 in _start() at ./client.jl:322
while loading /home/iapellaniz/.julia/v0.5/PyCall/src/../deps/deps.jl, in expression starting on line 3
while loading /home/iapellaniz/.julia/v0.5/PyCall/src/PyCall.jl, in expression starting on line 38
ERROR: Failed to precompile PyCall to /home/iapellaniz/.julia/lib/v0.5/PyCall.ji
 in compilecache(::String) at ./loading.jl:505
 in require(::Symbol) at ./loading.jl:364
 in eval(::Module, ::Any) at ./boot.jl:234
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```

Version info:
```
Julia Version 0.5.0-dev+5345
Commit adbd3ac (2016-07-12 07:29 UTC)
Platform Info:
  System: Linux (x86_64-redhat-linux)
  CPU: Intel(R) Core(TM) i5-3230M CPU @ 2.60GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, ivybridge)
```

stevengj:
Should work now after you run `Pkg.build("PyCall")`

i-apellaniz:
Thanks for your rapid response. I have had to tweak the julia code to make it work. More details will come if necesary

PR #302, opened 2016-08-02
Add LegacyStrings for julia 0.5 compatability

davidanthoff:
This should make things run on current julia 0.5 builds.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,3 +2,4 @@ julia 0.4
 Compat 0.8.0
 Conda 0.2
 MacroTools 0.2
+LegacyStrings
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -30,6 +30,7 @@ import Compat.String
 @compat import Base.show
 import Base.unsafe_convert
 import MacroTools   # because of issue #270
+using LegacyStrings
 
 #########################################################################
 


davidanthoff:
I just realized that the travis config still relies on the staticfloat PPA, which is not maintained and won't give us a recent julia 0.5 build. I think that explains the CI errors. I'll try to figure out what to do about the travis stuff.

stevengj:
This should not be needed with PyCall master, thanks to cd80a6e5fac0e3e70f72fb9bef87bfdbf354ab6b

stevengj:
See also #295

davidanthoff:
Ah, I had tried ``master`` but hadn't run ``Pkg.build``.

davidanthoff:
@stevengj Could you tag a new version? Would be really helpful to have a tagged version that runs on 0.5 RC0 so that downstream packages can more easily be brought in shape for 0.5. If that version still has deprecation warnings, no problem, but having something that doesn't error out would be great.

stevengj:
Yeah, I'm planning on tagging new versions of a bunch of my packages in the next couple of days.

davidanthoff:
Cool. I think PyCall especially would help. The downstream packages really need a tagged version that works on travis and appveyor to get their julia 0.5 porting working. I actually don't need a version that doesn't throw deprecation warnings, from my point of view just a tag on the current master would already help a great deal. The deprecation warnings could be fixed later, they don't really stop downstream packages from getting their stuff in order.

stevengj:
JuliaLang/METADATA.jl#5793

PR #312, opened 2016-08-22
Test on 3.5

PallHaraldsson:
[3.5 is latest supported Python (I assume PyCall/Travis etc. works with it), 3.3 is no longer officially supported.]

You may also want to skip: "In Julia 0.3, sometimes calling a Python function fails"..

to simplify the README.. as no longer supported (and I assume Julia 0.3 itself about to be unsupported (or used), after 0.5 is released).

[ci skip]

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,12 +7,12 @@ environment:
     PYTHON: "C:\\Python27\\python.exe"
 
   - JULIAVERSION: "julialang/bin/winnt/x86/0.4/julia-0.4-latest-win32.exe"
-    PYTHONDIR: "C:\\Python33"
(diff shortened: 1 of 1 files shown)


stevengj:
I don't see any reason to drop 3.3 testing, unless AppVeyor drops it.  Adding 3.5 testing is fine.

PallHaraldsson:
Ok, I thought it might be a burden. I can add it back if you think four tests are good/needed (maybe 3.5 testing is overkill?).

What about the README text for 0.3? Do you want me to simplify at the same time?

stevengj:
No, let's leave the README as-is for now; in any case, it would be a separate commit.

PR #316, opened 2016-08-27
Doc: typo

PallHaraldsson:
[skip ci]


PallHaraldsson:
I just notices also: "define Python classe", I guess "classes".

PR #318, opened 2016-09-05
add suppport for Float16

maximsch2:
Support Float16 arrays as arrays with dtype=np.float16

### DIFF
--- a/src/numpy.jl
+++ b/src/numpy.jl
@@ -123,6 +123,7 @@ const NPY_OBJECT = Int32(17)
 const NPY_STRING = Int32(18)
 const NPY_UNICODE = Int32(19)
 const NPY_VOID = Int32(20)
+const NPY_HALF = Int32(23)
 
 # NPY_ORDER:
 const NPY_ANYORDER = Int32(-1)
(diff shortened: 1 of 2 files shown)


src/numpy.jl
```diff
@@ -124,6 +124,8 @@ const NPY_STRING = Int32(18)
 const NPY_UNICODE = Int32(19)
 const NPY_VOID = Int32(20)
 
+const NPY_HALF = Int32(23)
```
stevengj: No blank line before this.

stevengj:
Needs a test in runtests.jl, e.g.
```jl
if PyCall.npy_initialized
   @test roundtripeq(Float16[17 18])
   @test isa(roundtrip(Float16[17]), Vector{Float16})
end
```

src/numpy.jl
```diff
@@ -154,13 +156,14 @@ npy_type(::Type{Int32}) = NPY_INT
...
 npy_type(::Type{Float64}) = NPY_DOUBLE
 npy_type(::Type{Complex64}) = NPY_CFLOAT
 npy_type(::Type{Complex128}) = NPY_CDOUBLE
 npy_type(::Type{PyPtr}) = NPY_OBJECT
 
-typealias NPY_TYPES Union{Bool,Int8,UInt8,Int16,UInt16,Int32,UInt32,Int64,UInt64,Float32,Float64,Complex64,Complex128,PyPtr}
+typealias NPY_TYPES Union{Bool,Int8,UInt8,Int16,UInt16,Int32,UInt32,Int64,UInt64,Float16,Float32,Float64,Complex64,Complex128,PyPtr}
 
 # conversions from __array_interface__ type strings to supported Julia types
 const npy_typestrs = Dict( "b1"=>Bool,
```
stevengj: You also need an entry `"f2"=>Float16` in `npy_typestrs`, right (in order for conversions in the other direction to work)?

maximsch2:
Thanks for a quick response!

I've addressed the comments.

src/numpy.jl
```diff
@@ -170,7 +172,8 @@ const npy_typestrs = Dict( "b1"=>Bool,
                            "i8"=>Int64,     "u8"=>UInt64,
                            "f4"=>Float32,   "f8"=>Float64,
                            "c8"=>Complex64, "c16"=>Complex128,
-                           "O"=>PyPtr, "O$(div(Sys.WORD_SIZE,8))"=>PyPtr )
+                           "O"=>PyPtr, "O$(div(Sys.WORD_SIZE,8))"=>PyPtr,
+                           "f2"=>Float16)
```
stevengj: You should put this on the line with `f4` and `f8`.

(In general, when you make a patch, you want to make it blend in with the surrounding code, as if it were all written at one time by the same person.)

test/runtests.jl
```diff
@@ -282,3 +282,10 @@ if PyCall.conda
         end
     end
 end
+
+# Float16 support:
+
```
stevengj: no blank line

src/numpy.jl
maximsch2: I was thinking that it will look weird as all the rest "f"s and "u"s and "i"s were paired, up but I think it looks fine now that I've moved "f2" up there.

PR #324, opened 2016-10-12
Only use Conda on x86

simonbyrne:


deps/build.jl
```diff
@@ -149,9 +149,14 @@ const python = try
         py
     end
 catch e1
-    info( "No system-wide Python was found; got the following error:\n",
-          "$e1\nusing the Python distribution in the Conda package")
-    abspath(Conda.PYTHONDIR, "python" * ( is_windows() ? ".exe" : ""))
+    if Sys.ARCH in (:i686, :x86_64)
+        info( "No system-wide Python was found; got the following error:\n",
+              "$e1\nusing the Python distribution in the Conda package")
+        abspath(Conda.PYTHONDIR, "python" * ( is_windows() ? ".exe" : ""))
```
simonbyrne: that was already there, but i can fix it

PR #329, opened 2016-11-21
More flexible pyimport syntax. Fixes #66

### LINKED ISSUE #66: pyimport syntax: from <module> import <symbol>
Maybe in @pyimport macros you could add some other sintaxis, for example if the user dont need to load all the package and only one function or more of it:

```
@pyimport from numpy import zeros, array, fromtxt
@pyimport from mayavi import mlab
@pyimport from scipy.ndimage import gaussian_filter as filter
```

cstjean:


stevengj:
How does this handle "merge conflicts", i.e. if the imported function is already defined?

I have a growing sense that `@pyimport` was a mistake.  It is inefficient (because it has to convert everything in the module at import time, and doesn't work with precompilation), it tricks people into thinking that PyCall supports Pythonic dot syntax, and it is awkward (`pymember`) if the default conversion is not the one that is wanted.  I'm thinking it would would be better to deprecate `@pyimport` and just teach people to use `math = pyimport("math")` and then use `math[:pow]` etcetera.  If they want to import a single function they can do `foo = pyimport("bar")[:foo]`.

cstjean:
```julia
@pyimport math: sin
> Cannot @pyimport sin. It is already defined in the current module.
```

> I have a growing sense that @pyimport was a mistake.

I see what you mean. The new syntax doesn't rely on the `pywrap` trick, at least, when importing functions/classes.

```julia
@pyimport sklearn.linear_model: (LinearRegression, LogisticRegression)
```

I'm OK with taking it out of PyCall, but I will keep it alive elsewhere in my codebase, as I find it rather elegant.

stevengj:
We could always support
```jl
LinearRegression, LogisticRegression = pyimport("sklearn.linear_model")[:LinearRegression, :LogisticRegression]
```
by special-casing a tuple of symbols in `getindex`

cstjean:
Another option would be to unexport it and take it out of the README.

stevengj:
The migration path would be to `@deprecate` the `pywrap` function for a while, so it would continue to work but with a deprecation warning.

cstjean:
That's fine for me. I'm already providing this PR's macro in ScikitLearn.jl, because it makes my [examples](https://github.com/cstjean/ScikitLearn.jl/blob/master/examples/Plot_Kmeans_Digits.ipynb) look friendlier for people coming from Python. I can keep doing that.

stevengj:
(If Julia would just get dot overloading already then this would be mostly moot, grrr.)

PR #330, opened 2016-11-21
Fix writemime deprecation warnings

cstjean:
In IJulia, I sometimes get deprecation warnings from PyCall. Eg.

```julia
using PyCall
@pyimport math
math.sin
```

(has to be run in three different cells!)

This PR fixes this for Julia 0.5.


stevengj:
LGTM.

PR #334, opened 2016-12-11
Typo

PallHaraldsson:
[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This package provides the ability to directly call and **fully
 interoperate with Python** from [the Julia
 language](http://julialang.org/).  You can import arbitrary Python
 modules from Julia, call Python functions (with automatic conversion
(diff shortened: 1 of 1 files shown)


PallHaraldsson:
Another thing:

I thought, you can call more than functions, such as methods/objects, right? I even remembered, "fully interoperate with Python".

I thought of this at julialang.org:

"Call Python functions: use the PyCall package
Call C functions directly: no wrappers or special APIs"

Shouldn't it be updated with similar language? It doesn't apply to C but does to C++; at least good to make sure for Python, since listed.

PallHaraldsson:
Is there a good way to use Qt5 with (or without) Python?

http://pyqt.sourceforge.net/Docs/PyQt5/introduction.html#an-explanation-of-version-numbers

PR #354, opened 2017-01-25
Expose docs to Julia

### LINKED ISSUE #46: [Feature request] Integrate Python docstrings with Julia help
Not sure how implausible this is, but it would be great if typing
?[python function] 
at the julia REPL gave back the docstring of that function/module, just as IPython's ? does.

malmaud:
Fixes #46.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -708,7 +708,7 @@ for (mime, method) in ((MIME"text/html", "_repr_html_"),
 end
 
 #########################################################################
-# Expose Python docstrings to Julia doc system
+# Expose Python docstrings to the Julia doc system
 
 if isdefined(Docs, :getdoc)  # Introduced in Julia .6
     Docs.getdoc(o::PyObject) = Text(String(o["__doc__"]))
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -711,7 +711,7 @@ end
 # Expose Python docstrings to Julia doc system
 
 if isdefined(Docs, :getdoc)  # Introduced in Julia .6
-    Docs.getdoc(o::PyObject) = Markdown.parse(o[:__doc__])
+    Docs.getdoc(o::PyObject) = Text(String(o["__doc__"]))
 end
 
 
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -708,6 +708,14 @@ for (mime, method) in ((MIME"text/html", "_repr_html_"),
 end
 
 #########################################################################
+# Expose Python docstrings to Julia doc system
+
+if isdefined(Docs, :getdoc)  # Introduced in Julia .6
+    Docs.getdoc(o::PyObject) = Markdown.parse(o[:__doc__])
+end
+
+
+#########################################################################
 
 include("pyeval.jl")
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -322,3 +322,15 @@ let weakdict = pyimport("weakref")["WeakValueDictionary"]
     #  weak references to type objects)
     @test weakdict(Dict(3=>weakdict)) == Dict(3=>weakdict)
 end
+
+# Expose python docs to Julia doc system
+if isdefined(Docs, :getdoc)
+    py"""
+    def foo():
+        "foo docstring"
+        return 0
+    """
+    foo = py"foo"
+    # use 'content' since `Text` objects test equality by object identity
+    @test @doc(foo).content == "foo docstring"
+end


src/PyCall.jl
```diff
@@ -708,6 +708,14 @@ for (mime, method) in ((MIME"text/html", "_repr_html_"),
 end
 
 #########################################################################
+# Expose Python docstrings to Julia doc system
+
+if isdefined(Docs, :getdoc)  # Introduced in Julia .6
+    Docs.getdoc(o::PyObject) = Markdown.parse(o[:__doc__])
```
stevengj: Python docstrings aren't generally in Markdown format...

stevengj: See also https://github.com/coldfix/doc2md

stevengj: Though the amateur-written license for that code is problematic.

malmaud: Hmm, at least some are (https://www.tensorflow.org/versions/master/how_tos/documentation/), and the Julia markdown parser gave decent for Pandas and Seaborn. 
Should we bundle/reimplement something like doc2md?

stevengj: I don't think we can rely on that always working.  At the very least, we have to worry about escaping, since there's no reason a Python docstring couldn't be malformed markdown.

It seems like we might need to re-implement something like doc2md, at least in a limited form.

stevengj: For now, however, can't we just return the string?

(Since we know it is supposed to return a string, it would be better to do `String(o["__doc__"])` to skip the type-introspection step.)

stevengj: Ultimately, it seems like we should do something like:
```
function Docs.getdoc(o::PyObject)
    s = String(o["__doc__"])
    try
        return doc2md(s) # try to convert s to Markdown
    catch
        return s # malformed markdown
    end
end
````

Comment (stevengj) [inline src/PyCall.jl]: One thing that worries me is that docstrings may often contain lists that are supposed to be one item per line, and Markdown parsing will convert that into a single paragraph.

e.g. [PEP 257](https://www.python.org/dev/peps/pep-0257/) recommends an example:
```py
def complex(real=0.0, imag=0.0):
    """Form a complex number.

    Keyword arguments:
    real -- the real part (default 0.0)
    imag -- the imaginary part (default 0.0)
    """
    if imag == 0.0 and real == 0.0:
        return complex_zero
    ...
```
and the Markdown parsing gives:
```
julia> Markdown.parse(s)
  Form a complex number.

  Keyword arguments: real – the real part (default 0.0) imag – the imaginary
  part (default 0.0)
```

stevengj: As a result, I think we may have no choice but just to use the plain text string.  I doubt there is any reliable way to convert docstrings to good-looking Markdown.  (The doc2md package seems specialized for docstrings written in a particular way.)

stevengj: If for some reason we need a Markdown object, I think we could do:
```jl
m = Markdown.MD()
push!(md, Markdown.Code("", String(o["__doc__"]))
return m
```

malmaud: Just returning a string doesn't seem to look too great since newlines are printed literally:

```
help?> pandas.Series
"\n    One-dimensional ndarray with axis labels (including time series).\n\n    Labels need not be unique but must be any hashable type. The object\n    supports both integer- and label-based indexing and provides a host of\n    methods for performing operations involving the index. Statistical\n    methods from ndarray have been overridden to automatically exclude\n    missing data (currently represented as NaN)\n\n    Operations between Series (+, -, /, *, **) align values based on their\n    associated index values-- they need not be the same length. The result\n    index will be the sorted union of the two indexes.\n\n    Parameters\n    ----------\n    data : array-like, dict, or scalar value\n        Contains data stored in Series\n    index : array-like or Index (1d)\n        Values must be unique and hashable, same length as data. Index\n        object (or other iterable of same length as data) Will default to\n        RangeIndex(len(data)) if not provided. If both a dict and index\n        sequence are used, the index will override the keys found in the\n        dict.\n    dtype : numpy.dtype or None\n        If None, dtype will be inferred\n    copy : boolean, default False\n        Copy input data\n    "
```

stevengj: Oh right, but you can return `Text(String(o["__doc__"]))`

stevengj: The main advantage of returning `MD(Code(string))` is probably for the case where there are multiple docstrings, some of which might be markdown, since there is a `Base.Docs.catdoc(::MD...)`` method.  But since a `PyObject` will presumably only have a single docstring, `Text` seems like it should be fine.

malmaud: Ah, didn't know about `Text`. That seems good.

stevengj:
Can you add a test?  You can create a PyObject with a docstring via, e.g.:
```jl
julia> py"""
       def foo():
           "docstring of foo"
           return 0
       """

julia> py"foo"
PyObject <function foo at 0x31679c9b0>
```

malmaud:
Tests added

stevengj:
Thanks.  (We need to fix `pyeval` for Julia 0.4 or just drop 0.4 support.)

stevengj:
Hmm, the 0.4 tests pass on my machine.

PR #356, opened 2017-01-28
Fix 0.6 depwarns

yuyichao:
Test on master might fail with an error similar to

```
signal (4): Illegal instruction
while loading /home/yuyichao/projects/contrib/PyCall.jl/test/runtests.jl, in expression starting on line 318
copy at /usr/share/julia/site/v0.6/PyCall/src/numpy.jl:340
unknown function (ip: 0x7fe154c964f2)
jl_call_method_internal at /build/julia-git/src/julia-avx2/src/julia_internal.h:248
```

and that's https://github.com/JuliaLang/julia/issues/20305

I also decided to work around https://github.com/JuliaLang/julia/issues/19987 since it doesn't seems to have very high priority or trivial to fix (but trivial to workaround)....

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -33,7 +33,7 @@ convert{T<:Integer}(::Type{T}, po::PyObject) =
 
 if Sys.WORD_SIZE == 32
   convert{T<:Union{Int64,UInt64}}(::Type{T}, po::PyObject) =
-    @pycheck ccall((@pysym :PyLong_AsLongLong), T, (PyPtr,), asscalar(po))
+    @pycheck(ccall((@pysym :PyLong_AsLongLong), UInt64, (PyPtr,), asscalar(po))) % T
 end
 
 convert(::Type{Bool}, po::PyObject) =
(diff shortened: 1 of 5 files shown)


yuyichao:
The CI failure on 0.4 is caused by using the conditional comprehension syntax which exists on master.

yuyichao:
0.4 syntax error fixed.

yuyichao:
So the error doesn't occur on travis..... Maybe because numpy isn't installed? (It was triggered in a numpy specific test...)

src/conversions.jl
```diff
@@ -140,9 +140,10 @@ pyptr_query(po::PyObject) = pyisinstance(po, c_void_p_Type) || pyisinstance(po,
 # typealias PyAny Union{PyObject, Int, Bool, Float64, Complex128, AbstractString, Function, Dict, Tuple, Array}
 abstract PyAny
 
-pyany_toany(T::Type) = T
-pyany_toany(T::Type{PyAny}) = Any
-pyany_toany(T::Type{Vararg{PyAny}}) = Vararg{Any}
+function pyany_toany(T::Type)
+    T === Vararg{PyAny} ? Vararg{Any} : T
+end
+pyany_toany(::Type{PyAny}) = Any
```
stevengj: Why was this changed?

yuyichao: `Type{Vararg{PyAny}}` isn't a valid type anymore.

stevengj: Why not?

yuyichao: I don't know the full story but it was never an actual type (you can't find a value that matches this type)

stevengj: I'm confused, so why does `T === Vararg{PyAny}` still work if that is not a type?

yuyichao: The usage of it has always been strange. It is essentially only meaningful in a tuple but it's hard to only allow it there without breaking many other things. It is used to be allowed to use it just like a type though almost all types constructed this way are invalid. This is not disallowed.

It is still allowed to be used as a value and is still treated as a type for dispatch purpose even if it never actually was one. I believe the long term goal is to get rid of it or somehow being able to pass it around without dealing with this special case or generating invalid types.

malmaud:
@stevengj Any objection to tagging PyCall now that we have this?

stevengj:
Tagged 1.9

PR #361, opened 2017-02-13
0.4 not really required (even 0.3 works)

PallHaraldsson:
[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ without copying them.
 ## Installation
 
 Within Julia, just use the package manager to run `Pkg.add("PyCall")` to
-install the files.  Julia 0.4 is required.
+install the files.  Julia 0.4  (or later) is required, for better support.
 
 The latest development version of PyCall is available from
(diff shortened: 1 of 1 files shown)


PallHaraldsson:
[Possibly you meant with "is required", "at least"; good reason not to be explicit?]

I note your Conda change. I assume you dropped Python requirements as you just fetch to correct Python.

[But not on Linux (all assumed 2.7+?); Some old (still supported) RedHat/CentOS have only Python 2.6.. (or does a minor release update?) not sure if it's a problem, or worth worrying about.]

PallHaraldsson:
Badges/REQUIRE says all that is needed?:

When I think about it, maybe just ignore this PR, and just drop requirement text. You can of course ignore my questions and/or PR. I'm not sure what happens if you try Pkg.add("PyCall") on 0.3, if you get an error or hint to install older; I guess not a worthy worry even if not as 0.3 is no longer supported.

stevengj:
0.3 is not supported, see 7d2fda566e2f265f1e1fdb749c59571d87d204cb

Issue #154, opened 2015-06-10
Using python context managers and missing values?


datnamer:
How can I use python context managers from within pycall?

Also how does Pycall handle missing values? For example, If I'm using a python function that manipulates a masked numpy array or pandas df with missing values.

stevengj:
Regarding context managers, according to [PEP 0343](https://www.python.org/dev/peps/pep-0343/), the `with` statement
```
        with EXPR as VAR:
            BLOCK
```
is equivalent to:
```
        mgr = (EXPR)
        exit = type(mgr).__exit__  # Not calling it yet
        value = type(mgr).__enter__(mgr)
        exc = True
        try:
            try:
                VAR = value  # Only if "as VAR" is present
                BLOCK
            except:
                # The exceptional case is handled here
                exc = False
                if not exit(mgr, *sys.exc_info()):
                    raise
                # The exception is swallowed if exit() returns true
        finally:
            # The normal and non-local-goto cases are handled here
            if exc:
                exit(mgr, None, None, None)
```
so you could always try calling the `__enter__` and `exit` functions manually.  I suppose PyCall could supply a `@with` macro to automate this.   Do you have a particular example of something you'd like to call this way?

stevengj:
PyCall does not have any specific awareness of missing data.  Of course, you can still use it, but by calling the Python functions on the pandas `PyObject` or whatever.

tkelman:
That already exists - https://github.com/malmaud/Pandas.jl

datnamer:
@stevengj  thanks.

Intended use case  for both is to access pymc 3 through pycall and calling it on a dataframe with missing values to do imputation and inference:

Basic example:
https://github.com/pymc-devs/pymc3/blob/master/docs/notebooks/getting_started.ipynb

```python
from pymc3 import NUTS, sample
with basic_model:
    
    # obtain starting values via MAP
    start = find_MAP(fmin=optimize.fmin_powell)
    
    # instantiate sampler
    step = NUTS(scaling=start) 
    
    # draw 500 posterior samples
    trace = sample(500, step, start=start)
```

@tkelman Cool thanks!

PR #375, opened 2017-03-23
add @with macro

marius311:
I needed to use a with statement with a ContextManager, so I went ahead an implemented the suggestion from https://github.com/JuliaPy/PyCall.jl/issues/154#issuecomment-117811984. Let me know if this is something you'd like to have, I'm also happy to make any further changes.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -8,7 +8,7 @@ export pycall, pyimport, pybuiltin, PyObject, PyReverseDims,
        pyisinstance, pywrap, pytypeof, pyeval, PyVector, pystring,
        pyraise, pytype_mapping, pygui, pygui_start, pygui_stop,
        pygui_stop_all, @pylab, set!, PyTextIO, @pysym, PyNULL, @pydef,
-       pyimport_conda, @py_str, @with
+       pyimport_conda, @py_str, @pywith, @pycall
 
 import Base: size, ndims, similar, copy, getindex, setindex!, stride,
        convert, pointer, summary, convert, show, haskey, keys, values,
@@ -438,33 +438,40 @@ end
 #########################################################################
 
 """
-    @with
+    @pywith
 
 Mimics a Python 'with' statement. Usage:
 
-@with EXPR as VAR begin
+@pywith EXPR[::TYPE_EXPR] [as VAR[::TYPE_VAR]] begin
     BLOCK
 end
 
-or 
+TYPE_EXPR and TYPE_VAR can optionally be used to override automatic conversion
+to Julia types in cases where this is not desired. 
 
-@with EXPR begin
-    BLOCK
-end
 """
-macro with(EXPR,as,VAR,BLOCK)
-    @assert as==:as "usage: @with EXPR as VAR BLOCK."
-    _with(EXPR,VAR,BLOCK)
+macro pywith(EXPR,as,VAR,BLOCK)
+    if isa(VAR,Expr) && (VAR.head==:(::))
+        VAR,TYPE = VAR.args
+    else
+        TYPE = PyAny
+    end
+    @assert (as==:as) && isa(VAR,Symbol) "usage: @pywith EXPR [as VAR[::TYPE]] BLOCK."
+    _pywith(EXPR,VAR,TYPE,BLOCK)
 end
-macro with(EXPR,BLOCK)
-    _with(EXPR,nothing,BLOCK)
+macro pywith(EXPR,BLOCK)
+    _pywith(EXPR,nothing,PyObject,BLOCK)
 end
         
-function _with(EXPR,VAR,BLOCK)        
+        
+function _pywith(EXPR,VAR,TYPE,BLOCK)
+    EXPR_str = string(EXPR)
     quote
-        mgr = $(esc(EXPR))
-        exit = pybuiltin("type")(mgr)[:__exit__]  # Not calling it yet
-        value = pybuiltin("type")(mgr)[:__enter__](mgr)
+        mgr = $(isa(EXPR,Expr) && EXPR.head==:(::) ? :(@pycall $(esc(EXPR))) : esc(EXPR))
+        @assert isa(mgr,PyObject) "@pywith: `$($EXPR_str)` did not return a PyObject. If this is a call to a Python function, try `$($EXPR_str)::PyObject` to turn off automatic conversion."
+        mgrT = pytypeof(mgr)
+        exit = mgrT["__exit__"]  # Not calling it yet
+        value = pycall(mgrT["__enter__"],$(esc(TYPE)),mgr)
         exc = true
         try
             try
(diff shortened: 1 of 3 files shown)


src/PyCall.jl
```diff
@@ -440,6 +440,59 @@ end
...
+
+@with EXPR begin
+    BLOCK
+end
+"""
+macro with(EXPR,as,VAR,BLOCK)
+    @assert as==:as "usage: @with EXPR as VAR BLOCK."
+    _with(EXPR,VAR,BLOCK)
+end
+macro with(EXPR,BLOCK)
```
stevengj: I'd prefer `@pywith` to make it clear that this is for Python objects.

src/PyCall.jl
```diff
@@ -440,6 +440,59 @@ end
...
+    _with(EXPR,nothing,BLOCK)
+end
+        
+function _with(EXPR,VAR,BLOCK)        
+    quote
+        mgr = $(esc(EXPR))
+        exit = pybuiltin("type")(mgr)[:__exit__]  # Not calling it yet
+        value = pybuiltin("type")(mgr)[:__enter__](mgr)
```
stevengj: We should probably define
```jl
pytypeof(o::PyObject) = PyObject(@pycheckn ccall(@pysym(:PyObject_Type), PyPtr, (PyPtr,), o))
```
for these kinds of queries.  And probably use `mgrT = pytypeof(mgr)` to avoid calling it twice.

Use `["__exit__"]` and `["__enter__"]` to circumvent the type-introspection process, since we know we want PyObjects here.

Probably there should be a way to specify the type of object you want for `value`, by e.g. `@pywith expr as var::T ...`.   `T` would default to `PyAny` if `VAR` is specified, or `PyObject` if `VAR==nothing` (since in that case we don't need to convert the `__enter__` result to a native Julia type).  The you would call `value = pycall(mgrT["__enter__"], T, mgr)` to get the desired value type.

src/PyCall.jl
```diff
@@ -440,6 +440,59 @@ end
...
+        exc = true
+        try
+            try
+                $(VAR==nothing ? :() : :($(esc(VAR)) = value))
+                $(esc(BLOCK))
+            catch err
+                # The exceptional case is handled here
+                exc = false
+                if !(pybuiltin("bool")(exit(mgr, pyimport(:sys)[:exc_info]()...)))
```
stevengj: What does the `bool` call do here?

stevengj:
Seems like a reasonable feature to have, since it is hard for users to do the enter/exit things themselves.

src/PyCall.jl
marius311: Good points, had not thought of that.

src/PyCall.jl
marius311: Just that Python interprets None as False, and all sorts of other cases, but Julia doesn't. This makes it work consistently in Julia.

marius311: Maybe there's already a builtin way in PyCall of doing this kind of thing I don't know about?

marius311:
Cool, thanks. I'll try to tackle some of these changes over the weekend.

marius311:
Ok, I added all the changes you mentioned. Also, I realized if the context manager returns a convertable type, this won't work either, so I made it so you can put `::PyObject` after the `EXPR` part too. This was in fact the thing I originally wanted to do, wihch now looks like, 

```julia
@pywith seaborn.plotting_context("paper")::PyObject begin
    plot([1,2,3])
end
```

the `::PyObject` part needed since `seaborn.plotting_context("paper")` returns a Dict (which is also a context manager, but would otherwise be converted to a Julia Dict before our macro gets to call `__enter__`, etc..)


Actually, as this kind  of thing seems to come up often and I thought the syntax was convenient, I split that off into its own macro `@pycall` which turns `@pycall func(args...)::T` into `pycall(func,T,args...)` which exists on its own (and which is used by this `@pywith` macro) . That's included in this PR, but I can take it out if there's problems and just leave the functionality inside `@pywith`.

src/PyCall.jl
stevengj: I was hoping you could do `@pycall exit(mgr, pyimport(:sys)[:exc_info]()...)::Bool` and it would work, but this expects an integer type.

It would be better if the `Bool` conversion were changed to use `PyObject_IsTrue`, and then you could use `::Bool`.

src/PyCall.jl
```diff
@@ -576,6 +636,25 @@ pycall(o::Union{PyObject,PyPtr}, ::Type{PyAny}, args...; kwargs...) =
...
+pytypeof(o::PyObject) = PyObject(@pycheckn ccall(@pysym(:PyObject_Type), PyPtr, (PyPtr,), o))
+
+
+"""
+    @pycall func(args...)::T
+
+Convenience macro which turns `func(args...)::T` into pycall(func, T, args...)
+"""
+macro pycall(ex)
+    @assert (isa(ex,Expr) && (ex.head==:(::))
```
stevengj: This should just be a standard exception (probably `ArgumentError`), not an assertion failure.  Assertion failures should only be used when failure would indicate a bug in PyCall itself.

src/PyCall.jl
```diff
@@ -576,6 +636,25 @@ pycall(o::Union{PyObject,PyPtr}, ::Type{PyAny}, args...; kwargs...) =
...
+    @pycall func(args...)::T
+
+Convenience macro which turns `func(args...)::T` into pycall(func, T, args...)
+"""
+macro pycall(ex)
+    @assert (isa(ex,Expr) && (ex.head==:(::)) 
+            && isa(ex.args[1],Expr) && (ex.args[1].head==:call)) "Usage: @pycall f(args...)::T"
+    func = ex.args[1].args[1]
+    args, kwargs = ex.args[1].args[2:end], []
+    if isa(args[1],Expr) && (args[1].head==:parameters)
```
stevengj: You can use `Meta.isexpr(args[1], :parameters)` for tests like this.

stevengj:
`@pycall` sounds reasonable to me.

Needs tests as well.

marius311:
Is it bad form to do disk IO from a test? `open` is about the only context manager in the Python standard library I can think of.

stevengj:
You can test with a `tempname()` file, and delete it after you are done.

stevengj:
Bump.  Would be great to merge this given tests.

src/PyCall.jl
stevengj: I've updated the latest PyCall master to use `PyObject_IsTrue`, so you can just do `!pycall(exit, Bool, mgr, pyimport(:sys)[:exc_info]()...)`

marius311:
Yes, sorry, been on my todo list to finish this, I'll try to take a look soon!

src/PyCall.jl
```diff
@@ -438,6 +438,73 @@ end
...
+                if !(@pycall exit(mgr, pyimport(:sys)[:exc_info]()...)::Bool)
+                    throw(err)
+                end
+            end
+        finally
+            if exc
+                exit(mgr, nothing, nothing, nothing)
+            end
+        end
+        nothing
```
marius311: The reasoning here btw was that Python with-statements don't return a value, so this is for consistency. You could certainly let them return something in Julia, but then you have to make a decision what to return in case there was an error that was swallowed up by the context manager's exit function. I figured this was just simpler.

test/runtests.jl
```diff
@@ -383,3 +383,33 @@ end
...
+fname = tempname()
+try
+    @test begin
+        @pywith pybuiltin("open")(fname,"w") as f begin
+            f[:write]("test")
+        end
+        open(readstring,fname)=="test"
+    end
+finally
+    rm(fname)
```
stevengj: `rm(fname, force=true)` in case there was an error creating the file.

marius311:
Just remembered this, I'm not sure if you're waiting on more from me here, but I think its ready (the test failure above was just random, I think)

stevengj:
Looks good, just needs a rebase to be merged.

marius311:
Sounds good. Although I don't know enough how a rebase plays with Github. Would I then open a new PR? Or do you mean a merge? (or something else?)

stevengj:
You do a forced push to the same branch, which will update this PR.

Ismael-VC:
@marius311 nowadays you should be able to fix conflicts using GitHub's web GUI:

* https://help.github.com/articles/resolving-a-merge-conflict-on-github

After that @stevengj should be able to also squash the commits prior to merging using the GitHub web GUI: 

* https://github.com/blog/2141-squash-your-commits

I hope this helps!

marius311:
Thanks, I didn't know about the Github conflict fixing feature, and also for the reminder that I forgot to ping @stevengj that this was OK to merge. Anyway, it looks like there's still no conflicts despite the time that's passed since I rebased this, so I think you go ahead whenever now (I think the one build error is unrelated).

stevengj:
Looks good; Travis error seems unrelated.

Issue #376, opened 2017-03-27
Python TypeError


fuzzybear3965:
I am trying to use PyCall with [fbprophet](https://facebookincubator.github.io/prophet/docs/quick_start.html) and I'm having an issue with sending a Julia DataFrames.DataFrame type to Python.  Below is the code that I'm trying 

```julia
import CSV
import PyCall

data = CSV.read("..\\some.csv")
PyCall.@pyimport fbprophet as fbp
m = fbp.Prophet()
m[:fit](data)
```
`typeof(data) == DataFrames.DataFrame` and the stack trace that I get is the following.
```
ERROR: LoadError: PyError (:PyObject_Call) <class 'TypeError'>
TypeError("'PyCall.jlwrap' object is not subscriptable",)
  File "C:\Users\JohnRinehart\AppData\Local\Programs\Python\Python36\lib\site-packages\fbprophet-0.1.post1-py3.6.egg\fbprophet\forecaster.py", line 473, in fit
    history = df[df['y'].notnull()].copy()

 in pyerr_check at C:\Users\JohnRinehart\.julia\v0.5\PyCall\src\exception.jl:56 [inlined]
 in pyerr_check at C:\Users\JohnRinehart\.julia\v0.5\PyCall\src\exception.jl:61 [inlined]
 in macro expansion at C:\Users\JohnRinehart\.julia\v0.5\PyCall\src\exception.jl:81 [inlined]
 in #_pycall#66(::Array{Any,1}, ::Function, ::PyCall.PyObject, ::DataFrames.DataFrame, ::Vararg{DataFrames.DataFrame,N}) at C:\Users\JohnRinehart\.julia\v0.5\PyCall\src\PyCall.jl:551
 in _pycall(::PyCall.PyObject, ::DataFrames.DataFrame, ::Vararg{DataFrames.DataFrame,N}) at C:\Users\JohnRinehart\.julia\v0.5\PyCall\src\PyCall.jl:539
 in #pycall#70(::Array{Any,1}, ::Function, ::PyCall.PyObject, ::Type{PyCall.PyAny}, ::DataFrames.DataFrame, ::Vararg{DataFrames.DataFrame,N}) at C:\Users\JohnRinehart\.julia\v0.5\PyCall\src\PyCall.jl:573
[...]

Issue #15, opened 2013-03-20
Segfault related to PyCall


kmsquire:
I've been getting random julia segfaults recently.  Nothing specifically reproducible, but in this case, it happened at the REPL on the command immediately after I interrupted a PyCall function:

``` julia
julia> df = E.extract_db_info()
^CERROR: interrupt
 in pycall at /scratch1/tmp/kmsquire/.julia/PyCall/src/PyCall.jl:62
 in anonymous at /scratch1/tmp/kmsquire/.julia/PyCall/src/conversions.jl:136
 in extract_db_info at /scratch1/tmp/kmsquire/Studies/TCGA_data/patients/ExtractDbInfo.jl:52

julia> df = E.extract_db_info();
Segmentation fault (core dumped)
```

The core dump gives 

``` gdb
Core was generated by `julia'.
Program terminated with signal 11, Segmentation fault.
#0  0x00002b7fe2244967 in PyObject_IsInstance (inst=0x73e96e0, cls=0x2b7fe259b580) at Objects/abstract.c:2959
2959                if (Py_EnterRecursiveCall(" in __instancecheck__")) {
```

Perhaps `pycall` needs to be atomic in this case, or something isn't getting cleaned up by my interruption?

PR #384, opened 2017-04-19
Make dictionary conversion more conservative

malmaud:
Stop mapping all Python objects that support the mapping protocol to
Julia dictionaries.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -341,7 +341,7 @@ end
 let weakdict = pyimport("weakref")["WeakValueDictionary"]
     # (use weakdict for the value, since Python supports
     #  weak references to type objects)
-    @test weakdict(Dict(3=>weakdict)) == Dict(3=>weakdict)
+    @test weakdict(Dict(3=>weakdict))[:__getitem__](3) == weakdict
 end
 
 # Expose python docs to Julia doc system
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -341,7 +341,8 @@ end
 let weakdict = pyimport("weakref")["WeakValueDictionary"]
     # (use weakdict for the value, since Python supports
     #  weak references to type objects)
-    @test weakdict(Dict(3=>weakdict))[:__getitem__](3) == weakdict
+    @test convert(Dict{Int,PyObject}, weakdict(Dict(3=>weakdict))) == Dict(3=>weakdict)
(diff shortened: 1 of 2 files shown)


fuzzybear3965:
This fixes [#15 from Pandas.jl](https://github.com/JuliaPy/Pandas.jl/issues/15) and #376.

malmaud:
I fixed the issue stopping PyPlot from finding a backend

fuzzybear3965:
Does this close [PyPlot issue #289](https://github.com/JuliaPy/PyPlot.jl/issues/289)?

malmaud:
Ya, might as well close that.

test/runtests.jl
```diff
@@ -341,7 +341,7 @@ end
 let weakdict = pyimport("weakref")["WeakValueDictionary"]
     # (use weakdict for the value, since Python supports
     #  weak references to type objects)
-    @test weakdict(Dict(3=>weakdict)) == Dict(3=>weakdict)
+    @test weakdict(Dict(3=>weakdict))[:__getitem__](3) == weakdict
```
stevengj: Probably:
```jl
@test Dict(weakdict(Dict(3=>weakdict))) == Dict(3=>weakdict)
@test weakdict(Dict(3=>weakdict))[3] == weakdict
```
The first test would have caught the PyPlot problem, and for the second test I think you can just do `[3]` rather than calling `__getitem__`?

malmaud: You can if you wrap the result of `weakdict` in a PyDict. With this PR, `weakdict` is just going to return a raw `PyObject`.

stevengj: @malmaud, `getindex(o::PyObject, i::Integer)` is defined and calls Python's getitem method.

stevengj: Hmm, actually it calls `PySequence_GetItem` which is not quite the same thing.  It should actually be changed to call `PyObject_GetItem`, which does the right thing (calling `__getitem__` for a mapping and `PySequence_GetItem` for a sequence).

malmaud: OK, I revised the test. Looks like that commit causes the tests to fail, though.

stevengj: Whoops, it breaks Python 3.  I'll probably have to revert it.  However, you can still do `get(pyobject, 3)`.

stevengj:
Rebased and updated the test.

malmaud:
Great

Issue #325, opened 2016-10-13
crashes after workspace() in 0.5


stevengj:
`using PyCall; workspace(); using PyCall` crashes.

(I have to say that I hate the whole `workspace` hack; just an endless source of fragility.)

PR #388, opened 2017-05-04
Custom serialization for PyObject

iamed2:
This is my first crack at doing this. This allows passing PyObjects around in RemoteChannels. 

One issue with this:

This works:
```julia
using PyCall
@everywhere using PyCall
```

But this segfaults:
```julia
@everywhere using PyCall
```

I know the latter does the NxN loading thing and isn't advized, but it's something people will do. Advice is welcome!

### DIFF
--- a/src/serialize.jl
+++ b/src/serialize.jl
@@ -1,0 +1,18 @@
+const pickle = PyNULL()
+
+function get_pickle()
+    if pickle.o == C_NULL
+        copy!(pickle, pyimport("pickle"))
+    end
+
+    return pickle
+end
+
+function Base.serialize(s::AbstractSerializer, pyo::PyObject)
+    Base.serialize_type(s, PyObject)
+    serialize(s, get_pickle()[:dumps](pyo))
+end
+
+function Base.deserialize(s::AbstractSerializer, t::Type{PyObject})
+    get_pickle()[:loads](deserialize(s))
+end
--- a/src/serialize.jl
+++ b/src/serialize.jl
@@ -1,18 +1,29 @@
-const pickle = PyNULL()
+const _pickle = PyNULL()
 
-function get_pickle()
-    if pickle.o == C_NULL
-        copy!(pickle, pyimport("pickle"))
-    end
-
-    return pickle
-end
+pickle() = _pickle.o == C_NULL ? copy!(_pickle, pyimport("pickle")) : _pickle
 
 function Base.serialize(s::AbstractSerializer, pyo::PyObject)
     Base.serialize_type(s, PyObject)
-    serialize(s, get_pickle()[:dumps](pyo))
+
+    b = PyBuffer(pycall(pickle()[:dumps], PyObject, pyo))
+
+    serialize(s, unsafe_wrap(Array, Ptr{UInt8}(pointer(b)), sizeof(b)))
 end
 
 function Base.deserialize(s::AbstractSerializer, t::Type{PyObject})
-    get_pickle()[:loads](deserialize(s))
+    b = deserialize(s)
+
+    pycall(
+        pickle()[:loads],
+        PyObject,
+        PyObject(
+            @pycheckn ccall(
+                @pysym(PyString_FromStringAndSize),
+                PyPtr,
+                (Ptr{UInt8}, Int),
+                b,
+                sizeof(b)
+            )
+        )
+    )
 end
--- a/src/serialize.jl
+++ b/src/serialize.jl
@@ -3,6 +3,7 @@ const _pickle = PyNULL()
 pickle() = _pickle.o == C_NULL ? copy!(_pickle, pyimport(PyCall.pyversion.major ≥ 3 ? "pickle" : "cPickle")) : _pickle
 
 function Base.serialize(s::AbstractSerializer, pyo::PyObject)
+    pyo.o == C_NULL && return invoke(serialize, (AbstractSerializer, Any), s, pyo)
     Base.serialize_type(s, PyObject)
     b = PyBuffer(pycall(pickle()["dumps"], PyObject, pyo))
     serialize(s, unsafe_wrap(Array, Ptr{UInt8}(pointer(b)), sizeof(b)))
@@ -10,7 +11,7 @@ end
 
 function Base.deserialize(s::AbstractSerializer, t::Type{PyObject})
     b = deserialize(s)
-    pycall(pickle()["loads"], PyObject,
+    return isa(b, PyPtr) ? PyObject(b) : pycall(pickle()["loads"], PyObject,
         PyObject(@pycheckn ccall(@pysym(PyString_FromStringAndSize),
                 PyPtr, (Ptr{UInt8}, Int),
                 b, sizeof(b))))
--- a/src/serialize.jl
+++ b/src/serialize.jl
@@ -3,16 +3,24 @@ const _pickle = PyNULL()
 pickle() = _pickle.o == C_NULL ? copy!(_pickle, pyimport(PyCall.pyversion.major ≥ 3 ? "pickle" : "cPickle")) : _pickle
 
 function Base.serialize(s::AbstractSerializer, pyo::PyObject)
-    pyo.o == C_NULL && return invoke(serialize, (AbstractSerializer, Any), s, pyo)
     Base.serialize_type(s, PyObject)
-    b = PyBuffer(pycall(pickle()["dumps"], PyObject, pyo))
-    serialize(s, unsafe_wrap(Array, Ptr{UInt8}(pointer(b)), sizeof(b)))
+    if pyo.o == C_NULL
+        serialize(s, pyo.o)
+    else
+        b = PyBuffer(pycall(pickle()["dumps"], PyObject, pyo))
+        serialize(s, unsafe_wrap(Array, Ptr{UInt8}(pointer(b)), sizeof(b)))
+    end
 end
 
 function Base.deserialize(s::AbstractSerializer, t::Type{PyObject})
     b = deserialize(s)
-    return isa(b, PyPtr) ? PyObject(b) : pycall(pickle()["loads"], PyObject,
-        PyObject(@pycheckn ccall(@pysym(PyString_FromStringAndSize),
-                PyPtr, (Ptr{UInt8}, Int),
-                b, sizeof(b))))
+    if isa(b, PyPtr)
+        @assert b == C_NULL
+        return PyNULL()
+    else
+        return pycall(pickle()["loads"], PyObject,
+            PyObject(@pycheckn ccall(@pysym(PyString_FromStringAndSize),
+                    PyPtr, (Ptr{UInt8}, Int),
+                    b, sizeof(b))))
+    end
 end
--- a/src/serialize.jl
+++ b/src/serialize.jl
@@ -1,29 +1,28 @@
 const _pickle = PyNULL()
 
-pickle() = _pickle.o == C_NULL ? copy!(_pickle, pyimport("pickle")) : _pickle
+function pickle()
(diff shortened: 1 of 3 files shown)


src/serialize.jl
```diff
@@ -0,0 +1,18 @@
+const pickle = PyNULL()
```
stevengj: I would just call this `_pickle` and the function `pickle()`

src/serialize.jl
```diff
@@ -0,0 +1,18 @@
...
+    if pickle.o == C_NULL
+        copy!(pickle, pyimport("pickle"))
+    end
+
+    return pickle
+end
+
+function Base.serialize(s::AbstractSerializer, pyo::PyObject)
+    Base.serialize_type(s, PyObject)
+    serialize(s, get_pickle()[:dumps](pyo))
```
stevengj: In low-level functions like this you want to avoid the automatic type detection.  You could do `pycall(get_pickle()["dumps"], String, pyo)`.

However, it would be even better to use the low-level buffer interface here.  That way you can avoid copying the data entirely, and work with the raw byte array.

stevengj: In particular:
```
b = PyBuffer(pycall(p["dumps"], PyObject, [1,2,3]))
serialize(s, unsafe_wrap(Array, Ptr{UInt8}(b.buf.buf), b.buf.len))
```
should be about as good as we can get without dropping down deeper into the `serialize` internals to work with raw pointers.

src/serialize.jl
```diff
@@ -0,0 +1,18 @@
...
+end
+
+function Base.deserialize(s::AbstractSerializer, t::Type{PyObject})
+    get_pickle()[:loads](deserialize(s))
```
stevengj: After changing to use `PyBuffer` above, this should become:
```jl
b = deserialize(s) # Vector{UInt8}
pycall(get_pickle()["loads"], PyObject,
          PyObject(@pycheckn ccall(@pysym(PyString_FromStringAndSize),
                                        PyPtr, (Ptr{UInt8}, Int), b, sizeof(b))))
```
Note that you want to return a `PyObject` here, avoiding any conversion of the result to a Julia object.

test/runtests.jl
```diff
@@ -383,3 +383,8 @@ end
 @test convert(Bool, PyObject(17.3)) === true
 @test convert(Bool, PyObject(Any[0])) === true
 @test Bool(PyVector{PyObject}(PyObject([false]))[1]) === false
+
+# serialization
+let py_sum_obj = pybuiltin("sum")
+    @test py_sum_obj == deserialize(IOBuffer(sprint(io->serialize(io, py_sum_obj))))
```
stevengj: Maybe better:
```jl
b = IOBuffer()
serialize(b, py_sum_obj)
@test py_sum_obj == deserialize(seekstart(b))
```
(Seems weird to force this to go through a string.

iamed2:
Thanks for the review! I'll get on this.

stevengj:
The failures on Python 3 should hopefully be fixed by my suggestions.

src/serialize.jl
stevengj: (This unfortunately makes a copy of the buffer, but I haven't implemented a copy-free Julia pybuffer wrapper around a Julia array yet.)

src/serialize.jl
```diff
@@ -0,0 +1,18 @@
+const pickle = PyNULL()
+
+function get_pickle()
+    if pickle.o == C_NULL
+        copy!(pickle, pyimport("pickle"))
+    end
+
+    return pickle
+end
```
stevengj: Probably just a one-liner:
```
pickle() = pickle.o == C_NULL ? copy!(_pickle, pyimport("pickle")) : _pickle
```

src/serialize.jl
```diff
@@ -0,0 +1,29 @@
+const _pickle = PyNULL()
+
+pickle() = _pickle.o == C_NULL ? copy!(_pickle, pyimport("pickle")) : _pickle
+
+function Base.serialize(s::AbstractSerializer, pyo::PyObject)
+    Base.serialize_type(s, PyObject)
+
```
stevengj: no blank lines in a short function like this.

src/serialize.jl
```diff
@@ -0,0 +1,29 @@
...
+    b = PyBuffer(pycall(pickle()[:dumps], PyObject, pyo))
```
stevengj: `pickle()["dumps"]` to avoid the type-introspection step

src/serialize.jl
```diff
@@ -0,0 +1,29 @@
...
+    pycall(
+        pickle()[:loads],
+        PyObject,
+        PyObject(
+            @pycheckn ccall(
+                @pysym(PyString_FromStringAndSize),
+                PyPtr,
+                (Ptr{UInt8}, Int),
+                b,
+                sizeof(b)
```
stevengj: this whole function call should fit on 2 lines

src/serialize.jl
iamed2: Oh! I had them backwards, that was my confusion.

iamed2:
We disagree on style but it's your repo :) I've made the changes. 

Tests appear to pass, so the last issue left is whether we care about the issue mentioned in my original PR comment.

src/serialize.jl
```diff
@@ -0,0 +1,28 @@
+const _pickle = PyNULL()
+
+function pickle()
+    if _pickle.o == C_NULL
+        if pyexists(:cPickle)
+            copy!(_pickle, pyimport(:cPickle))
+        else
+            copy!(_pickle, pyimport(:pickle))
+        end
```
stevengj: Isn't cPickle always available in Python 2, and never available in Python 3?  So you could just do:
```jl
pickle() = _pickle.o == C_NULL ? copy!(_pickle, PyCall.pyversion.major ≥ 3 ? "pickle" : "cPickle") : _pickle
```

src/serialize.jl
```diff
@@ -0,0 +1,28 @@
...
+
+function Base.serialize(s::AbstractSerializer, pyo::PyObject)
+    Base.serialize_type(s, PyObject)
+    b = PyBuffer(pycall(pickle()["dumps"], PyObject, pyo))
+    serialize(s, unsafe_wrap(Array, Ptr{UInt8}(pointer(b)), sizeof(b)))
+end
+
+function Base.deserialize(s::AbstractSerializer, t::Type{PyObject})
+    b = deserialize(s)
+
```
stevengj: no blank line here

stevengj:
I don't really see a way to fix `@everywhere using`.   I really think this should be fixed in Base Julia: `@everywhere using Foo` should turn into `import Foo; @everywhere using Foo`

iamed2:
I agree

stevengj:
I'm actually a bit unclear on why `@everywhere using PyCall` fails, or what it has to do with serialization.   Aren't all the workers separate processes, and hence PyCall will spin up separate `libpython` instances?

iamed2:
I'm not sure, but it could be because compilation uses serialization at some point for something. The issue does not occur if I start Julia with `julia --compilecache=no -p auto`.

iamed2:
Here's the traceback for the segfault with only one worker (so it's actually one trace in order instead of jumbled): https://gist.github.com/iamed2/6164289f973c713f7995d356054a890c

stevengj:
Oh I see now.    The problem comes in two pieces:

* When we compile the cache, it serializes some global PyObjects.   We don't actually want to save these, but previously it just saved a garbage pointer that got overwritten by `__init__`.  With this PR, it now saves the pickled Python data.

* If you load the cache file, it deserializes these object.  Initially, however (before PyCall is loaded), it doesn't know how to deserialize PyObject and so it skips these.  Fine.

* If you do `@everywhere using PyCall`, however, it deserializes the module a *second* time.  At which point the PyObject deserializer method is registered and it tries to depickle, and things break.

This suggests a couple of changes:

* First, the `serialize` / `deserialize` methods need to be updated to handle a NULL PyObject without crashing.

* Second, if possible it would be better if the caching process could serialize PyObjects as if they were NULL (even if they aren't), i.e. avoid pickling.

iamed2:
I was considering experimenting with the former, so I'll do that now.

The latter I'm not sure about. I'll experiment once I've done the NULL-handling.

stevengj:
Actually, now that I look at Julia's `src/dump.c`, I'm not sure that it uses the Julia `serialize` method at all.  It seems like it uses a more low-level `serialize` routine restricted to a small set of types, in which case we may be okay.

(`@everywhere using PyCall` will still break if you don't `import PyCall` first because PyCall does not like to be initialized twice.  This is essentially the same problem as #325.)

iamed2:
>  At which point the PyObject deserializer method is registered and it tries to depickle, and things break

Shouldn't depickling work at that point? 

> Second, if possible it would be better if the caching process could serialize PyObjects as if they were NULL (even if they aren't), i.e. avoid pickling.

I don't know how we would hook into the caching process. My only idea at this point is to put all of PyCall's global PyObject references in a special Dict type that serializes PyObjects to NULL. Thoughts?

stevengj:
It looks like caching might use a custom serialization limited to a few types, so I don't think it is affected by this after all?

To test this, you can try declaring a global `const x = PyObject("foobarbaz")` and see if the `.ji` file contains the string `S'foobarbaz'` (which appears in the pickled representation of that Python string).

src/serialize.jl
```diff
@@ -0,0 +1,18 @@
+const _pickle = PyNULL()
+
+pickle() = _pickle.o == C_NULL ? copy!(_pickle, pyimport(PyCall.pyversion.major ≥ 3 ? "pickle" : "cPickle")) : _pickle
+
+function Base.serialize(s::AbstractSerializer, pyo::PyObject)
+    pyo.o == C_NULL && return invoke(serialize, (AbstractSerializer, Any), s, pyo)
+    Base.serialize_type(s, PyObject)
+    b = PyBuffer(pycall(pickle()["dumps"], PyObject, pyo))
```
stevengj: Wouldn't it be sufficient to do:
```
Base.serialize_type(s, PyObject)
if pyo.o == C_NULL
    serialize(s, pyo.o)
else
   ...
end
```
rather than calling `invoke`?

iamed2:
> To test this, you can try declaring a global const x = PyObject("foobarbaz") and see if the .ji file contains the string S'foobarbaz' (which appears in the pickled representation of that Python string).

It is indeed not in there, so it looks like we don't have to worry about that, thankfully.

src/serialize.jl
```diff
@@ -0,0 +1,18 @@
...
+    serialize(s, unsafe_wrap(Array, Ptr{UInt8}(pointer(b)), sizeof(b)))
+end
+
+function Base.deserialize(s::AbstractSerializer, t::Type{PyObject})
+    b = deserialize(s)
+    return isa(b, PyPtr) ? PyObject(b) : pycall(pickle()["loads"], PyObject,
```
stevengj: Probably better to assert that the pointer is null:
```jl
if isa(b, PyPtr)
    @assert b == C_NULL)
    return PyNULL()
else
    return ...
end
```

iamed2: I guess if a pointer happens to not be null when being deserialized someone will file an issue and we'll have to reevaluate anyway.

src/serialize.jl
iamed2: You're right; if I handle both input and output PyNULL PyPtr serialization under PyObject it'll be fine.

iamed2:
The one test failure got a SIGINT and Travis was having infrastructure issues so I think that failure is unrelated to this PR.

stevengj:
Thanks!

iamed2:
Thanks for your help!

PR #416, opened 2017-07-22
RFC: Use invokelatest in jl_Function_call

### LINKED ISSUE #343: failure on 0.6 in anonymous callback

ihnorton:
Fixes https://github.com/JuliaPy/pyjulia/issues/105

I thought it would also fix #343, but the example in that issue description seems to work fine for me on 0.6 release without this PR (but the issue is still open).

Directly calls `Core._apply_latest` with a closure over kw args as in https://github.com/JuliaLang/julia/pull/22646 for 0.6.

cc @TsurHerman

### DIFF
--- a/src/callback.jl
+++ b/src/callback.jl
@@ -31,13 +31,31 @@ function jl_Function_call(self_::PyPtr, args_::PyPtr, kw_::PyPtr)
     args = PyObject(args_) # don't need pyincref because of finally clause below
     try
         f = unsafe_pyjlwrap_to_objref(self_)::Function
-        if kw_ == C_NULL
-            ret = PyObject(f(convert(PyAny, args)...))
-        else
-            kw = PyDict{Symbol,PyAny}(pyincref(kw_))
-            kwargs = [ (k,v) for (k,v) in kw ]
-            ret = PyObject(f(convert(PyAny, args)...; kwargs...))
+
+        # on 0.6 we need to use invokelatest to get execution in newest world
+        @static if isdefined(Base, :invokelatest)
+            if kw_ == C_NULL
+                ret = PyObject(Base.invokelatest(f, convert(PyAny, args)...))
+            else
+                kw = PyDict{Symbol,PyAny}(pyincref(kw_))
+                kwargs = [ (k,v) for (k,v) in kw ]
+
+                # 0.6 `invokelatest` doesn't support kwargs, instead
+                # use a closure over kwargs. see:
+                #   https://github.com/JuliaLang/julia/pull/22646
+                f_kw_closure() = f(convert(PyAny, args)...; kwargs...)
+                ret = PyObject(Core._apply_latest(f_kw_closure))
+            end
+        else # 0.5 support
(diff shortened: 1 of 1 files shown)


Keno:
So @vtjnash should correct me here, but I think this is the wrong place to do it. Instead, pyjulia should update the worldcounter in the ptls whenever it has imported some code. I'm just not sure if there's currently a C Api for that.

src/callback.jl
```diff
@@ -8,7 +8,7 @@
 ################################################################
 
 # Define a Python method/function object from f(PyPtr,PyPtr)::PyPtr.
-# Requires f to be a top-level function.
```
Keno: typo

Keno:
On the other hand, I also don't really see a downside to doing this here.

src/callback.jl
```diff
@@ -32,11 +32,16 @@ function jl_Function_call(self_::PyPtr, args_::PyPtr, kw_::PyPtr)
     try
         f = unsafe_pyjlwrap_to_objref(self_)::Function
         if kw_ == C_NULL
-            ret = PyObject(f(convert(PyAny, args)...))
```
Keno: This package still works on 0.5, so you need a check here e.g. `isdefined(Base, :invokelatest)`.

vtjnash:
`jl_Function_call` is already wrapped in a trampoline (via cfunction) which does the equivalent of `invokelatest`, so this shouldn't be necessarily (unless https://github.com/JuliaLang/julia/pull/20167 isn't doing it's job correctly)

ihnorton:
Updated.

@vtjnash I don't know whether that PR is doing its job, but there's a simple repro here: https://github.com/JuliaPy/pyjulia/issues/105#issuecomment-316405459

Keno:
Ok, so the underlying problem here will be addressed by https://github.com/JuliaLang/julia/pull/22987. However, we should probably still merge this here to fix this on current versions of 0.5. Once base is merged, we can put in an appropriate version check.

stevengj:
Thanks!

PR #419, opened 2017-07-25
Fix in-process symbol resolution

Keno:
We've since re-arranged the way we load dynamic libraries a bit, making
the current detection code ineffective. This should address most instances
of pyjulia accidentally loading the wrong python version.


stevengj:
Looks good to me.

stevengj:
Any idea on Travis failure?

Keno:
One of them is a network timeout, the other is an error that I believe we've seen before on other packages and should be unrelated.

PR #420, opened 2017-07-28
Try to fix in-process symbol resolution on OS X

Keno:
It turns out that, for some python versions on OS X, `main` is not a public symbol. However, all OS X executables have the `_mh_execute_header` Symbol, so try to use that instead on OS X.


stevengj:
It's amazing that by far the hardest part of PyCall has been reliably loading libpython.

Keno:
Travis failure is the same as in the PR, not related to this.

PR #422, opened 2017-07-28
Allow calling things that are not Functions

Keno:
All objects are callable nowadays, restricting to `Function` is odd and
precludes things like constructing types from Python in pyjulia.

### DIFF
--- a/src/callback.jl
+++ b/src/callback.jl
@@ -21,13 +21,13 @@ end
 
 ################################################################
 
-# To pass an arbitrary Julia function to Python, we wrap it
-# in a jl_Function Python class, where jl_Function.__call__
-# executes the jl_Function_callback function in Julia, which
-# in turn fetches the actual Julia function from the "f" attribute
+# To pass an arbitrary Julia object to Python, we wrap it
+# in a jlwrap Python class, where jlwrap.__call__
+# executes the pyjlwrap_call function in Julia, which
+# in turn fetches the actual Julia value from the python wrapper
 # and calls it.
 
-function jl_Function_call(self_::PyPtr, args_::PyPtr, kw_::PyPtr)
+function pyjlwrap_call(self_::PyPtr, args_::PyPtr, kw_::PyPtr)
     args = PyObject(args_) # don't need pyincref because of finally clause below
     try
         f = unsafe_pyjlwrap_to_objref(self_)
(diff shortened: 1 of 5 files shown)


stevengj:
Sounds good, but in that case we should probably just get rid of `jl_FunctionType` in `callback.jl` and merge its features with the generic `jlWrapType` (since generic Julia objects can have docstrings and attributes too).

Keno:
Updated to remove the `PyCall.jl_Function` wrapper.

src/pytype.jl
```diff
@@ -363,13 +363,46 @@ function pyjlwrap_hash32(o::PyPtr)
...
+    try
+        f = unsafe_pyjlwrap_to_objref(self_)
+        attr = convert(String, attr_)
+        if attr in ("__name__","func_name")
+            return pystealref!(PyObject(string(f)))
+        elseif attr in ("__doc__", "func_doc")
+            return pystealref!(PyObject(docstring(f)))
+        elseif attr in ("__module__","__defaults__","func_defaults","__closure__","func_closure")
+            return pystealref!(PyObject(nothing))
+        else
```
stevengj: The fallback should call getfield?

Keno: Done. I retained this code path for things that start with `__`, because it seems like python needs that to work correctly. If you'd prefer, we can probably find the complete list of such things, but that seemed easier.

src/callback.jl
```diff
@@ -65,42 +65,4 @@ function jl_Function_call(self_::PyPtr, args_::PyPtr, kw_::PyPtr)
...
-    end
-    return PyPtr_NULL
-end
-
-function pycallback(f::Function)
-    pyjlwrap_new(jl_FunctionType, f)
-end
-
-PyObject(f::Function) = pycallback(f)
+pycallback(f::Function) = PyObject(f)
```
stevengj: I think we can just delete the `pycallback` function entirely.

Keno: Ok, I wasn't sure if people were relying on it, but I guess it's not exported.

src/pytype.jl
```diff
@@ -363,13 +363,46 @@ function pyjlwrap_hash32(o::PyPtr)
     return h == reinterpret(UInt32, Int32(-1)) ? pysalt32 : h::UInt32
 end
 
+# just string(Docs.doc(f)) doesn't work on VERSION < v"0.5"
+function docstring(f::Function)
```
stevengj: Should not be restricted to `::Function`.

Now that we require 0.5, this can presumably become
```jl
docstring(x) = string(Docs.doc(x))
```

PR #425, opened 2017-08-12
Fix most of 0.7 depwarns

yuyichao:


yuyichao:
Updated to include more fixes. One depwarn waiting for https://github.com/JuliaLang/Compat.jl/pull/400 and not included.

yuyichao:
Note that getting loading to pass on nightly requires https://github.com/MikeInnes/MacroTools.jl/pull/57

yuyichao:
Ping.

yuyichao:
Ping.

yuyichao:
Ping. Added some more fix now that the Compat change is merged. CI pass will still need the MacroTools PR.

yuyichao:
Ping, anyone? @stevengj ?

This PR has been sitting here for a month now. And have been updated 3 times to include new fixes...............................

yuyichao:
And ref https://github.com/JuliaPy/Conda.jl/pull/82#issuecomment-318900851 https://github.com/JuliaPy/Conda.jl/pull/67#issuecomment-280509930 about organization permission setup.....

src/pyinit.jl
```diff
@@ -87,15 +99,7 @@ function __init__()
...
-        if pyversion.major < 3
-            argv_s = ""
-            argv = unsafe_convert(Ptr{UInt8}, argv_s)
-            ccall(@pysym(:PySys_SetArgvEx), Void, (Cint,Ptr{Ptr{UInt8}},Cint), 1, &argv, 0)
-        else
-            argv_s = Cwchar_t[0]
-            argv   = unsafe_convert(Ptr{Cwchar_t}, argv_s)
-            ccall(@pysym(:PySys_SetArgvEx), Void, (Cint, Ptr{Ptr{Cwchar_t}}, Cint), 1, &argv, 0)
-        end
+        ccall(@pysym(:PySys_SetArgvEx), Void, (Cint, Ptr{UInt32}, Cint), 1, EmptyStringList(), 0)
```
stevengj: Isn't there a simpler way to do this?

yuyichao: Keeping the branch and using `Ptr{Cstring}`, `Ptr{Cwstring}` and pass in an actual array of string should work too. I did it this way since,

1. This eliminate the branch.
2. So the code is actually not much longer
3. I want to see how easy it is to implement sth like this properly
4. It doesn't allocate anything (optimized out on 0.7) (yeah, I know it's just `__init__`)

When 0.6 support is dropped, this can be done with `@gc_preserve` without external definition.

```julia
ref0 = Ref{UInt32}(0)
@gc_preserve ref0 ccall(@pysym(:PySys_SetArgvEx), Void, (Cint, Ref{Ptr{Void}}, Cint), 1, pointer_from_objref(ref0), 0)
```

stevengj: The branch and the performance are irrelevant here.   But as your code shows, we don't actually need the branch, since a pointer to a pointer to an array of a 32-bit zero will work in both Python 2 and 3 (even though Python 2 needs only one byte of zero).

I would prefer to just use `Ptr{Cwstring}` to pass an an actual array rather than declaring a custom type here.

StefanKarpinski:
@yuyichao, I changed the default member privilege to write so you should be able to merge this yourself now – let me know if you can't.

yuyichao:
I can merge this now. I do want to wait for a new nightly though since the code here exposes a base bug....

PR #426, opened 2017-08-14
Don't attempt to get a non-existent field

Keno:
I thought this was ok, because the error would get propagated, but pyjulia
doesn't have a toplevel error handler for various things, causing problems here.
Instead, try to fall back to the generic python method for missing fields.

Fixes https://github.com/JuliaPy/pyjulia/issues/116

### DIFF
--- a/src/pytype.jl
+++ b/src/pytype.jl
@@ -365,6 +365,14 @@ end
 
 docstring(x) = string(Docs.doc(x))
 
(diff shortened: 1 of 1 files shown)


Keno:
Never mind, I think I found it.

stevengj:
LGTM.  Add a `@test_throws` test?

PR #428, opened 2017-08-15
Tighten return type for conversion to `Array`

Keno:
Makes the following work in pyjulia:
```python
import julia
j = julia.Julia()
j.eval("""
module FooBar
export Foo, bar
type Foo
  i::Int
end
function bar(a::Array{Foo,1})
  return sum([f.i for f in a])
end
end # module
""")
j.using("FooBar")
foo1 = j.Foo(1)
foo2 = j.Foo(2)
j.bar([foo1, foo2])
```

before this would fail because the julia-side type was `Vector{Any}`.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -392,7 +392,7 @@ function convert{T}(::Type{Vector{T}}, o::PyObject)
     py2array(T, Array{pyany_toany(T)}(len), o, 1, 1)
 end
 
-convert(::Type{Array}, o::PyObject) = py2array(PyAny, o)
+convert(::Type{Array}, o::PyObject) = map(identity, py2array(PyAny, o))
 convert{T}(::Type{Array{T}}, o::PyObject) = py2array(T, o)
 
 PyObject(a::BitArray) = PyObject(Array(a))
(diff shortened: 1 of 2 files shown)


src/conversions.jl
```diff
@@ -392,7 +392,7 @@ function convert{T}(::Type{Vector{T}}, o::PyObject)
     py2array(T, Array{pyany_toany(T)}(len), o, 1, 1)
 end
 
-convert(::Type{Array}, o::PyObject) = py2array(PyAny, o)
+convert(::Type{Array}, o::PyObject) = map(identity, py2array(PyAny, o))
```
stevengj: Doesn't this make an extra copy of the array, though?

Keno: It does, but I think it's an acceptable tradeoff. I had a different version that tried really hard to avoid that, but it ended up just being more complicated and hard to understand. At some future point, the compiler might even be able to optimize this away.

stevengj:
A lot of tests are failing...

Keno:
Ah, looks like the return-type tightening code in Base is new in 0.6. Nightly is failing because of the macro changes I believe (and unreleated to this PR). 0.6 is passing. I'd be fine with just disabling the new test on 0.5, since you're gonna have array type problems on 0.5 anyway because of the way that `map` worked back then.

stevengj:
Sure.

Keno:
Fixed on 0.5. Nightly is still broken, but unrelated.

PR #432, opened 2017-08-18
support iterating over Julia objects from Python

### LINKED ISSUE #431: Support Julia iterators in Python code
It would be great if PyCall could support iterating through Julia generators in Python code:

```julia
julia> gen = (x for x in 1:3)
Base.Generator{UnitRange{Int64},##1#2}(#1, 1:3)

julia> py"""
       for el in $gen:
           print el
       """
ERROR: PyError (ccall(@pysym(:PyEval_EvalCode), PyPtr, (PyPtr, PyPtr, PyPtr), o, globals, locals)) <type 'exceptions.TypeError'>
TypeError("'PyCall.jlwrap' object is not iterable",)
  File "/Users/omus/.julia/v0.6/PyCall/src/pyeval.jl", line 1, in <module>
    const Py_single_input = 256  # from Python.h

Stacktrace:
 [1] pyerr_check at /Users/omus/.julia/v0.6/PyCall/src/exception.jl:56 [inlined]
 [2] pyerr_check at /Users/omus/.julia/v0.6/PyCall/src/exception.jl:61 [inlined]
 [3] macro expansion at /Users/omus/.julia/v0.6/PyCall/src/exception.jl:81 [inlined]
 [4] pyeval_(::String, ::PyCall.PyDict{String,PyCall.PyObject,false}, ::PyCall.PyDict{String,PyCall.PyObject,false}, ::Int64, ::String) at /Users/omus/.julia/v0.6/PyCall/src/pyeval.jl:23
```

stevengj:
Closes #431.

For example:
```jl
julia> gen = (x for x in 1:3)
Base.Generator{UnitRange{Int64},##1#2}(#1, 1:3)

julia> py"""
              for el in $gen:
                  print el
         """
1
2
3

julia> py"[x**2 for x in $(x for x in 1:4)]"
4-element Array{Int64,1}:
  1
  4
  9
 16
```

cc @omus


src/pytype.jl
```diff
@@ -292,7 +292,7 @@ end
 
 # Often, PyTypeObject instances are global constants, which we initialize
 # to 0 via PyTypeObject() and then initialize at runtime via PyTypeObject!
-function PyTypeObject!(t::PyTypeObject, name::AbstractString, basicsize::Integer, init::Function)
+function PyTypeObject!(init::Function, t::PyTypeObject, name::AbstractString, basicsize::Integer)
```
stevengj: Note that I moved the `init` function to go first here because that allows us to use `do` syntax.

This is an undocumented internal function so I'm guessing we should be okay without adding a deprecation?    I can't find any packages using this low-level code to define their own `pyjlwrap` subtypes.

omus:
Works great!

stevengj:
(Hmm, `1:4` is not a good test for this, actually, since that already gets converted to an `xrange`.  But `Generator` objects are good.)

PR #452, opened 2017-11-27
Typo introduced [ci skip]

PallHaraldsson:


PallHaraldsson:
Doc was good as is, not sure if meant to add anything meaningful..

PR #457, opened 2018-01-11
fixed missing conversion of Int128

ranocha:
Before this PR, `Int128`s could not be converted to `PyObject`s:
```julia
using PyCall

julia> i=12345678901234567890; @show typeof(i); PyObject(i)
typeof(i) = Int128
ERROR: InexactError()
Stacktrace:
 [1] macro expansion at ...(.julia/v0.6/PyCall/src/exception.jl:78 [inlined]
 [2] PyCall.PyObject(::Int128) at .../.julia/v0.6/PyCall/src/conversions.jl:10
```

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -653,24 +653,17 @@ end
 pymp_query(o::PyObject) = pyisinstance(o, mpf) ? BigFloat : pyisinstance(o, mpc) ? Complex{BigFloat} : Union{}
 
 #########################################################################
-# Int128 and BigInt conversion to Python "long" integers
+# (Int64), Int128 and BigInt conversion to Python "long" integers
 
-function PyObject(i::Union{Int128,BigInt})
+const LongInt = @static (Sys.WORD_SIZE==32) ? Union{Int64,UInt64,Int128,UInt128,BigInt} : Union{Int128,UInt128,BigInt}
+
+function PyObject(i::LongInt)
     PyObject(@pycheckn ccall((@pysym :PyLong_FromString), PyPtr,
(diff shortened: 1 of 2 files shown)


stevengj:
Tests are failing on 32-bit Windows ..

ranocha:
I don't have access to any computer running Windows here. The `InexactError` is raised by the code
```
let i = 1234567890123456789 # Int64
    @test PyObject(i) - i == 0
end
```
There, `i` should be represented as `Int64`; it cannot be represented as `Int32==Int` on 32-bit Windows. This failure is not caused by the changes I did. I think it should be possible to fix it by using the same conversion to long integers for `Int64` if `Int==Int32`. I hope this change fixes it.

src/conversions.jl
```diff
@@ -653,16 +653,25 @@ end
...
-function PyObject(i::BigInt)
+function PyObject(i::Union{Int128,BigInt})
     PyObject(@pycheckn ccall((@pysym :PyLong_FromString), PyPtr,
                              (Ptr{UInt8}, Ptr{Void}, Cint),
                              String(string(i)), C_NULL, 10))
 end
 
 convert(::Type{BigInt}, o::PyObject) = parse(BigInt, pystr(o))
 
+# If 32-bit systems are used, Int64 has to be converted to long integes as well.
```
stevengj: Probably cleaner to do
```jl
const LongInt = @static (Sys.WORD_SIZE==32) ? Union{Int64,UInt64,Int128,UInt128,BigInt} : Union{Int128,UInt128,BigInt}
```
and then do `function PyObject(i::LongInt)`

Issue #470, opened 2018-03-15
@pydef not overloading method correctly


aterenin:
The following code does not work.

```julia
@pydef type SDTExpression <: fenics.Expression
  eval(self, value, x) = begin
    value[1] = 1.0
    return
  end
end
```

The error it gives is as follows.
```
TypeError("expected an overload 'eval' or 'eval_cell' method",)
```

This makes no sense because I've defined one.

aterenin:
To add, the following Python code works perfectly fine.
```
class MyExpression0(Expression):
    def eval(self, value, x):
        value[0] = 1.0
        return
```

stevengj:
Is this because the function name `eval` is special in Julia.  What happens with other method names?

cc @cstjean

aterenin:
I don't think so. The following fails too.

```julia
@pydef type SDTExpression <: fenics.Expression
  eval_cell(self, value, x,y) = begin
    value[1] = 1.0
    return
  end
end
```

Python version works fine.

cstjean:
I don't have time to look into this in detail, but googling the error, it seems to come from a project called [DOLFIN](https://fenicsproject.org/qa/13454/compiled-expression-expected-overload-eval-eval_data-error/)?

```
Calling DOLFIN just-in-time (JIT) compiler, this may take some time.
Traceback (most recent call last):
  File "test_dolfin.py", line 148, in <module>
    spray_exp = Expression(ccp_expression)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/dolfin/functions/expression.py", line 666, in __new__
    return object.__new__(create_compiled_expression_class(cpp_base))
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/dolfin/functions/expression.py", line 187, in create_compiled_expression_class
    {"__init__": __init__})
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/dolfin/functions/expression.py", line 357, in __new__
    raise TypeError("expected an overload 'eval' or 'eval_cell' method")
TypeError: expected an overload 'eval' or 'eval_cell' method
```

It looks like it's trying to JIT compile the function...

aterenin:
DOLFIN is a submodule of a differential equation solver called FEniCS.

[...]

PR #473, opened 2018-03-17
@pydef improvements

cstjean:
I used [`MacroTools.splitdef`](https://github.com/MikeInnes/MacroTools.jl#function-definitions) to clean up the method definition code. This enables support for long-form function definitions (along with parametric types and return types, which might not get much use). I also now create types with a method dictionary. It will help #470.

(Sorry for the other PR; I had forgotten to branch)

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,10 +409,14 @@ For instance,
 
     @pyimport numpy.polynomial as P
     @pydef type Doubler <: P.Polynomial
-        __init__(self, x=10) = (self[:x] = x)
+        function __init__(self, x=10)
+	    self[:x] = x
+	end
         my_method(self, arg1::Number) = arg1 + 20
         x2.get(self) = self[:x] * 2
-        x2.set!(self, new_val) = (self[:x] = new_val / 2)
+        function x2.set!(self, new_val)
+	    self[:x] = new_val / 2
+	end
     end
     Doubler()[:x2]
 
--- a/README.md
+++ b/README.md
@@ -410,13 +410,13 @@ For instance,
     @pyimport numpy.polynomial as P
     @pydef type Doubler <: P.Polynomial
         function __init__(self, x=10)
-	    self[:x] = x
-	end
+            self[:x] = x
+        end
         my_method(self, arg1::Number) = arg1 + 20
         x2.get(self) = self[:x] * 2
         function x2.set!(self, new_val)
-	    self[:x] = new_val / 2
-	end
+            self[:x] = new_val / 2
+        end
     end
     Doubler()[:x2]
 
--- a/src/io.jl
+++ b/src/io.jl
@@ -74,7 +74,7 @@ function pyio_initialize()
     global pyio_initialized
     if !pyio_initialized::Bool
         copy!(PyIO, @pydef_object mutable struct PyIO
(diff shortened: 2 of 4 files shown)


stevengj:
What are all the 0.6 test failures?

cstjean:
Fixed.

README.md
```diff
@@ -409,10 +409,14 @@ For instance,
...
-        __init__(self, x=10) = (self[:x] = x)
+        function __init__(self, x=10)
+	    self[:x] = x
+	end
         my_method(self, arg1::Number) = arg1 + 20
         x2.get(self) = self[:x] * 2
-        x2.set!(self, new_val) = (self[:x] = new_val / 2)
+        function x2.set!(self, new_val)
+	    self[:x] = new_val / 2
+	end
```
stevengj: indent

stevengj: What's wrong with the one-liner definitions?

cstjean:
Fixed indentation.

rdeits:
I think this is causing errors on v0.7 that are being masked by the fact that the 0.7 build was stalling out. Previously, the 0.7 build would hit https://github.com/JuliaLang/julia/issues/26357 but now it errors out earlier: https://travis-ci.org/JuliaPy/PyCall.jl/jobs/357206714#L674

As I understand it, that means that the `with_ioraise` macro is being called with 0 arguments instead of the 1 it's expecting. It's not clear how that could be happening, but I guess macro parsing inside other macros is different in 0.7?

cstjean:
Looks like some interaction with https://github.com/JuliaLang/julia/pull/21746 ?

> All macros receive an extra argument __source__::LineNumberNode which describes the parser location in the source file for the @ of the macro call. It can be accessed as a normal argument variable in the body of the macro. This is implemented by inserting an extra leading argument into the Expr(:macrocall, :@name, LineNumberNode(...), args...) surface syntax.

It shouldn't affect us, though. Maybe it's a bug?

[This line](https://github.com/JuliaPy/PyCall.jl/blob/55648f27682835cbe886b9c6de39a25ad8c8c1bd/src/exception.jl#L68) may have to be changed, but it looks unrelated.

rdeits:
I don't think it's the new line number argument. The error message is: 

```
ERROR: LoadError: LoadError: LoadError: MethodError: no method matching @with_ioraise(::LineNumberNode, ::Module)
Closest candidates are:
  @with_ioraise(::LineNumberNode, ::Module, ::Any) at /home/travis/.julia/v0.7/PyCall/src/io.jl:22
```

which shows that the macro was defined to take the `LineNumberNode`, the `Module`, and then it's "actual" argument `expr::Any`, but was called with just the `LineNumberNode` and `Module`. If it were an issue of the missing LineNumberNode, then the first argument types wouldn't match.

Keno:
What's happening is that pydef now generates definitions like this:
```
    $(Expr(:escape, :(function ##fileno#1531(self; )::Any
      #= /home/keno/.julia/v0.7/MacroTools/src/utils.jl:260 =#
      begin
          @with_ioraise
      end
```

Keno:
I'll have a fix shortly.

Keno:
See https://github.com/MikeInnes/MacroTools.jl/pull/68.

PR #482, opened 2018-04-04
Wrap tests in a testset

JobJob:
First of a few PRs coming up - add `?w=1` or `&w=1` to the url to see this without whitespace changes.

```
Test Summary: | Pass  Total
PyCall        |  424    424
```

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -369,9 +369,9 @@ end
         "foo docstring"
         return 0
     """
-    global foo = py"foo"
+    global foo354 = py"foo"
     # use 'content' since `Text` objects test equality by object identity
-    @test @doc(foo).content == "foo docstring"
+    @test @doc(foo354).content == "foo docstring"
 
     # binary operators
     for b in (4, PyObject(4))
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -13,212 +13,10 @@ roundtrip(x) = roundtrip(PyAny, x)
 roundtripeq(T, x) = roundtrip(T, x) == x
 roundtripeq(x) = roundtrip(x) == x
 
-
-# test handling of type-tuple changes in Julia 0.4
-import PyCall.pyany_toany
-@test pyany_toany(Int) == Int
-@test pyany_toany(PyAny) == Any
-@test pyany_toany(Tuple{Int,PyAny}) == Tuple{Int,Any}
-@test pyany_toany(Tuple{Int,Tuple{PyAny,Int8}}) == Tuple{Int,Tuple{Any,Int8}}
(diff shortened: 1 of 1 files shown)


stevengj:
Why the `invokelatest`?

JobJob:
Seemed to world age error without it. Wasn't sure why.

JobJob:
I moved the function outside the testset and it's ok, I'll push an update.

test/runtests.jl
```diff
@@ -228,310 +26,515 @@ catch
...
-    @test begin
-        @pywith pybuiltin("open")(fname,"w") as f begin
-            f[:write]("test")
+    # Expose python docs to Julia doc system
+    py"""
+    def foo():
+        "foo docstring"
+        return 0
+    """
+    global foo = py"foo"
```
stevengj: Either `let foo=...` or just substitute `py"foo"` directly into the test since `foo` is used in only one place.

stevengj:
Might be nice to group the tests into smaller testsets when there is a comment for a group of tests.

test/runtests.jl
JobJob:
```
using PyCall, BenchmarkTools, Base.Test
@testset "py docs" begin
    # Expose python docs to Julia doc system
    py"""
    def foo():
        "foo docstring"
        return 0
    """
    let foo = py"foo"
    # use 'content' since `Text` objects test equality by object identity
        @test @doc(foo).content == "foo docstring"
    end
end
```
Gives
```
py docs: Test Failed
  Expression: @doc(foo).content == "foo docstring"
   Evaluated: Any[Base.Markdown.Paragraph(Any["No documentation found."]), Base.Markdown.Paragraph(Any["Binding ", Base.Markdown.Code("", "foo"), " does not exist."])] == "foo docstring"
```
using `@doc(py"foo").content`
gives
```
Expression: @doc(@py_str("foo")).content == "foo docstring"
   Evaluated: Any[```
py".....python code....."

Evaluate the given Python code string in the main Python module.
...
```
Since
```
julia> @macroexpand @doc(py"foo").content
:((Base.Docs.doc)((Base.Docs.Binding)((current_module)(), Symbol("@py_str")), Union{Tuple{Any}}).content)
```

Seems like it needs to be defined on the module? Can I just change it to `global the_foo_function_in_python = py"foo"` ?

stevengj: Okay, I didn't realize `@doc` was so fragile.   Let's use `global foo354` then, since this was introduced in #354.

JobJob: I personally think `the_foo_function_in_python` has a better ring to it, possibly with an extra leading underscore, but ok, np 😄

PR #483, opened 2018-04-04
Add PyFuncWrap

JobJob:
Some performance issues were reported here: https://github.com/JuliaML/OpenAIGym.jl/issues/9

This adds a wrapper for a python function (similar to the FuncWrapper in callback.jl for julia functions) so that when calling it in tight loops you reduce the number of PyObjects created to store the arguments and the return value. The included benchmarks show significant increase in calling speed.

```
Mean times
----------
nprand_pyo             : TrialEstimate(16.380 μs)  (original 2-arg function)
nprand2d_wrap          : TrialEstimate(2.071 μs)   (wrapped - args set at call time)
nprand2d_wrap_noargs   : TrialEstimate(1.239 μs)   (wrapped - args previously set)

nprand_pyo2            : TrialEstimate(925.201 μs) (original 15-arg function)
nprand2d_wrap2         : TrialEstimate(2.348 μs)   (wrapped - args set at call time)
nprand2d_wrap_noargs2  : TrialEstimate(1.320 μs)   (wrapped - args previously set)
```

### DIFF
--- a/benchmarks/callperf.jl
+++ b/benchmarks/callperf.jl
@@ -6,40 +6,30 @@ let
     np = pyimport("numpy")
     nprand = np["random"]["rand"]
     nprand_pyo(sz...) = pycall(nprand, PyObject, sz...)
-    nprand2d_wrap = PyFuncWrap(nprand, (Int, Int))
+    ret = PyNULL()
+    args_lens = (0,3,7,12,17)
+    arr_sizes = (ntuple(i->1, len) for len in args_lens)
+    nprand_wraps = [PyFuncWrap(nprand, map(typeof, arr_size)) for arr_size in arr_sizes]
+    @show typeof(nprand_wraps)
+    for (i, arr_size) in enumerate(arr_sizes)
+        nprand_wrap = nprand_wraps[i]
+        arr_size_str = args_lens[i] < 5 ? "$arr_size" : "$(args_lens[i])*(1,1,...)"
+        results["nprand_pyo $arr_size_str"] = @benchmark $nprand_pyo($arr_size...)
+        println("nprand_pyo $arr_size_str:\n"); display(results["nprand_pyo $arr_size_str"])
+        println("--------------------------------------------------")
 
-    arr_size = (2,2)
+        results["nprand_wrap $arr_size_str"] = @benchmark $nprand_wrap($arr_size...)
+        println("nprand_wrap $arr_size_str:\n"); display(results["nprand_wrap $arr_size_str"])
+        println("--------------------------------------------------")
 
-    results["nprand_pyo"] = @benchmark $nprand_pyo($arr_size...)
-    println("nprand_pyo:\n"); display(results["nprand_pyo"])
-    println("--------------------------------------------------")
-
-    results["nprand2d_wrap"] = @benchmark $nprand2d_wrap($arr_size...)
-    println("nprand2d_wrap:\n"); display(results["nprand2d_wrap"])
-    println("--------------------------------------------------")
-
-    # args already set by nprand2d_wrap calls above
-    results["nprand2d_wrap_noargs"] = @benchmark $nprand2d_wrap()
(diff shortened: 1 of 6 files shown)


JobJob:
Not threadsafe (in my possibly flawed mental model) because e.g. say thread 1 is setting some args, then say thread 2 gets scheduled and sets some args, then thread 1 is scheduled again - thread 1 could make the call with thread 2's args.

JobJob:
This method was also faster, I think, can't remember by how much. Let me know if you want me to check.

stevengj:
I don't think the CPython API is thread-safe anyway, but if we were worried about that we could always use a per-thread array similar to JuliaLang/julia#26562.

stevengj:
In particular, if we want to make `pycall` thread-safe we will need to [acquire the global interpreter lock (GIL) anyway](https://docs.python.org/2/c-api/init.html#non-python-created-threads), and once we have a lock we can manipulate our own global state safely too.

JobJob:
Yeah, all good points. I will benchmark the difference between this and a `pycall!` with global set of args.
I'm not sure if every call could share the `const pycall_ret = PyNULL()`,I think I'd just get `pycall(...)` to call `pycall!(PyNULL(),...)` - and if users want they can call `pycall!` directly with their own  `ret::PyObject`. I have an earlier commit with this, so I'll check it out.

JobJob:
You can check out a version of `pycall!` at https://github.com/JobJob/PyCall.jl/blob/pycall-bang/src/PyCall.jl#L674-L716

The benchmark results (from that branch - shown below) suggest that the wrap in this PR is faster. Having said that, I'm not really sure why the `pycall!` version is benchmarking slower - it's pretty similar. Presumably the extra typing is helping somehow?

A cleaned up version of `pycall!` could be done in a separate PR, but I don't think I'll have time for that right now, so anyone interested should feel free to adapt it to something merge worthy.

```
Mean times
----------
nprand_pyo ()                    : TrialEstimate(943.182 ns) (pycall version)
nprand_pyo! ()                   : TrialEstimate(589.341 ns) (pycall! version)
nprand_wrap ()                   : TrialEstimate(426.096 ns) (wrapped version)
nprand_wrap_noargs ()            : TrialEstimate(427.233 ns) (wrapped version - args already set)

nprand_pyo (1, 1, 1)             : TrialEstimate(18.899 μs)
nprand_pyo! (1, 1, 1)            : TrialEstimate(6.176 μs)
nprand_wrap (1, 1, 1)            : TrialEstimate(2.215 μs)
nprand_wrap_noargs (1, 1, 1)     : TrialEstimate(1.133 μs)

nprand_pyo 7*(1,1,...)           : TrialEstimate(34.965 μs)
nprand_pyo! 7*(1,1,...)          : TrialEstimate(7.714 μs)
nprand_wrap 7*(1,1,...)          : TrialEstimate(2.786 μs)
nprand_wrap_noargs 7*(1,1,...)   : TrialEstimate(1.338 μs)

nprand_pyo 12*(1,1,...)          : TrialEstimate(50.625 μs)
nprand_pyo! 12*(1,1,...)         : TrialEstimate(9.137 μs)
nprand_wrap 12*(1,1,...)         : TrialEstimate(4.344 μs)
nprand_wrap_noargs 12*(1,1,...)  : TrialEstimate(1.517 μs)

nprand_pyo 17*(1,1,...)          : TrialEstimate(68.642 μs)
nprand_pyo! 17*(1,1,...)         : TrialEstimate(11.808 μs)
nprand_wrap 17*(1,1,...)         : TrialEstimate(5.610 μs)
nprand_wrap_noargs 17*(1,1,...)  : TrialEstimate(1.726 μs)
```

stevengj:
I'd rather have the time spent on making `pycall` and `pycall!` faster than on maintaining a separate function-wrapper type.

JobJob:
So, you've managed to sucker me in to cleaning up my `pycall!` branch, well played old sport, should have something to push soonish.

JobJob:
Closing in favour of #492

PR #485, opened 2018-04-04
Misc fixes

JobJob:
* Avoid using exceptions for control flow when a PyObject doesn't have the `"__array_interface__"` attribute. Provides 5-6x speed up for PyAny conversions for heterogenous lists: `@benchmark convert(PyAny, PyObject([1,2,"c"]))`
  * **Was that the only exception that try block was supposed to catch?**
* Avoid double check for isdict in PyDict constructor
* Couple of other minor changes


stevengj:
Sounds good to me.   (To really speed up PyAny conversions we need some kind of memo-izing mechanism based on `pytypeof`.)

PR #487, opened 2018-04-04
PyArray conversion speedups and PyArrayFromBuffer

JobJob:
This is the last one for now 😄

Speedups for PyArray conversion using Python's Buffer interface. `PyArrayFromBuffer` could become the default constructor for PyArray, if this looks good.

* ~15x speedup for `PyArray(o::PyObject)`
* `PyArrayFromBuffer(o::PyObject)` further 4x faster PyArray creation (~60x faster than master)
* `setdata!` for fast reassignment of data ptr in PyArray when you know the array is the right shape (~500x faster than creating a new PyArray on master - not an apples to apples comparison, but still good to have)
* Workhorse is `PyArrayInfoFromBuffer(o::PyObject)` - a faster way to get PyArray_Info from numpy than extracting values from the `__array_interface__`
* (speculative) ArrayFromBuffer no copy conversion to Julia Array
* PyArray_Info parameterised T,N and immutable
* PyArray_Info size and stride changed to Tuples (from Vectors)
* PyArray_Info data changed from Ptr{Cvoid} to Ptr{T}
* moved PyBuffer tests to separate file

In no major rush for review, so please don't feel pressured.

---
master results:
```
Mean times
----------
nprand_pyo(2, 2)    : TrialEstimate(17.451 μs)
convert_pyarr(2, 2) : TrialEstimate(403.306 μs)
nprand_pyo(100, 100): TrialEstimate(152.220 μs)
convert_pyarr(100, 100): TrialEstimate(418.158 μs)
```
This branch results:
```
Mean times
----------
nprand_pyo(2, 2)           : TrialEstimate(16.688 μs)
convert_pyarr(2, 2)        : TrialEstimate(21.455 μs)
PyArrayInfoFromBuffer(2, 2): TrialEstimate(3.519 μs)
convert_pyarrbuf(2, 2)     : TrialEstimate(5.450 μs)
convert_arr(2, 2)          : TrialEstimate(100.200 μs)
convert_arrbuf(2, 2)       : TrialEstimate(10.339 μs)
setdata!(2, 2)             : TrialEstimate(745.279 ns)
setdata! bufprealloc(2, 2) : TrialEstimate(469.487 ns)
nprand_pyo(100, 100)       : TrialEstimate(158.003 μs)
convert_pyarr(100, 100)    : TrialEstimate(21.049 μs)
PyArrayInfoFromBuffer(100, 100): TrialEstimate(3.402 μs)
convert_pyarrbuf(100, 100) : TrialEstimate(5.024 μs)
convert_arr(100, 100)      : TrialEstimate(4.559 ms)
convert_arrbuf(100, 100)   : TrialEstimate(10.037 μs)
setdata!(100, 100)         : TrialEstimate(745.049 ns)
setdata! bufprealloc(100, 100): TrialEstimate(465.865 ns)
```

Benchmark code for master.
```
using PyCall, BenchmarkTools, DataStructures

results = OrderedDict{String,Any}() 
let 
    np = pyimport("numpy") 
    nprand = np["random"]["rand"] 
    # nparray_pyo(x) = pycall(np["array"], PyObject, x) 
    # pytestarray(sz::Int...) = pycall(np["reshape"], PyObject, nparray_pyo(1:prod(sz)), sz) 

    nprand_pyo(sz...)   = pycall(nprand, PyObject, sz...) 

    for arr_size in [(2,2), (100,100)] 
        pyo_arr = nprand_pyo(arr_size...) 
        results["nprand_pyo$arr_size"] = @benchmark $nprand_pyo($arr_size...) 
        println("nprand_pyo $arr_size:\n"); display(results["nprand_pyo$arr_size"]) 
        println("--------------------------------------------------") 

        results["convert_pyarr$arr_size"] = @benchmark $convert(PyArray, $pyo_arr) 
        println("convert_pyarr $arr_size:\n"); display(results["convert_pyarr$arr_size"]) 
        println("--------------------------------------------------")
    end
end

println("")
println("Mean times")
println("----------")
foreach((r)->println(rpad(r[1],20), ": ", mean(r[2])), results)
```

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -5,7 +5,8 @@ module PyCall
 using Compat, VersionParsing
 
 export pycall, pyimport, pyimport_e, pybuiltin, PyObject, PyReverseDims,
-       PyPtr, pyincref, pydecref, pyversion, PyArray, PyArray_Info,
+       PyPtr, pyincref, pydecref, pyversion,
+       PyArray, PyArray_Info, PyBuffer, PyArrayFromBuffer, setdata!, ArrayFromBuffer,
        pyerr_check, pyerr_clear, pytype_query, PyAny, @pyimport, PyDict,
        pyisinstance, pywrap, pytypeof, pyeval, PyVector, pystring, pystr, pyrepr,
        pyraise, pytype_mapping, pygui, pygui_start, pygui_stop,
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -6,7 +6,7 @@ using Compat, VersionParsing
 
 export pycall, pyimport, pyimport_e, pybuiltin, PyObject, PyReverseDims,
        PyPtr, pyincref, pydecref, pyversion,
-       PyArray, PyArray_Info, PyBuffer, PyArrayFromBuffer, setdata!, ArrayFromBuffer,
+       PyArray, PyArray_Info, PyBuffer, PyArrayFromBuffer, setdata!, ArrayFromBuffer, isbuftype,
        pyerr_check, pyerr_clear, pytype_query, PyAny, @pyimport, PyDict,
        pyisinstance, pywrap, pytypeof, pyeval, PyVector, pystring, pystr, pyrepr,
        pyraise, pytype_mapping, pygui, pygui_start, pygui_stop,
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -6,7 +6,7 @@ using Compat, VersionParsing
 
 export pycall, pycall!, pyimport, pyimport_e, pybuiltin, PyObject, PyReverseDims,
        PyPtr, pyincref, pydecref, pyversion,
-       PyArray, PyArray_Info, PyBuffer, setdata!, NoCopyArray, isbuftype,
+       PyArray, PyArray_Info, PyBuffer,
        pyerr_check, pyerr_clear, pytype_query, PyAny, @pyimport, PyDict,
        pyisinstance, pywrap, pytypeof, pyeval, PyVector, pystring, pystr, pyrepr,
        pyraise, pytype_mapping, pygui, pygui_start, pygui_stop,
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -6,7 +6,7 @@ using Compat, VersionParsing
 
 export pycall, pyimport, pyimport_e, pybuiltin, PyObject, PyReverseDims,
        PyPtr, pyincref, pydecref, pyversion,
-       PyArray, PyArray_Info, PyBuffer, PyArrayFromBuffer, setdata!, ArrayFromBuffer, isbuftype,
+       PyArray, PyArray_Info, PyBuffer, setdata!, NoCopyArray, isbuftype,
        pyerr_check, pyerr_clear, pytype_query, PyAny, @pyimport, PyDict,
        pyisinstance, pywrap, pytypeof, pyeval, PyVector, pystring, pystr, pyrepr,
        pyraise, pytype_mapping, pygui, pygui_start, pygui_stop,
@@ -174,6 +174,7 @@ PyObject(o::PyPtr, keep::Any) = pyembed(PyObject(o), keep)
 
 const TypeTuple = Union{Type,NTuple{N, Type}} where {N}
 include("pybuffer.jl")
+include("pyarray.jl")
 include("conversions.jl")
 include("pytype.jl")
 include("pyiterator.jl")
--- a/src/pyarray.jl
+++ b/src/pyarray.jl
@@ -1,0 +1,317 @@
+#########################################################################
+# Extract shape and other information about arrays that support Python's
+# Buffer Interface/Protocol (PEP 3118)
+#########################################################################
+struct PyArray_Info{T,N}
+    native::Bool # native byte order?
+    sz::NTuple{N,Int}
+    st::NTuple{N,Int} # strides, in multiples of bytes!
+    data::Ptr{T}
+    readonly::Bool
+end
+
+function PyArray_Info(o::PyObject)
+    # n.b. the pydecref(::PyBuffer) finalizer handles releasing the PyBuffer
+    pybuf = PyBuffer(o, PyBUF_ND_CONTIGUOUS)
+    T, native_byteorder = array_format(pybuf)
+    sz = size(pybuf)
+    strd = strides(pybuf)
+    length(strd) == 0 && (sz = ())
+    N = length(sz)
+    isreadonly = pybuf.buf.readonly==1
+    return PyArray_Info{T,N}(native_byteorder, sz, strd, pybuf.buf.buf, isreadonly)
+end
+
+aligned(i::PyArray_Info{T,N}) where {T,N} = #  FIXME: also check pointer alignment?
+  all(m -> m == 0, mod.(i.st, sizeof(T))) # strides divisible by elsize
+
+eltype(i::PyArray_Info{T,N}) where {T,N} = T
+ndims(i::PyArray_Info{T,N})  where {T,N} = N
+
+function default_stride(sz::NTuple{N, Int}, ::Type{T}) where {T,N}
+    stv = Vector{Int}(N)
+    stv[end] = sizeof(T)
+    for i = N-1:-1:1
+        stv[i] = stv[i+1]*sz[i+1]
+    end
+    ntuple(i->stv[i], N)
+end
+
+# whether a contiguous array in column-major (Fortran, Julia) order
+function f_contiguous(T::Type, sz::NTuple{N,Int}, st::NTuple{N,Int}) where N
+    if prod(sz) == 1 || length(sz) == 1
+        # 0 or 1-dim arrays should default to f-contiguous in julia
+        return true
+    end
+    if st[1] != sizeof(T)
+        return false
+    end
+    for j = 2:N
+        if st[j] != st[j-1] * sz[j-1]
+            return false
+        end
+    end
+    return true
+end
+
+f_contiguous(T::Type, sz::NTuple{N1,Int}, st::NTuple{N2,Int}) where {N1,N2} =
+    error("stride and size are different lengths, size: $sz, strides: $sz")
+
+f_contiguous(i::PyArray_Info{T,N}) where {T,N} = f_contiguous(T, i.sz, i.st)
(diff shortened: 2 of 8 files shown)


stevengj:
That is, `PyArray` (the no-copy conversion type) should just use the buffer API.

JobJob:
Yeah I think you can remove all the stuff related to Python Array -> PyArray conversion. But this doesn't do anything for julia array -> python array

JobJob:
as in everything related to `__aray_interface__`

stevengj:
Right.... for that we'd need to create an object that implements the buffer API.   One concern is that I don't know how well this would work with Python functions expecting a NumPy array.

Okay, it makes sense to replace the Python -> Julia array code with buffer code in this PR, but leave the Julia -> Python stuff using NumPy for now and work on that in a separate PR.

JobJob:
The other issue with replacing `PyArray(o::PyObject)` with `PyArrayFromBuffer` is if something implements `__array_interface__` but not the buffer interface. Not sure which is more supported in the wild? But we could just add a fallback to `__array_interface__` if the buffer interface isn't implemented.

stevengj:
> if something implements `__array_interface__` but not the buffer interface. 

I'm guessing that this never happens in practice.

JobJob:
> I'm guessing that this never happens in practice.

Ok, let's not support it then unless someone tells us it does 😄

JobJob:
What's the issue with having numpy as a test dependency? Seems that using numpy will allow us to cover more cases, more easily.

I thought I might be able to get by with python's [array module](https://docs.python.org/3/library/array.html), but then I did a little digging and found out that it doesn't support the buffer interface on Python 2.7 🤦. You can get typed byte data with [struct.pack](https://docs.python.org/2/library/struct.html#struct.pack), but you can't reshape a memoryview using `cast` in py2 (you can in Python 3). So, the only way I could see to create a multi-dim array without using numpy in py2.7 is messing around in ctypes.

It could be done, but would take effort, and since I'm guessing the majority of the arrays these functions will see in the wild will be coming from numpy anyway (?), could be good to test with it.

stevengj:
NumPy as a test dependency is fine; we should definitely test with NumPy arrays and operations.

stevengj:
Bump.

JobJob:
Assuming the tests pass I think this is pretty good to go. (they pass for me on mac with py 2/3 julia 0.6.3/0.7-dev-xxx)

There are things in pep-3118 that it doesn't handle, but I think it handles all the cases of arrays (strides and types) that current master does.

Oh the one thing I wanted to ask about is there's a deprecation warning about `strides` on 0.7 - so pyarray should implement strides?

JobJob:
Umm nightly failures seem unrelated :/

stevengj:
Nightly failures on `master` should be fixed now if you want to rebase.

JobJob:
I think this is good to go now. Unlikely this PR is causing the Conda failures on Windows.

JobJob:
@stevengj tests pass. Can we get this over the line?

src/pybuffer.jl
```diff
@@ -37,6 +37,13 @@ mutable struct PyBuffer
     end
 end
 
+"""
+`pydecref(o::PyBuffer)`
+Release the reference to buffer `o`
+N.b. As per https://docs.python.org/3/c-api/buffer.html#c.PyBuffer_Release,
+It is an error to call this function on a PyBuffer that was not obtained via
+the python c-api function `PyObject_GetBuffer()`
```
tkf: Note that this docstring (which is "correct" according to cpython documentation) is in contradiction to the comment _`a no-op if o.obj is already NULL`_ just below in the source code (which reflects the actual cpython implementation).

IIUC, what is written in the comment is actually important for understanding why [`PyBuffer(o::Union{PyObject,PyPtr}, flags=PyBUF_SIMPLE)`](https://github.com/JuliaPy/PyCall.jl/pull/487/files#diff-4ca60bdd8276bb88144aeefa48f65a6dR118) is correct since it essentially does `pydecref(PyBuffer())`. If that's the case, maybe clarify this point in the docstring?

JobJob: Good point. Updated.

JobJob: Good point. Updated.

JobJob: Good point. Updated.

JobJob: Lol - github seems to be having some sync issues. At least I've made my agreement very clear 😄

src/pyarray.jl
```diff
@@ -0,0 +1,329 @@
...
+    if a.f_contig
+        ccall(:memcpy, Cvoid, (Ptr{T}, Ptr{T}, Int), A, a, sizeof(T)*length(a))
+        return A
+    else
+        return copyto!(A, a)
+    end
+end
+
+# TODO: need to do bounds-checking of these indices!
+
```
stevengj: These `unsafe_load` calls are also not rooting `a` properly… in Julia 0.7 or later, they should all be of the form `GC.@preserve a unsafe_load(...)`

src/pybuffer.jl
```diff
@@ -86,14 +93,16 @@ function Base.stride(b::PyBuffer, d::Integer)
     return Int(unsafe_load(b.buf.strides, d))
```
stevengj: all of these `unsafe_load` calls need rooting… but that is not the fault of this PR

src/PyCall.jl
```diff
@@ -5,7 +5,8 @@ module PyCall
 using Compat, VersionParsing
 
 export pycall, pycall!, pyimport, pyimport_e, pybuiltin, PyObject, PyReverseDims,
-       PyPtr, pyincref, pydecref, pyversion, PyArray, PyArray_Info,
+       PyPtr, pyincref, pydecref, pyversion,
+       PyArray, PyArray_Info, PyBuffer, setdata!, NoCopyArray, isbuftype,
```
stevengj: I'd prefer not to export `setdata!, NoCopyArray, isbuftype`

src/pyarray.jl
JobJob: I don't fully follow here, wouldn't user code always have a reference to the `a` that they're calling getindex on? So the array would be GC rooted that way?

test/testpybuffer.jl
```diff
@@ -0,0 +1,114 @@
...
+            @test isbuftype(PyObject([1:10...])) == true
+        end
+
+        # TODO maybe move these to a test_pyarray.jl
+        @testset "setdata!" begin
+            ao1 = arrpyo(1.0:10.0, "d")
+            pyarr = convert(PyArray, ao1)
+            ao2 = arrpyo(11.0:20.0, "d")
+            setdata!(pyarr, ao2)
+            @test all(pyarr[1:10] .== 11.0:20.0)
```
JobJob:
```
using PyCall
np = pyimport("numpy")
pyarr = pycall(np["array"], PyArray, [1:10;])
pyarr[1:2]
```
The above is an error on current master on Julia 1.0.1 🙀

src/pyarray.jl
tkf: @JobJob I'm not sure how Julia's GC works but if I do `np.zeros(100)[1]` there is no Julia variable referencing the array so I guess it could be GC'ed away?

JobJob: Ahh, yeah that seems plausible, added a TODO

JobJob:
Bumpity bump? @stevengj

JobJob:
Hello?

stevengj:
Thanks for keeping at this!

JobJob:
🎉 🎉 🎉 
Thanks to you @stevengj you brilliant beast 😄 
Sorry the diff became so big - will try to keep things more manageable in future.

JobJob:
I think it could be related to 48d730f293974f67543f2d5ef76173e4752cc30f
I can replicate on 0.7 with master locally, but when I reset  to 17dcb37441df2e68661c197049d88437bd3e0b18 on the branch of this PR (array-perf on https://github.com/JobJob/PyCall.jl) things are fine - i.e. no segfault at the end of testing

JobJob:
N.b. I didn't merge that commit into the branch of this PR

JobJob:
Simple fix in #615

marius311:
Apologies, I haven't followed the discussion above, but I did bisect a failure of my code when upgrading to PyCall#master down to this commit. It seems that before this commit, 

```
julia> using PyPlot
julia> subplots(1, 1; squeeze=false)[2]
1×1 Array{PyCall.PyObject,2}:
 PyObject <matplotlib.axes._subplots.AxesSubplot object at 0x7f75f95c7610>
```

but after

```
julia> using PyPlot
julia> subplots(1, 1; squeeze=false)[2]
PyObject array([[<matplotlib.axes._subplots.AxesSubplot object at 0x7ff0846ed610>]],
      dtype=object)
```

Was this intended or not? 

In either case, is there a workaround to force conversion to a Julia `Matrix` in the second case?

stevengj:
Probably not intended.  Have you tried `convert(Matrix, ans)` on the result?

marius311:
Thanks. Neither of these two attempts on current master seem to work:

```
julia> convert(Matrix, subplots(1,1;squeeze=false)[2])
ERROR: MethodError: Cannot `convert` an object of type PyObject to an object of type Array{T,2} where T
Closest candidates are:
  convert(::Type{T<:Array}, ::AbstractArray) where T<:Array at array.jl:474
  convert(::Type{T<:AbstractArray}, ::T<:AbstractArray) where T<:AbstractArray at abstractarray.jl:14
  convert(::Type{T<:AbstractArray}, ::LinearAlgebra.Factorization) where T<:AbstractArray at /home/marius/src/julia/usr/share/julia/stdlib/v1.1/LinearAlgebra/src/factorization.jl:46
  ...
Stacktrace:
 [1] top-level scope at none:0

julia> PyArray(subplots(1,1;squeeze=false)[2])
ERROR: KeyError: key "O" not found
Stacktrace:
 [1] getindex at ./dict.jl:478 [inlined]
 [2] array_format(::PyBuffer) at /home/marius/.julia/dev/PyCall/src/pybuffer.jl:261
 [3] PyArray_Info(::PyObject) at /home/marius/.julia/dev/PyCall/src/pyarray.jl:17
 [4] PyArray(::PyObject) at /home/marius/.julia/dev/PyCall/src/pyarray.jl:123
 [5] top-level scope at none:0
```

But `permutedims(getindex.(Ref(_), 1:m, 1:n))` works fine so I'm happy with that workaround.

PR #492, opened 2018-04-13
Faster pycall. Adds pycall!

JobJob:
Speeds up pycalls. For functions with 1-7 arguments it reduces the overhead of calling the function from Julia by about 10-40x

Adds `pycall!(ret, o, ...)` to avoid allocating a PyObject for the return value each time the function is called.

One query: I'm not 100% about the `const pyarg_tuples = PyPtr[]`, and growing it when functions have more arguments. Maybe we should just initialise it to a certain size (32 arguments?), check for `isassigned(pyarg_tuples, nargs+1)`, and if not assigned, create the tuple, and grow the `pyarg_tuples` vector if someone requires more than the 32 arg default.

~For reasons that I'll need to track down over the weekend, one of the tuples that we use for passing args, is ending up with a refcount of 2, which is not allowed when setting an item, since tuples are supposed to be immutable (but when there's only one ref, setting an item is allowed):~
~https://github.com/python/cpython/blob/23ab5ee667a9b29014f6f7f01797c611f63ff743/Objects/tupleobject.c#L166
so this is WIP until I solve that.~

~Happens after the `@pydef type Doubler`, see the `@test_broken` in runtests. Any ideas appreciated.~

Creating new builtin exceptions increment the refcount of the arguments tuple passed to it. The Doubler class in that test was subclassing `AssertionError` and instantiating triggered this. Handled this situation (callees incrementing the refcount of args tuple) now.

to-do:

- [x] fix a bug in the wrap ~and possibly some others about.~ (there was another minor bug in the test for having the only reference to the (empty) arguments tuple)
- [x] maybe remove the wrap, since maybe it's not faster after all... it's not, always. Let's remove it for now.
- [ ] remove pycall_legacy when everyone's happy with the PR (was just left in for benchmarking comparison purposes)

---
Benchmark results (Mean times for different arities)
```
pycall_legacy ()                 	 TrialEstimate(946.804 ns) (on master when this PR was opened)
pycall ()                        	 TrialEstimate(725.172 ns) (this PR)
pycall! ()                       	 TrialEstimate(566.083 ns) (this PR, pycall!) 

pycall_legacy (1,)               	 TrialEstimate(8.582 μs)
pycall (1,)                      	 TrialEstimate(1.454 μs)
pycall! (1,)                     	 TrialEstimate(1.262 μs)

pycall_legacy (1, 1)             	 TrialEstimate(12.365 μs)
pycall (1, 1)                    	 TrialEstimate(1.657 μs)
pycall! (1, 1)                   	 TrialEstimate(1.684 μs)

pycall_legacy (1, 1, 1)          	 TrialEstimate(16.136 μs)
pycall (1, 1, 1)                 	 TrialEstimate(1.694 μs)
pycall! (1, 1, 1)                	 TrialEstimate(1.764 μs)

pycall_legacy 7*(1,1,...)        	 TrialEstimate(29.335 μs)
pycall 7*(1,1,...)               	 TrialEstimate(2.881 μs)
pycall! 7*(1,1,...)              	 TrialEstimate(2.794 μs)

### CODE REVIEW

File: src/PyCall.jl [hunk 1]
```diff
@@ -663,11 +664,12 @@ function pybuiltin(name)
 end
 
 #########################################################################
+include("pycalls.jl")
 
 """
 Low-level version of `pycall(o, ...)` that always returns `PyObject`.
 """
-function _pycall(o::Union{PyObject,PyPtr}, args...; kwargs...)
+function _pycall_legacy(o::Union{PyObject,PyPtr}, args...; kwargs...)
```
JobJob: This is just here for the benchmarks, the whole function, and the `pycall_legacy` methods below, need to be removed before merging.

JobJob:
Anything else you need doing for this (and other open PRs)? Test failures on Windows here to do with importing NumPy, not sure what to do for those? 0.7 failures seem unrelated.

JobJob:
@stevengj any news here?

Should I bother trying to rebase #487 (again) yet? Or wait a bit for more 0.7 related changes? Seems this and the others don't have merge conflicts (so I rebased them).

Apologies if the file rearranging made this or the others harder to review. I can revert if needed I suppose.

stevengj:
Looks like this is breaking nightly?

JobJob:
Hmm, I can't replicate the nightly errors locally, both w julia on the commit that travis used `8e7e6fc00a`, and master from yesterday, tests pass with `juliadev -e 'Pkg.test("PyCall"; coverage=true)'`

stevengj:
Don't rely on numpy features in the tests, or only test the numpy aspects if `PyCall.npy_initialized` is true.

JobJob:
I've changed the tests to not use NumPy anymore.

However, for #487's most of the tests need numpy, so I'll just not run those if `pyimport("numpy")` fails.  N.b. `PyCall.npy_initialized` is always false while running the tests on my machine - so I'll just use `try  pyimport("numpy") catch`.

stevengj:
> Btw PyCall.npy_initialized is always false while running the tests

It's false in the beginning because numpy is loaded lazily, but it is true once array tests start.

However, you can also do `ispynull(pyimport_e("numpy"))`

src/pyfncall.jl
```diff
@@ -0,0 +1,168 @@
...
+    _pycall!(ret, o, args, length(args), kw)
+end
+
+"""
+Low-level version of `pycall!(ret, o, ...)` for when `kw` is already in python
+friendly format but you don't have the python tuple to hold the arguments
+(`pyargsptr`). Sets `ret.o` to the result of the call, and returns `ret::PyObject`.
+"""
+function _pycall!(ret::PyObject, o::Union{PyObject,PyPtr}, args,
+  nargs::Int=length(args), kw::Union{Ptr{Nothing}, PyObject}=C_NULL)
```
stevengj: Shouldn't this be a `PyPtr` rather than `Ptr{Nothing}`?

JobJob: No, that's for the `C_NULL` when there aren't any kwargs. Which reminded me to add some tests for the kwargs case.

JobJob:
"Everyone" happy with this? Should I remove the `pycall_legacy` stuff?

JobJob:
Oh and btw re
> It's false in the beginning because numpy is loaded lazily, but it should be true after other array tests execute.

I thought I *was* checking late in the tests, but perhaps not, it's fine now.

src/pyfncall.jl
iamed2: `Ptr{Nothing}` is an implementation detail; it's best to use `Ptr{Cvoid}` (Base in 0.7, available from Compat in 0.6)

JobJob: ta

stevengj:
Yes, go ahead and remove `pycall_legacy`

PR #495, opened 2018-04-20
Better link?

PallHaraldsson:
jverzani's is currently 1 commit behind.
[skip ci]


PallHaraldsson:
I can see how you want to give credit; I don't know the history, I assume was moved like your https://github.com/stevengj/PyPlot.jl while that one redirects automatically, so I didn't bother changing.

Maybe having lines as long is pointless, so I just changed a few.

Issue #481, opened 2018-04-03
When should npyinitialize be called?


JobJob:
Am I, mr user, supposed to call `npyinitialize()`? I'm guessing not, so this might be a bug.

Seems `asscalar`, and in turn `convert(Int, numpy_array_element::PyObject)`, depends on `npy_number` being initialized
```
using PyCall
@pyimport numpy as np
pyoarr = pycall(np.array, PyObject, 1:3)
el = get(pyoarr, PyObject, 0)

@show pytypeof(el) pytypeof(PyCall.asscalar(el)) PyCall.npy_number
convert(Int, el)
```
outputs
```
pytypeof(el) = PyObject <class 'numpy.int64'>
pytypeof(PyCall.asscalar(el)) = PyObject <class 'numpy.int64'>  (should be int)
PyCall.npy_number = PyObject NULL  (shouldn't be null for asscalar to work)
ERROR: PyError (ccall(@pysym(PyInt_AsSsize_t), Int, (PyPtr,), asscalar(po))) <class 'TypeError'>
TypeError('an integer is required',)
```

stevengj:
No, you are never supposed to call it.  You shouldn't be calling internal functions like `PyCall.asscalar` either.

You can just use 
```jl
julia> pycall(np.asscalar, PyObject, el)
PyObject 1
```
for example.

JobJob:
but is 
```
using PyCall
@pyimport numpy as np
pyoarr = pycall(np.array, PyObject, 1:3)
el = get(pyoarr, PyObject, 0)
convert(Int, el)
```
supposed to work? That calls asscalar, and PyCall.npy_number is PyNULL then

stevengj:
Oh, right, I see the problem.

It actually works fine for me in Python 2.7:
```jl
julia> convert(Int, el)
1
```
but my recollection from 9304bbc5fe6fa00137e1949f104acfc86bedb107 was that the `asscalar` is needed in Python 3.   The current code works if you get numpy arrays via Julia's automatic conversions, but not if you import numpy manually yourself.

JobJob:
Ok cheers, yep I'm on Python 3. Feel free to close this.

stevengj:
No, Python 3 is supported too, so I need to fix this.

JobJob:
Ok, just fyi the actual use case that landed me here was fairly contrived. I was just testing converting an array to a tuple, it wasn't part of any actual work/project.
```
using PyCall
[...]

PR #499, opened 2018-04-29
bugfix in numpy scalars

stevengj:
See #481, cc @JobJob.


stevengj:
The tests should exercise python3+numpy on Linux now... test failures on 0.7 are unrelated, waiting for a MacroTools update.

PR #505, opened 2018-06-09
Fix more 0.7 deprecations

stevengj:


stevengj:
@tkelman, the failure on Travis is somewhat mysterious to me.  Is there any way to get more test output?  e.g. is it failing when precompiling?

tkelman:
signal 9? maybe oom killed? try a debug build and ssh'ing in. and/or adding `sudo dmesg` output on failure

stevengj:
Thanks for the suggestion @tkelman; doing a debug build means messing around with Travis authorization tokens, unfortunately, but I may have to resort to that if I can't figure out how to reproduce the problem offline.

stevengj:
It looks like defining all the constructor methods explicitly was causing compilation resources to explode in 0.6... hopefully it is now fixed.

PR #509, opened 2018-07-20
Add conversion for IPv4/IPv6 types

maxmouchet:
Hi,

This adds support for converting between Julia `IPv4`/`IPv6` types and Python `IPv4Address`/`IPv6Address`.

For some reasons I could not cast  `IPv6Address._ip` directly to `UInt128`. As a workaround I did `UInt128(BigInt(o["_ip"]))`.

I don't know if this is useful enough to be included into PyCall, but anyway, here is the code.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -19,6 +19,7 @@ import Base: size, ndims, similar, copy, getindex, setindex!, stride,
        next, filter!, hash, splice!, pop!, ==, isequal, push!,
        append!, insert!, prepend!, unsafe_convert
 import Compat: pushfirst!, popfirst!, firstindex, lastindex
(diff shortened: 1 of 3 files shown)


maxmouchet:
This is failing on Python 2, as `ipaddress` has been introduced in Python 3. What would be the best way to check the Python version ?

stevengj:
You can use `pyversion` to check the Python version.  However, I'm not sure this is of general enough utility to put into PyCall … is there a package where you need this?

maxmouchet:
Yes indeed, the use cases are quite limited. I just use it in my own code.
No problem if this is out of scope, I'll put in a package of my own instead.

Issue #435, opened 2017-08-28
Make PyCall default to up-to-date most secure Python; meaning python3 (possibly python2) through Conda


PallHaraldsson:
You may not care to support python3 over python2, just consider it; distros e.g. Ubuntu are mainly using python3 now (I'm worried python2 is left behind security wise.. maybe also python3? at least feature wise in Ubuntu).

I realize python 3[.6] by default is a breaking change, so you would just update PyCall's version number and people can always override to [distro's] python2. It's unclear that you can support newest version python of 3 or 2, if not installed; except through Conda.

I'm looking into the code to make a PR:

`let py = get(ENV, "PYTHON", isfile("PYTHON") ? readchomp("PYTHON") :`

`isfile("PYTHON")` is cryptic, a way to support python on (case-insensitive) Windows?

https://docs.python.org/3.6/whatsnew/3.6.html
"secrets: PEP 506 – Adding A Secrets Module To The Standard Library.
[..]
The hashlib and ssl modules now support OpenSSL 1.1.0."

Note that there's a new RC for 2.7.14 (Ubuntu already missing out on 2.7.13 and 12?):

https://raw.githubusercontent.com/python/cpython/c707893f9cee870bba8364b3a06eb9cfa3b80e58/Misc/NEWS

shell> python
Python 2.7.12 (default, Nov 19 2016, 06:48:10) [..]

shell> python3
Python 3.5.2 (default, Nov 17 2016, 17:05:23) [..]

Changelog for python[2] in Ubuntu is deceiving or really at 2.7.11 but some of 2.7.12 backported?:

python-defaults (2.7.11-1) unstable; urgency=medium

  * Bump version to 2.7.11.
  * Stop shipping menu files.
  * Fix some lintian warnings.

 -- Matthias Klose <doko@debian.org>  Wed, 09 Dec 2015 13:07:35 +0100

python3-defaults (3.5.1-3) unstable; urgency=medium

  * Drop build dependency on python3.4.
  * Bump standards and debhelper versions.

 -- Matthias Klose <doko@debian.org>  Mon, 21 Mar 2016 12:58:34 +0100

PallHaraldsson:
I see, as with PyCall, Conda supports now Python 3 and 2: https://github.com/JuliaPy/Conda.jl/commit/65d2b37d8f0b3d5a83b10e397f43aec3ef48be09

Both are done through ENV variable. I'm not sure if that is the best option; if it works to use Python 2 and 3 together.

[...]

PR #512, opened 2018-07-25
default to python3 on Unix

stevengj:
Now that Conda defaults to Python 3 (JuliaPy/Conda.jl#108), to be consistent PyCall should default to `python3` on non-Mac Unices.

Python 2 is still supported, you just need to configure PyCall with `ENV["PYTHON"]="python2"` or similar.  Existing installations will not be affected because PyCall remembers what Python you last configured with.

See also #435.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ private to Julia (not in your `PATH`).  You can use the `Conda` Julia
 package to install more Python packages, and `import Conda` to print
 the `Conda.PYTHONDIR` directory where `python` was installed.
 On GNU/Linux systems, PyCall will default to using
-the `python` program (if any) in your PATH.
+the `python3` program (if any, otherwise `python`) in your PATH.
 
 The advantage of a Conda-based configuration is particularly
 compelling if you are installing PyCall in order to use packages like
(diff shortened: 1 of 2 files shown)


PallHaraldsson:
I see you took my advice, but only partly.

Since you did, should Python 3.7 be preferred (and at least tested)? At least this feature:

https://www.python.org/downloads/release/python-370/ 
"Avoiding the use of ASCII as a default text encoding (PEP 538, legacy C locale coercion and PEP 540, forced UTF-8 runtime mode)"

I'm not sure, but I think this "forced UTF-8 runtime mode" may align better with Julia. At least if Python 3.7 is used, maybe it should be enabled by default?

It may or may not (already) be backported to earlier versions:
https://pythoninsider.blogspot.com/2018/08/python-356-and-python-349-are-now.html

--
I'm a little skeptical of leaving macOS at Python2 (in some cases it may not matter, then could have stayed with Python2 on Linux, but in other cases code will cork on one but not the other platform).

Should at least Windows also default to Python3? It usually has no Python installed, so if conda gets one for you it may as well be Python 3.7.

stevengj:
MacOS and Windows default to using Conda, which now defaults to Python 3.

stevengj:
It would be reasonable to enable UTF-8 I/O in Python, but not by an environment variable (which would affect child processes too).  We need to find an API for this.

Issue #507, opened 2018-06-21
Misidentifying subclass of (python) list?


rjplevin:
I'm trying to wrap a package that defines a class that subtypes python's `list` and adds methods. When I create one of these, it appears to be misidentified as a `list`, so the additional methods are not available. 

The offending line seems to be conversions.jl:762:
```
        # only handle PyList for now
        return pyisinstance(o, @pyglobalobj :PyList_Type) ? Array : Union{}
```
Would it break things to check if `type(o) == list` rather than whether its an instance?

Example:
```
Main> py"class MyClass1(list):    # subclass of list
            def fn(self):
                print 'waldo was here'
       "

Main> py"class MyClass2():     # NOT a subclass of list
            def fn(self):
                print 'waldo was here'
       "

Main> c1 = py"MyClass1()"     # misidentified as a list => Array
0-element Array{Any,1}

Main> c2 = py"MyClass2()"
PyObject <__main__.MyClass2 instance at 0x1a48492758>

Main> py"$c1.fn()"
ERROR: PyError (ccall(@pysym(:PyEval_EvalCode), PyPtr, (PyPtr, PyPtr, PyPtr), o, globals, locals)) <
type 'exceptions.AttributeError'>
AttributeError("'list' object has no attribute 'fn'",)
  File "/Users/rjp/.julia/v0.6/PyCall/src/pyeval.jl", line 1, in <module>
    const Py_single_input = 256  # from Python.h

Stacktrace:
 [1] pyerr_check at /Users/rjp/.julia/v0.6/PyCall/src/exception.jl:60 [inlined]
 [2] pyerr_check at /Users/rjp/.julia/v0.6/PyCall/src/exception.jl:64 [inlined]
 [3] macro expansion at /Users/rjp/.julia/v0.6/PyCall/src/exception.jl:84 [inlined]
 [4] pyeval_(::String, ::PyCall.PyDict{String,PyCall.PyObject,false}, ::PyCall.PyDict{String,PyCall.
PyObject,false}, ::Int64, ::String) at /Users/rjp/.julia/v0.6/PyCall/src/pyeval.jl:23
 [5] eval(::Module, ::Any) at ./boot.jl:235

Main> py"$c2.fn()"
waldo was here
```

oxinabox:
I think this would be a good change.
It is hitting me for Dicts right now. And more generally it has been bugging me for years.
The work around is to use `pycall(func, PyObject, args...)` which avoids converting the type.

[...]

PR #517, opened 2018-08-12
WIP: override get/setproperty and propertynames

crstnbr:
The intent of this PR is to use 0.7/1.0's getproperty functionality to make `o.foo` behave as in Python, i.e. to map it to the current `o[:foo]`.

### DIFF
--- a/benchmarks/pywrapfn.jl
+++ b/benchmarks/pywrapfn.jl
@@ -9,10 +9,7 @@ end
 
 function PyWrapFn(o::Union{PyObject, PyPtr}, nargs::Int, returntype::Type=PyObject)
     pyargsptr = ccall((@pysym :PyTuple_New), PyPtr, (Int,), nargs)
-    ret = PyNULL()
-    optr = o isa PyPtr ? o : o.o
-    pyincref_(optr)
-    return PyWrapFn{nargs, returntype}(optr, pyargsptr, ret)
+    return PyWrapFn{nargs, returntype}(pyincref_(PyPtr(o)), pyargsptr, PyNULL())
 end
 
 (pf::PyWrapFn{N, RT})(args...) where {N, RT} =
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -882,14 +882,14 @@ for (mime, method) in ((MIME"text/html", "_repr_html_"),
         function show(io::IO, mime::$mime, o::PyObject)
             if !ispynull(o) && hasproperty(o, $method)
                 r = pycall(o[$method], PyObject)
-                PyPtr(r) != pynothing[] && return write(io, convert($T, r))
+                !(r ≛ pynothing[]) && return write(io, convert($T, r))
             end
             throw(MethodError(show, (io, mime, o)))
         end
         Base.$showable(::$mime, o::PyObject) =
             !ispynull(o) && hasproperty(o, $method) && let meth = o[$method]
-                PyPtr(meth) != pynothing[] &&
-                PyPtr(pycall(meth, PyObject)) != pynothing[]
+                !(meth ≛ pynothing[]) &&
+                !(pycall(meth, PyObject) ≛ pynothing[])
             end
     end
 end
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -331,20 +331,9 @@ function getindex(o::PyObject, s::T) where T<:Union{Symbol, AbstractString}
     return getproperty(o, s)
 end
 
-
-# If I make this
-# setindex!(o::PyObject, v, s::Union{Symbol,AbstractString}) = setproperty!(o, s, v)
-# the tests (Pkg.test("PyCall")) won't go through. Don't know why.....
-function setindex!(o::PyObject, v, s::Union{Symbol,AbstractString})
-    if ispynull(o)
-        throw(ArgumentError("assign of NULL PyObject"))
-    end
-    if -1 == ccall((@pysym :PyObject_SetAttrString), Cint,
-                   (PyPtr, Cstring, PyPtr), o, s, PyObject(v))
-        pyerr_clear()
-        throw(KeyError(s))
-    end
-    o
+function setindex!(o::PyObject, v, s::Union{Symbol, AbstractString})
+    Base.depwarn("`setindex!(o::PyObject, v, s::Union{Symbol, AbstractString})` is deprecated in favor of `setproperty!(o, s, v)`.", :setindex!)
+    return _setproperty!(o, s, v)
 end
 
 function haskey(o::PyObject, s::Union{Symbol,AbstractString})
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -322,8 +322,12 @@ function setproperty!(o::PyObject, s::Union{Symbol,AbstractString}, v)
     o
 end
 
-function getindex(o::PyObject, s::Union{Symbol, AbstractString})
-    Base.depwarn("`getindex(o::PyObject, s::Union{Symbol, AbstractString})` is deprecated in favor of dot overloading (`getproperty`) so elements should now be accessed as e.g. `o.s` instead of `o[:s]`.", :getindex)
+function getindex(o::PyObject, s::T) where T<:Union{Symbol, AbstractString}
+    if T == Symbol
+        Base.depwarn("`getindex(o::PyObject, s::Symbol)` is deprecated in favor of dot overloading (`getproperty`) so elements should now be accessed as e.g. `o.s` instead of `o[:s]`.", :getindex)
+    else
+        Base.depwarn("`getindex(o::PyObject, s::AbstractString)` is deprecated in favor of dot overloading (`getproperty`) so elements should now be accessed as e.g. `o.\"s\"` instead of `o[\"s\"]`.", :getindex)
+    end
     return getproperty(o, s)
 end
 
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -86,10 +86,7 @@ PyPtr(o::PyObject) = getfield(o, :o)
 
 `PyPtr` based comparison of `x` and `y`, which can be of type `PyObject` or `PyPtr`. 
 """
-≛(o::PyObject, p::PyPtr) = PyPtr(o) == p
-≛(p::PyPtr, o::PyObject) = o ≛ p
-≛(o1::PyObject, o2::PyObject) = PyPtr(o1) == PyPtr(o2)
-≛(p1::PyPtr, p2::PyPtr) = p1 == p2
+≛(o1::Union{PyObject,PyPtr}, o2::Union{PyObject,PyPtr}) = PyPtr(o1) == PyPtr(o2)
 
 """
     PyNULL()
@@ -260,8 +257,8 @@ end
 
 function Base.Docs.doc(o::PyObject)
     if hasproperty(o, "__doc__")
-        d = o["__doc__"]
-        if !(PyPtr(d) ≛ pynothing[])
+        d = o."__doc__"
+        if !(d ≛ pynothing[])
             return Base.Docs.Text(convert(AbstractString, d))
         end
     end
@@ -311,7 +308,7 @@ end
 getproperty(o::PyObject, s::Symbol) = convert(PyAny, getproperty(o, string(s)))
 
 propertynames(o::PyObject) = map(x->Symbol(first(x)), 
-                                pycall(inspect["getmembers"], PyObject, o))
+                                pycall(inspect."getmembers", PyObject, o))
 
 function setproperty!(o::PyObject, s::Union{Symbol,AbstractString}, v)
     if ispynull(o)
@@ -325,7 +322,10 @@ function setproperty!(o::PyObject, s::Union{Symbol,AbstractString}, v)
     o
 end
 
-getindex(o::PyObject, s::Union{Symbol, AbstractString}) = getproperty(o, s)
+function getindex(o::PyObject, s::Union{Symbol, AbstractString})
+    Base.depwarn("`getindex(o::PyObject, s::Union{Symbol, AbstractString})` is deprecated in favor of dot overloading (`getproperty`) so elements should now be accessed as e.g. `o.s` instead of `o[:s]`.", :getindex)
+    return getproperty(o, s)
+end
 
 
 # If I make this
@@ -343,7 +343,10 @@ function setindex!(o::PyObject, v, s::Union{Symbol,AbstractString})
     o
 end
 
-haskey(o::PyObject, s::Union{Symbol,AbstractString}) = hasproperty(o, s)
+function haskey(o::PyObject, s::Union{Symbol,AbstractString})
+    Base.depwarn("`haskey(o::PyObject, s::Union{Symbol, AbstractString})` is deprecated, use `hasproperty(o, s)` instead.", :haskey)
+    return hasproperty(o, s)
+end
 
 function hasproperty(o::PyObject, s::Union{Symbol,AbstractString})
     if ispynull(o)
@@ -376,12 +379,12 @@ If the Python module contains identifiers that are reserved words in Julia (e.g.
 """
 function pywrap(o::PyObject, mname::Symbol=:__anon__)
     members = convert(Vector{Tuple{AbstractString,PyObject}},
-                      pycall(inspect["getmembers"], PyObject, o))
+                      pycall(inspect."getmembers", PyObject, o))
     filter!(m -> !(m[1] in reserved), members)
     m = Module(mname, false)
     consts = [Expr(:const, Expr(:(=), Symbol(x[1]), convert(PyAny, x[2]))) for x in members]
     exports = try
-                  convert(Vector{Symbol}, o["__all__"])
+                  convert(Vector{Symbol}, o."__all__")
               catch
                   [Symbol(x[1]) for x in filter(x -> x[1][1] != '_', members)]
               end
@@ -625,8 +628,8 @@ function _pywith(EXPR,VAR,TYPE,BLOCK)
             end
         end
         mgrT = pytypeof(mgr)
-        exit = mgrT["__exit__"]
-        value = @pycall mgrT["__enter__"](mgr)::$(esc(TYPE))
+        exit = mgrT."__exit__"
+        value = @pycall mgrT."__enter__"(mgr)::$(esc(TYPE))
         exc = true
         try
             try
@@ -634,7 +637,7 @@ function _pywith(EXPR,VAR,TYPE,BLOCK)
                 $(esc(BLOCK))
             catch err
                 exc = false
-                if !(@pycall exit(mgr, pyimport(:sys)[:exc_info]()...)::Bool)
+                if !(@pycall exit(mgr, pyimport(:sys).exc_info()...)::Bool)
                     throw(err)
                 end
             end
@@ -835,7 +838,7 @@ end
 pushfirst!(a::PyObject, item) = insert!(a, 1, item)
 
 function prepend!(a::PyObject, items)
-    if isa(items,PyObject) && PyPtr(items) == PyPtr(a)
+    if isa(items,PyObject) && items ≛ a
         # avoid infinite loop in prepending a to itself
         return prepend!(a, collect(items))
     end
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -18,6 +18,11 @@ import Base: size, ndims, similar, copy, getindex, setindex!, stride,
        eltype, get, delete!, empty!, length, isempty,
        filter!, hash, splice!, pop!, ==, isequal, push!,
        append!, insert!, prepend!, unsafe_convert
+
+@static if VERSION >= v"0.7-"
+    import Base: getproperty, setproperty!, propertynames
+end
+
 import Compat: pushfirst!, popfirst!, firstindex, lastindex
 
 # Python C API is not interrupt-safe.  In principle, we should
@@ -291,7 +296,7 @@ end
 # with the former returning an raw PyObject and the latter giving the PyAny
 # conversion.
 
-function Base.getproperty(o::PyObject, s::AbstractString)
+function getproperty(o::PyObject, s::AbstractString)
     if ispynull(o)
         throw(ArgumentError("ref of NULL PyObject"))
     end
@@ -303,12 +308,12 @@ function Base.getproperty(o::PyObject, s::AbstractString)
     return PyObject(p)
 end
 
-Base.getproperty(o::PyObject, s::Symbol) = convert(PyAny, getproperty(o, string(s)))
+getproperty(o::PyObject, s::Symbol) = convert(PyAny, getproperty(o, string(s)))
 
-Base.propertynames(o::PyObject) = map(x->Symbol(first(x)), 
+propertynames(o::PyObject) = map(x->Symbol(first(x)), 
                                 pycall(inspect["getmembers"], PyObject, o))
 
-function Base.setproperty!(o::PyObject, s::Union{Symbol,AbstractString}, v)
+function setproperty!(o::PyObject, s::Union{Symbol,AbstractString}, v)
     if ispynull(o)
         throw(ArgumentError("assign of NULL PyObject"))
     end
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -74,6 +74,18 @@ mutable struct PyObject
     end
 end
 
+PyPtr(o::PyObject) = getfield(o, :o)
+
+"""
+    ≛(x, y)
+
+`PyPtr` based comparison of `x` and `y`, which can be of type `PyObject` or `PyPtr`. 
+"""
+≛(o::PyObject, p::PyPtr) = PyPtr(o) == p
+≛(p::PyPtr, o::PyObject) = o ≛ p
+≛(o1::PyObject, o2::PyObject) = PyPtr(o1) == PyPtr(o2)
+≛(p1::PyPtr, p2::PyPtr) = p1 == p2
+
 """
     PyNULL()
 
@@ -96,7 +108,7 @@ PyNULL() = PyObject(PyPtr_NULL)
 Test where `o` contains a `NULL` pointer to a Python object, i.e. whether
 it is equivalent to a `PyNULL()` object.
 """
-ispynull(o::PyObject) = o.o == PyPtr_NULL
+ispynull(o::PyObject) = o ≛ PyPtr_NULL
 
 function pydecref_(o::Union{PyPtr,PyObject})
     ccall(@pysym(:Py_DecRef), Cvoid, (PyPtr,), o)
@@ -105,7 +117,7 @@ end
 
 function pydecref(o::PyObject)
     pydecref_(o)
-    o.o = PyPtr_NULL
+    setfield!(o, :o, PyPtr_NULL)
     return o
 end
 
@@ -128,8 +140,8 @@ will be performed when `o` is garbage collected.  (This means that
 you can no longer use `o`.)  Used for passing objects to Python.
 """
 function pystealref!(o::PyObject)
-    optr = o.o
-    o.o = PyPtr_NULL # don't decref when o is gc'ed
+    optr = PyPtr(o)
+    setfield!(o, :o, PyPtr_NULL) # don't decref when o is gc'ed
     return optr
 end
 
@@ -141,16 +153,16 @@ a `PyObject`, the refcount is incremented.  Otherwise a `PyObject`
 wrapping/converted from `x` is created.
 """
 pyreturn(x::Any) = pystealref!(PyObject(x))
-pyreturn(x::PyObject) = pyincref_(x.o)
+pyreturn(x::PyObject) = pyincref_(PyPtr(x))
 
 function Base.copy!(dest::PyObject, src::PyObject)
     pydecref(dest)
-    dest.o = src.o
+    setfield!(dest, :o, PyPtr(src))
     return pyincref(dest)
 end
 
 pyisinstance(o::PyObject, t::PyObject) =
-  !ispynull(t) && ccall((@pysym :PyObject_IsInstance), Cint, (PyPtr,PyPtr), o, t.o) == 1
+  !ispynull(t) && ccall((@pysym :PyObject_IsInstance), Cint, (PyPtr,PyPtr), o, t) == 1
 
 pyisinstance(o::PyObject, t::Union{Ptr{Cvoid},PyPtr}) =
   t != C_NULL && ccall((@pysym :PyObject_IsInstance), Cint, (PyPtr,PyPtr), o, t) == 1
@@ -159,7 +171,7 @@ pyquery(q::Ptr{Cvoid}, o::PyObject) =
   ccall(q, Cint, (PyPtr,), o) == 1
 
 # conversion to pass PyObject as ccall arguments:
-unsafe_convert(::Type{PyPtr}, po::PyObject) = po.o
+unsafe_convert(::Type{PyPtr}, po::PyObject) = PyPtr(po)
 
 # use constructor for generic conversions to PyObject
 convert(::Type{PyObject}, o) = PyObject(o)
@@ -230,7 +242,7 @@ function pystring(o::PyObject)
             s = ccall((@pysym :PyObject_Str), PyPtr, (PyPtr,), o)
             if (s == C_NULL)
                 pyerr_clear()
-                return string(o.o)
+                return string(PyPtr(o))
             end
         end
         return convert(AbstractString, PyObject(s))
@@ -242,9 +254,9 @@ function show(io::IO, o::PyObject)
 end
 
 function Base.Docs.doc(o::PyObject)
-    if haskey(o, "__doc__")
+    if hasproperty(o, "__doc__")
         d = o["__doc__"]
-        if d.o != pynothing[]
+        if !(PyPtr(d) ≛ pynothing[])
             return Base.Docs.Text(convert(AbstractString, d))
         end
     end
@@ -263,23 +275,23 @@ function hash(o::PyObject)
     elseif is_pyjlwrap(o)
         # call native Julia hash directly on wrapped Julia objects,
         # since on 64-bit Windows the Python 2.x hash is only 32 bits
-        hashsalt(unsafe_pyjlwrap_to_objref(o.o))
+        hashsalt(unsafe_pyjlwrap_to_objref(PyPtr(o)))
     else
         h = ccall((@pysym :PyObject_Hash), Py_hash_t, (PyPtr,), o)
         if h == -1 # error
             pyerr_clear()
-            return hashsalt(o.o)
+            return hashsalt(PyPtr(o))
         end
         hashsalt(h)
     end
 end
 
 #########################################################################
-# For o::PyObject, make o["foo"] and o[:foo] equivalent to o.foo in Python,
+# For o::PyObject, make o["foo"], o[:foo], and o.foo equivalent to o.foo in Python,
 # with the former returning an raw PyObject and the latter giving the PyAny
 # conversion.
 
-function getindex(o::PyObject, s::AbstractString)
+function Base.getproperty(o::PyObject, s::AbstractString)
     if ispynull(o)
         throw(ArgumentError("ref of NULL PyObject"))
     end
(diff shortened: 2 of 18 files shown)


crstnbr:
Minimal example:
```julia
@pyimport numpy as np
np.random.rand() #works now
```

Autocompletion works as well, i.e. `np.random.<tab>` gives a list of all available members of the python object.

stevengj:
Thanks.   I was planning on waiting a little while to do this, so that can:

* Drop 0.6 support.
* Deprecate all the old `getindex` methods.  (Eventually they will be made equivalent to `o[i]` in Python once the deprecations are removed.)
* Deprecate `@pyimport` and `pywrap` since you can now just do `np = pyimport("numpy")`

src/PyCall.jl
```diff
@@ -17,7 +17,7 @@ import Base: size, ndims, similar, copy, getindex, setindex!, stride,
        convert, pointer, summary, convert, show, haskey, keys, values,
        eltype, get, delete!, empty!, length, isempty, start, done,
        next, filter!, hash, splice!, pop!, ==, isequal, push!,
-       append!, insert!, prepend!, unsafe_convert
+       append!, insert!, prepend!, unsafe_convert, getproperty, propertynames
```
stevengj: I'd rather move towards the style of explicitly defining `Base.getproperty` etcetera, rather than importing.

src/PyCall.jl
```diff
@@ -286,6 +286,20 @@ end
 
 getindex(o::PyObject, s::Symbol) = convert(PyAny, getindex(o, string(s)))
 
+function getproperty(o::PyObject, s::Symbol)
+    if s == :o # we still want to be able to access o.o
```
stevengj: `===`

I think we want to deprecate `o.o`, however, in favor of a `PyPtr(o)` method (and an `unsafe_convert(PyPtr, o)` method for `ccall`).   Otherwise you won't be able to access fields named `o` in Python objects.

src/PyCall.jl
```diff
@@ -286,6 +286,20 @@ end
...
+        getfield(o, s)
+    else
+        getindex(o, s)
+    end
+end
+
+function propertynames(o::PyObject)
+    members = convert(Vector{Tuple{AbstractString,PyObject}},
```
stevengj: I don't think we need the `convert` call here since the `getmembers` result is iterable.  We can just do `map` on the result.

src/PyCall.jl
```diff
@@ -286,6 +286,20 @@ end
 
 getindex(o::PyObject, s::Symbol) = convert(PyAny, getindex(o, string(s)))
 
+function getproperty(o::PyObject, s::Symbol)
```
stevengj: Note that I think we also want to accept an `AbstractString` here so that you can do `o."foo"` and get the raw `PyObject`, similar to the current `o["foo"]`.

src/PyCall.jl
```diff
@@ -298,6 +312,14 @@ function setindex!(o::PyObject, v, s::Union{Symbol,AbstractString})
     o
 end
 
+function setproperty!(o::PyObject, s::Symbol, v)
+    if s == :o
+        setfield(o, s, v)
```
stevengj: This should be depecated in favor of explicit `setfield` calls.   Most PyCall users should never modify the internal `.o` field.

benchmarks/pywrapfn.jl
```diff
@@ -10,7 +10,7 @@ end
 function PyWrapFn(o::Union{PyObject, PyPtr}, nargs::Int, returntype::Type=PyObject)
     pyargsptr = ccall((@pysym :PyTuple_New), PyPtr, (Int,), nargs)
     ret = PyNULL()
-    optr = o isa PyPtr ? o : o.o
+    optr = o isa PyPtr ? o : PyPtr(o)
```
stevengj: You can just do `optr = PyPtr(o)`, since this is the identity if `o isa PyPtr`.

stevengj: Or condense it to `return PyWrapFn{nargs, returntype}(pyincref_(PyPtr(o)), pyargsptr, PyNULL())`

stevengj: However, it looks like this benchmark will no longer work after 8d300021792b8146483e0df42afd11ccd665fd99, since I elminated that `_pycall!` method.

src/PyCall.jl
```diff
@@ -17,7 +17,7 @@ import Base: size, ndims, similar, copy, getindex, setindex!, stride,
        convert, pointer, summary, convert, show, haskey, keys, values,
        eltype, get, delete!, empty!, length, isempty,
        filter!, hash, splice!, pop!, ==, isequal, push!,
-       append!, insert!, prepend!, unsafe_convert
+       append!, insert!, prepend!, unsafe_convert, getproperty, propertynames
```
stevengj: Rather than importing, I'd prefer to move towards the style of overriding `Base.getproperty` etcetera, which is nowadays usually the preferred style in Julia.

src/PyCall.jl
```diff
@@ -115,7 +117,7 @@ function pyincref_(o::PyPtr)
 end
 
 function pyincref(o::PyObject)
-    pyincref_(o.o)
+    pyincref_(PyPtr(o))
```
stevengj: 849adab01aae00484615060ae66b07889fbbb858 will change these slightly…

src/PyCall.jl
```diff
@@ -131,19 +133,19 @@ will be performed when `o` is garbage collected.  (This means that
...
 function Base.copy!(dest::PyObject, src::PyObject)
     pydecref(dest)
-    dest.o = src.o
+    setfield!(dest, :o, PyPtr(src))
     return pyincref(dest)
 end
 
 pyisinstance(o::PyObject, t::PyObject) =
-  !ispynull(t) && ccall((@pysym :PyObject_IsInstance), Cint, (PyPtr,PyPtr), o, t.o) == 1
+  !ispynull(t) && ccall((@pysym :PyObject_IsInstance), Cint, (PyPtr,PyPtr), o, PyPtr(t)) == 1
```
stevengj: This should just pass `t`

src/PyCall.jl
```diff
@@ -298,6 +300,14 @@ function setindex!(o::PyObject, v, s::Union{Symbol,AbstractString})
     o
 end
 
+getproperty(o::PyObject, s::Symbol) = getindex(o, s)
+getproperty(o::PyObject, s::AbstractString) = getindex(o, s)
```
stevengj: Better to move the definition of `getindex` here, since `getindex` will be deprecated.

src/PyCall.jl
```diff
@@ -298,6 +300,14 @@ function setindex!(o::PyObject, v, s::Union{Symbol,AbstractString})
...
+
+propertynames(o::PyObject) = map(x->Symbol(first(x)), 
+                                pycall(inspect["getmembers"], PyObject, o))
+
+setproperty!(o::PyObject, s::Union{Symbol,AbstractString}, v) = setindex!(o, v, s)
```
stevengj: similarly, move the definition of `setindex!` here.

src/PyCall.jl
```diff
@@ -298,6 +300,14 @@ function setindex!(o::PyObject, v, s::Union{Symbol,AbstractString})
...
+
 function haskey(o::PyObject, s::Union{Symbol,AbstractString})
```
stevengj: We'll need to change this to `hasproperty` or something?  Not sure if there is a Julia `Base` function to override here.

stevengj: (https://github.com/JuliaLang/julia/issues/28850)

src/PyCall.jl
```diff
@@ -429,7 +439,7 @@ function pyimport(name::AbstractString)
     if ispynull(o)
         if pyerr_occurred()
             e = PyError("PyImport_ImportModule")
-            if e.T.o == @pyglobalobjptr(:PyExc_ImportError)
+            if PyPtr(e.T) == @pyglobalobjptr(:PyExc_ImportError)
```
stevengj: I feel like we should just define an `==` method for `PyObject == PyPtr`.

stevengj: Or maybe call it something else, since this isn't really equality.  Maybe `≛` (`\starequal`)

src/conversions.jl
```diff
@@ -844,7 +844,7 @@ function convert(::Type{PyAny}, o::PyObject)
     try
         T = pytype_query(o)
         if T == PyObject && is_pyjlwrap(o)
-            return unsafe_pyjlwrap_to_objref(o.o)
+            return unsafe_pyjlwrap_to_objref(PyPtr(o))
```
stevengj: `unsafe_pyjlwrap_to_objref` should probably just have a method that takes a `PyObject`

src/exception.jl
```diff
@@ -133,5 +133,8 @@ end
 function pyraise(e::PyError)
     ccall((@pysym :PyErr_Restore), Cvoid, (PyPtr, PyPtr, PyPtr),
           e.T, e.val, e.traceback)
-    e.T.o = e.val.o = e.traceback.o = C_NULL # refs were stolen
+    # refs were stolen
+    setfield!(e.T, :o, C_NULL)
+    setfield!(e.val, :o, C_NULL)
+    setfield!(e.traceback, :o, C_NULL)
```
stevengj: Upon reflection, I think it would be better to pass `pyincref(e.T), pyincref(e.val), pyincref(e.traceback)` to the `ccall` rather than setting the objects to null after the call

src/PyCall.jl
crstnbr: I'll rename it to `hasproperty` everyhere. We can override the `Base` function once it's there.

src/exception.jl
crstnbr: Done.

src/conversions.jl
crstnbr: Done.

src/PyCall.jl
crstnbr: Done.

src/PyCall.jl
crstnbr: Done.

src/PyCall.jl
crstnbr: Done.

src/PyCall.jl
crstnbr: Done.

src/PyCall.jl
crstnbr: Done. (But only substituted it in the comparison above so far)

src/PyCall.jl
crstnbr: I see. But is there something I should do in this PR?

benchmarks/pywrapfn.jl
crstnbr: Done.
What should I do about your last comment in this PR?

crstnbr:
BTW, regarding your third point above, I really like the "@pyimport numpy as np" syntax because it's basically the same in python. Maybe we can keep it as syntactic sugar?

oxinabox:
`pywrap(::PyObject)::Module` can be deprecated with this.  
Since that was just a hack to pretend to have `.` overloading.

(I can't workout if that has been done, so many things in this package are called `py.*wrap.*`)

crstnbr:
I understand why the macro was necessary. But as I said, I would see a benefit in the second syntax because it's close to what you do in python (and this is a python interface). I won't push this point any further though.

tkf:
I think rebasing onto master to resolve the conflicts would help.

src/PyCall.jl
```diff
@@ -96,16 +101,16 @@ PyNULL() = PyObject(PyPtr_NULL)
 Test where `o` contains a `NULL` pointer to a Python object, i.e. whether
 it is equivalent to a `PyNULL()` object.
 """
-ispynull(o::PyObject) = o.o == PyPtr_NULL
+ispynull(o::PyObject) = PyPtr(o) == PyPtr_NULL
```
tkf: Why not use `≛` here? Also, it seems you can use `≛` in other places too?

src/PyCall.jl
```diff
@@ -235,9 +240,9 @@ function show(io::IO, o::PyObject)
 end
 
 function Base.Docs.doc(o::PyObject)
-    if haskey(o, "__doc__")
+    if hasproperty(o, "__doc__")
         d = o["__doc__"]
-        if d.o != pynothing[]
+        if PyPtr(d) != pynothing[]
```
tkf: Do we want `≛̸` (`\starequal\not`) to handle this?

(I wish `a !≛ b` automatically lowered to `!(≛)(a, b)`...)

stevengj: Probably clearer to just use `!(a ≛ b)`

crstnbr:
I rebased an hopefully fixed conflicts with master. Passes all tests locally.

src/PyCall.jl
```diff
@@ -263,23 +275,23 @@ function hash(o::PyObject)
...
 end
 
 #########################################################################
-# For o::PyObject, make o["foo"] and o[:foo] equivalent to o.foo in Python,
+# For o::PyObject, make o["foo"], o[:foo], and o.foo equivalent to o.foo in Python,
 # with the former returning an raw PyObject and the latter giving the PyAny
 # conversion.
 
-function getindex(o::PyObject, s::AbstractString)
+function Base.getproperty(o::PyObject, s::AbstractString)
```
tkf: This fails in 0.6 because `getproperty` is new in 0.7 https://travis-ci.org/JuliaPy/PyCall.jl/jobs/427201767#L668

Same for other `*property*` methods.

tkf:
Thanks for rebasing! But I should have noticed that we need to drop Julia 0.6 support first to merge this, before recommending the rebase... 😅

src/PyCall.jl
crstnbr: Yes.

crstnbr:
Sure, no worries. I can rebase again :)

crstnbr:
Can I help with dropping 0.6 support somewhere? I really want this feature here merged, I was waiting for it for a long time! :)

Also, I'm hosting a local Julia workshop in October and it would be nice to be able to say "you can just copy past python code" :)

Datseris:
totally agree, 0.6 is a thing of the past. whoever needs it can use a previous version of pycall. This is something awesome!

test/runtests.jl
```diff
@@ -258,6 +258,14 @@ const PyInt = pyversion < v"3" ? Int : Clonglong
     end
     @test convert(BigInt, PyObject(1234)) == 1234
 
+    # hasproperty, getproperty, and propertynames
+    @test PyCall.hasproperty(np, "random")
```
tkf: These tests are failing on Windows because numpy is not installed by default:
https://ci.appveyor.com/project/StevenGJohnson/pycall-jl-nu3aa/build/1.0.342/job/3t0m1me5bkffvn3t#L113

If you need to test it against Python modules, you should use something from Python standard library.  If you need modules with three levels to test access pattern like `A.B.C` (and that is Python 2/3 compatible), use, e.g., `logging.config.dictConfig`:
https://docs.python.org/2/library/logging.config.html

But if you just need to have nested Python object, it's probably the easiest to create nested class:

```julia
py"""
class A:
    class B:
        C = 1
"""
A = py"A"
@test A.B.C == 1
```

tkf:
FYI, here is what I found in the [traffic page](https://github.com/JuliaPy/PyCall.jl/graphs/traffic):

image


[Pkg3 bypasses git clone](https://discourse.julialang.org/t/any-way-to-know-the-download-statistics-of-packages/15027) so I guess it's a good indicator of Julia 0.6 users.  (I'm not against dropping Julia 0.6 per se; Just bringing up some data.)

tkf:
It is probably not difficult to support Julia 0.6 using the same source code at this point.  We also probably better use `Base.@deprecate`:

```julia
@static if VERSION <= v"0.7-"
    import Base: getproperty
end

function getproperty(o::PyObject, s::AbstractString)
    # ...
end

@static if VERSION <= v"0.7-"
    Base.@deprecate getindex(o::PyObject, s::AbstractString) getproperty(o, s)
else
    getindex(o::PyObject, s::AbstractString) = getproperty(o, s)
end
```

and same for `hasproperty` and `setproperty`.

Datseris:
Any update here?,.....

crstnbr:
Sorry, I'm on vacation. Will continue on this next week.

George Datseris <notifications@github.com> schrieb am Fr., 19. Okt. 2018,
15:07:

crstnbr:
> We also probably better use `Base.@deprecate`

Internally, `somepyobj["something"]` is used quite often. If I deprecated `getindex` I have to replace all those `getindex` calls with `getproperty` calls. Should I do it?

crstnbr:
Rebased and adjusted such that 0.6 should still be supported. Passes all tests.

tkf:
I think it's better to wait for @stevengj.  We need to decided (1) if we are to support Julia 0.6 and (2) whether or not we add the deprecation period to wait for downstream packages to switch to `obj.attr` and `obj."attr"`.

If we are going to still support 0.6, I suggest to use `getproperty(someobj, "something")` in this PR.  Most of them can be replaced by `sed` automatically.  Something like:

```sh
cd ~/.julia/dev/PyCall/src
sed -E -i 's/([A-Za-z0-9_\.])+\[("[^"]+")\]/getproperty(\1, \2)/g' *.jl
```

(after handling the assignment `grep -E '\[".*=' *.jl` first)

Of course, some of them might be dictionary lookups but supposedly the tests should find them if we make mistakes.  There are "just" ~ 60 of them so going over a diff is not super crazy.

We need to do the check anyway if we are to use `obj."attr"`. Also, replacing `getproperty`/`setproperty` to `obj."attr"` afterward will be much easier.

BTW, there is another PR #487 which may be get merged soon.  So I think you need to rebase again...  (I sympathize that you have to rebase multiple times.)

test/runtests.jl
```diff
@@ -258,6 +258,16 @@ const PyInt = pyversion < v"3" ? Int : Clonglong
     end
     @test convert(BigInt, PyObject(1234)) == 1234
 
+    @static if VERSION >= v"0.7-"
```
tkf: Maybe do this

```julia
using PyCall: hasproperty
@static if VERSION < v"0.7-"
    using PyCall: getproperty, setproperty, propertynames
end
```

in the beginning of the `runtests.jl` and run the tests even in Julia 0.6?

test/runtests.jl
```diff
@@ -258,6 +258,16 @@ const PyInt = pyversion < v"3" ? Int : Clonglong
...
+        # hasproperty, getproperty, and propertynames
+        @test PyCall.hasproperty(np, "random")
```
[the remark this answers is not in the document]
tkf: I realized that you need to open the link to the comment in new tab/window (in chrome).

test/runtests.jl
crstnbr: I thought about it but thought that these methods will be tested by CI and julia 0.7 anyway. Doesn't hurt to add it though.

crstnbr:
Tests pass on 0.6 and 1.0. I'll wait for @stevengj's decision before I continue/rebase.

test/runtests.jl
```diff
@@ -258,6 +263,23 @@ const PyInt = pyversion < v"3" ? Int : Clonglong
...
+        class B:
+            C = 1
+    """
+    A = py"A"
+    @test hasproperty(A, "B")
+    @test getproperty(A, "B") == py"A.B"
+    @test :B in propertynames(A)
+    @static if VERSION >= v"0.7-"
+        @test A.B.C == 1
+        @test_throws KeyError A.X
```
tkf: It's nice that it's tested, but now that we write it as property access, isn't it a bit funny to throw KeyError?  But Julia doesn't have "UndefPropertyError"...  getindex seems to throw ErrorException but that's not super nice.  Still, I guess the best choice here is ErrorException as it would correctly communicate to the users to not rely on the exception type.

tkf: Maybe `UndefRefError` is OK to use?:

```julia
help?> UndefRefError
search: UndefRefError

  UndefRefError()

  The item or field is not defined for the given object.
```

(Julia has clear distinction between field and property so it may be misleading tough...)

It can be invoked by

```julia
julia> struct UndefFields
           a
           UndefFields() = new()
       end

julia> UndefField()
UndefFields(#undef)

julia> err = try UndefFields().a; catch err; err; end
UndefRefError()
```

What I was referring earlier was this behavior:

```julia
julia> err = try UndefFields().b; catch err; err; end
ErrorException("type UndefFields has no field b")
```

src/PyCall.jl
```diff
@@ -18,6 +18,11 @@ import Base: size, ndims, similar, copy, getindex, setindex!, stride,
        eltype, get, delete!, empty!, length, isempty,
        filter!, hash, splice!, pop!, ==, isequal, push!,
        append!, insert!, prepend!, unsafe_convert
+
+@static if VERSION >= v"0.7-"
```
musm: `@static` is not needed in global scope

Datseris:
@crstnbr is this branch safe-to-use? To my eyes this seems like an **extremely** useful thing, so I can't really understand why it isn't merged... If it is safe to use I'd much rather use this branch than the stable version.

(I am asking because I see conflicting files).

crstnbr:
> @crstnbr is this branch safe-to-use?  
> (I am asking because I see conflicting files).

Yes, it should be. All tests pass. The conflicts are likely due to other PRs which have been merged into master in the meantime. I'd need to rebase (again). But I see no point in rebasing before @stevengj has decided on the questions above (let's kindly ping him).

Datseris:
@crstnbr Thank you so much for this. I am now on your branch (which works). This is an amazing improvement. For the first time ever, you can copy/paste python code in Julia. I can't stress how productive this is and how much faster you code using both Julia and PyCall. Most of my typing when using PyCall was done turning `[: ]` to `.` and vice versa. NO MORE!!!!!

(and it yet remains unmerged for some reason... maybe JuliaPy needs more admins?)

Datseris:
Found a bug (that comes not from the PR but from PyPlot).
```
A = rand(5, 5)
ax = imshow(A)
cs = colorbar()
cs.cmap.set_under()
```
this fails, because `cs.cmap` returns a Julia type instead of a Python type, for some weird reason. Instead one has to do `cs.cmap[:set_under]()` for it to work.

oxinabox:
@Datseris 

> this fails, because cs.cmap returns a Julia type instead of a Python type, for some weird reason. Instead one has to do cs.cmap[:set_under]() for it to work.

That is issue #507

Basically, we julia-ize all subtype of List (etc),
when we really should only be julia-izing the exact type.

crstnbr:
Can we get this done and merged in 2018?

PallHaraldsson:
"merged in 2018?" No. :) In 2019 hopefully?

Is this really "WIP"? I mean should that be dropped from the title? @crstnbr in October: "Tests pass on 0.6 and 1.0. I'll wait for @stevengj's decision before I continue/rebase."

I understand Steven is teaching Julia 1.0 in 2019 (0.6 in Fall 2018), and I want to push others to do so. It seems merging this would help him and others? Maybe he's just really busy; and the only one that can merge, so I second "maybe JuliaPy needs more admins?"

@tkf on deciding:

> (1) if we are to support Julia 0.6 and (2) whether or not we add the deprecation period to wait for downstream packages to switch to obj.attr and obj."attr".

Do we? I understand with this merged, we support the new syntax (and old PyCall the old mangled Python syntax). I'm not sure if this PR does support both old and new syntax already, but it could if that helps get it merged? It at least already supports 1.0+ and 0.6.

tkf:
> > (1) if we are to support Julia 0.6 and (2) whether or not we add the deprecation period to wait for downstream packages to switch to obj.attr and obj."attr".
> 
> Do we?

IIRC that comment was about the internal usage of `somepyobj["something"]`. If we don't put any deprecation warning, I think this PR can probably be merged as-is (after resolving the conflict).

> Maybe he's just really busy; and the only one that can merge, so I second "maybe JuliaPy needs more admins?"

I actually can hit the merge button in this repository.  I'm hesitant to do so as I still feel I'm still a newcomer to this project.

AzamatB:
Would be great to merge this if there are no outstanding issues – can't wait for this :)

StefanKarpinski:
Bumpity?

stevengj:
Sorry, I've been pretty busy; I'll try to get to this later this week.  In the meantime, would be good to fix the merge conflicts.

src/PyCall.jl
```diff
@@ -291,8 +308,29 @@ function getindex(o::PyObject, s::AbstractString)
...
+    end
+    if -1 == ccall((@pysym :PyObject_SetAttrString), Cint,
+                   (PyPtr, Cstring, PyPtr), o, s, PyObject(v))
+        pyerr_clear()
+        throw(KeyError(s))
+    end
+    o
+end
+
+getindex(o::PyObject, s::Union{Symbol, AbstractString}) = getproperty(o, s)
```
stevengj: This should have a deprecation warning.

src/PyCall.jl
```diff
@@ -305,9 +343,11 @@ function setindex!(o::PyObject, v, s::Union{Symbol,AbstractString})
     o
 end
 
-function haskey(o::PyObject, s::Union{Symbol,AbstractString})
+haskey(o::PyObject, s::Union{Symbol,AbstractString}) = hasproperty(o, s)
```
stevengj: Should have a deprecation.

src/PyCall.jl
```diff
@@ -242,9 +259,9 @@ function show(io::IO, o::PyObject)
 end
 
 function Base.Docs.doc(o::PyObject)
-    if haskey(o, "__doc__")
+    if hasproperty(o, "__doc__")
         d = o["__doc__"]
-        if d.o != pynothing[]
+        if !(PyPtr(d) ≛ pynothing[])
```
stevengj: Why do you need the `PyPtr` call here?  Just `d ≛ pynothing[]`, no?

src/PyCall.jl
```diff
@@ -795,7 +835,7 @@ end
 pushfirst!(a::PyObject, item) = insert!(a, 1, item)
 
 function prepend!(a::PyObject, items)
-    if isa(items,PyObject) && items.o == a.o
+    if isa(items,PyObject) && PyPtr(items) == PyPtr(a)
```
stevengj: Use `≛`?

src/PyCall.jl
```diff
@@ -837,16 +877,16 @@ for (mime, method) in ((MIME"text/html", "_repr_html_"),
     showable = VERSION < v"0.7.0-DEV.4047" ? :mimewritable : :showable
     @eval begin
         function show(io::IO, mime::$mime, o::PyObject)
-            if !ispynull(o) && haskey(o, $method)
+            if !ispynull(o) && hasproperty(o, $method)
                 r = pycall(o[$method], PyObject)
-                r.o != pynothing[] && return write(io, convert($T, r))
+                PyPtr(r) != pynothing[] && return write(io, convert($T, r))
```
stevengj: Use `≛`?

stevengj: Similarly for other `pynothing` comparisons.

src/exception.jl
```diff
@@ -172,6 +172,5 @@ end
 
 function pyraise(e::PyError)
     ccall((@pysym :PyErr_Restore), Cvoid, (PyPtr, PyPtr, PyPtr),
-          e.T, e.val, e.traceback)
-    e.T.o = e.val.o = e.traceback.o = C_NULL # refs were stolen
-end
+          pyincref(e.T), pyincref(e.val), pyincref(e.traceback))
```
stevengj: Why not leave the code as-is, changed to `setfield`?

src/io.jl
```diff
@@ -65,7 +65,7 @@ end
 
 ##########################################################################
 
-pyio_jl(self::PyObject) = unsafe_pyjlwrap_to_objref(self["io"].o)::IO
+pyio_jl(self::PyObject) = unsafe_pyjlwrap_to_objref(PyPtr(self["io"]))::IO
```
stevengj: Maybe just add a `PyObject` method to `unsafe_pyjlwrap_to_objref`?

src/pyfncall.jl
```diff
@@ -43,8 +43,8 @@ function __pycall!(ret::PyObject, pyargsptr::PyPtr, o::Union{PyObject,PyPtr},
     try
         retptr = @pycheckn ccall((@pysym :PyObject_Call), PyPtr, (PyPtr,PyPtr,PyPtr), o,
                         pyargsptr, kw)
-        pydecref_(ret.o)
-        ret.o = retptr
+        pydecref_(PyPtr(ret))
```
stevengj: Can just call `pydecref_(ret)`, no?

stevengj:
Some of these `PyPtr(o)` calls may need `GC.@preserve o` to be safe nowadays, but that can be done in another PR since it isn't a regression.

src/PyCall.jl
```diff
@@ -291,8 +308,29 @@ function getindex(o::PyObject, s::AbstractString)
...
+    o
+end
+
+getindex(o::PyObject, s::Union{Symbol, AbstractString}) = getproperty(o, s)
+
+
+# If I make this
+# setindex!(o::PyObject, v, s::Union{Symbol,AbstractString}) = setproperty!(o, s, v)
+# the tests (Pkg.test("PyCall")) won't go through. Don't know why.....
 function setindex!(o::PyObject, v, s::Union{Symbol,AbstractString})
```
stevengj: should be deprecated

src/PyCall.jl
```diff
@@ -74,6 +79,18 @@ mutable struct PyObject
...
+
+"""
+    ≛(x, y)
+
+`PyPtr` based comparison of `x` and `y`, which can be of type `PyObject` or `PyPtr`. 
+"""
+≛(o::PyObject, p::PyPtr) = PyPtr(o) == p
+≛(p::PyPtr, o::PyObject) = o ≛ p
+≛(o1::PyObject, o2::PyObject) = PyPtr(o1) == PyPtr(o2)
+≛(p1::PyPtr, p2::PyPtr) = p1 == p2
```
stevengj: You could just have one method:
```jl
≛(o1::Union{PyObject,PyPtr}, o2::Union{PyObject,PyPtr}) = PyPtr(o1) == PyPtr(o2)
```
since `PyPtr(p::PyPtr) = p` should be defined automatically.

Datseris:
Hi I just want to add that PyPlot's `plot3D` does not work on this branch:
```
plot3D(x,y,z);

ArgumentError: ref of NULL PyObject

Stacktrace:
 [1] getproperty(::PyCall.PyObject, ::String) at C:\Users\datseris\.julia\packages\PyCall\wIrCn\src\PyCall.jl:301
 [2] getproperty at C:\Users\datseris\.julia\packages\PyCall\wIrCn\src\PyCall.jl:311 [inlined]
 [3] PyCall.PyObject(::PyPlot.LazyPyModule) at C:\Users\datseris\.julia\packages\PyPlot\fZuOQ\src\plot3d.jl:9
 [4] using3D at C:\Users\datseris\.julia\packages\PyPlot\fZuOQ\src\plot3d.jl:34 [inlined]
 [5] #plot3D#169(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Array{Float64,1}, ::Vararg{Array{Float64,1},N} where N) at C:\Users\datseris\.julia\packages\PyPlot\fZuOQ\src\plot3d.jl:48
 [6] plot3D(::Array{Float64,1}, ::Vararg{Array{Float64,1},N} where N) at C:\Users\datseris\.julia\packages\PyPlot\fZuOQ\src\plot3d.jl:48
 [7] top-level scope at In[8]:4
```

src/PyCall.jl
```diff
@@ -291,8 +308,29 @@ function getindex(o::PyObject, s::AbstractString)
     return PyObject(p)
 end
 
-getindex(o::PyObject, s::Symbol) = convert(PyAny, getindex(o, string(s)))
+getproperty(o::PyObject, s::Symbol) = convert(PyAny, getproperty(o, string(s)))
```
stevengj: If `s === :o`, it should call `PyPtr(o)` with a deprecation warning.

src/PyCall.jl
```diff
@@ -291,8 +308,29 @@ function getindex(o::PyObject, s::AbstractString)
...
+
+propertynames(o::PyObject) = map(x->Symbol(first(x)), 
+                                pycall(inspect["getmembers"], PyObject, o))
 
+function setproperty!(o::PyObject, s::Union{Symbol,AbstractString}, v)
+    if ispynull(o)
```
stevengj: If `s === :o`, it should call `setfield(o, :o, v)` with a deprecation warning.

stevengj:
@Datseris, that should be fixed by https://github.com/JuliaPy/PyPlot.jl/commit/6439f2b57ef8940741fd7992b877ad8e8f817a10

src/PyCall.jl
stevengj: Actually, it's probably fine to leave out this deprecation.  I did a quick audit of all of the registered packages using PyCall, and it looks like only PyPlot was accessing the internal `o` field of `PyObject`, and PyPlot has been fixed.

src/PyCall.jl
stevengj: Actually, it's probably fine to leave out this deprecation, see above.

src/io.jl
crstnbr: In fact, it already exists.

Datseris:
Could this be merged soon? I currently get a million of warnings in my Julia for PyPlot: 
```
┌ Warning: `getindex(o::PyObject, s::Union{Symbol, AbstractString})` is deprecated in favor of dot overloading (`getproperty`) so elements should now be accessed as e.g. `o.s` instead of `o[:s]`.
│   caller = init_colormaps() at colormaps.jl:45
└ @ PyPlot C:\Users\datseris\.julia\packages\PyPlot\mQXSC\src\colormaps.jl:45
┌ Warning: `getindex(o::PyObject, s::Union{Symbol, AbstractString})` is deprecated in favor of dot overloading (`getproperty`) so elements should now be accessed as e.g. `o.s` instead of `o[:s]`.
│   caller = init_colormaps() at colormaps.jl:47
└ @ PyPlot C:\Users\datseris\.julia\packages\PyPlot\mQXSC\src\colormaps.jl:47
┌ Warning: `getindex(o::PyObject, s::Union{Symbol, AbstractString})` is deprecated in favor of dot overloading (`getproperty`) so elements should now be accessed as e.g. `o.s` instead of `o[:s]`.
│   caller = init_colormaps() at colormaps.jl:49
└ @ PyPlot C:\Users\datseris\.julia\packages\PyPlot\mQXSC\src\colormaps.jl:49
┌ Warning: `getindex(o::PyObject, s::Union{Symbol, AbstractString})` is deprecated in favor of dot overloading (`getproperty`) so elements should now be accessed as e.g. `o.s` instead of `o[:s]`.
│   caller = init_colormaps() at colormaps.jl:50
└ @ PyPlot C:\Users\datseris\.julia\packages\PyPlot\mQXSC\src\colormaps.jl:50
┌ Warning: `getindex(o::PyObject, s::Union{Symbol, AbstractString})` is deprecated in favor of dot overloading (`getproperty`) so elements should now be accessed as e.g. `o.s` instead of `o[:s]`.
│   caller = init_colormaps() at colormaps.jl:52
└ @ PyPlot C:\Users\datseris\.julia\packages\PyPlot\mQXSC\src\colormaps.jl:52
┌ Warning: `getindex(o::PyObject, s::Union{Symbol, AbstractString})` is deprecated in favor of dot overloading (`getproperty`) so elements should now be accessed as e.g. `o.s` instead of `o[:s]`.
│   caller = init_colormaps() at colormaps.jl:53
└ @ PyPlot C:\Users\datseris\.julia\packages\PyPlot\mQXSC\src\colormaps.jl:53
┌ Warning: `getindex(o::PyObject, s::Union{Symbol, AbstractString})` is deprecated in favor of dot overloading (`getproperty`) so elements should now be accessed as e.g. `o.s` instead of `o[:s]`.
│   caller = top-level scope at none:0
```

crstnbr:
@Datseris  I'm fixing the merge conflicts (today). Afterwards I'll try to fix those warnings, which are caused by internal uses of `getindex`. I guess will also have to update the README etc before this can be merged.

Datseris:
perfect, thanks for the amazing work on this PR! 

Notice that most of my warnings come from files within PyPlot so someone else should open a PR there.

src/PyCall.jl
```diff
@@ -325,7 +322,10 @@ function setproperty!(o::PyObject, s::Union{Symbol,AbstractString}, v)
     o
 end
 
-getindex(o::PyObject, s::Union{Symbol, AbstractString}) = getproperty(o, s)
+function getindex(o::PyObject, s::Union{Symbol, AbstractString})
+    Base.depwarn("`getindex(o::PyObject, s::Union{Symbol, AbstractString})` is deprecated in favor of dot overloading (`getproperty`) so elements should now be accessed as e.g. `o.s` instead of `o[:s]`.", :getindex)
```
stevengj: Should be a separate deprecation warning for `s::Symbol` and `s::AbstractString`

crstnbr:
Ok, AFAICS this PR should now be technically mergable.  I fixed all the merge conflicts, deprecated the `getindex` and `setindex!` methods and fixed all corresponding internal deprecation warnings (at least those that show up in the tests).

However,
* the README.md should be revisited. I made some basic adjustments but I guess a proper revisit/rewrite of the relevant section would be appreciated. As I'm not a native english speaker, it might make sense if someone else takes care of that?
* I haven't tested the `getindex` -> `getproperty` changes in `gui.jl` because I'm on Windows.

Since it's already after midnight, I guess it's time to go to bed :)

stevengj:
AppVeyor failure is an unrelated network glitch.

tkf:
Big thanks to @crstnbr for the big merge!

PR #521, opened 2018-08-14
remove imports of start, next and done from Base

mkborregaard:
and replace with import of iterate.
Finalizes the update to the new iterator protocol.


stevengj:
I would rather not import iterate. Nowadays I think the preferred style is to explicitly qualify functions that you are adding methods to rather than importing.

mkborregaard:
Sure thing, I just aimed for consistency with the existing code. I've removed the import. There are still 3 test errors when I test PyCall locally on 1.0, but they seem unrelated to this.

Datseris:
Also added in #522 which fixes "all" of 1.0 issues.

mkborregaard:
Alright, I'll close this then.

PR #534, opened 2018-08-22
fix non-reentrancy and lack of gc in recent pycall optimization

### LINKED ISSUE #533: re-entrancy issue in new pycall!
Because the new `pycall!` implementation in #492 uses a global cache of Python tuples, we encounter some sticky problems due to lack of re-entrancy.   In particular, if the `PyObject` call in `pysetarg!` itself calls `pycall` with the same number of arguments, we get the wrong results.

This can happen with the range conversions:
```jl
julia> f = py"lambda x,y,z: (x,y,z)"
PyObject <function <lambda> at 0x145cba510>

julia> pycall(f, PyObject, 3,4,5)
PyObject (3, 4, 5)

julia> pycall(f, PyObject, 3:6,4:10,5:11)  # WRONG RESULT:
PyObject (5, 12, range(5, 12))
```

cc @JobJob

stevengj:
Fixes #533

### DIFF
--- a/src/pyfncall.jl
+++ b/src/pyfncall.jl
@@ -29,7 +29,23 @@ function _pycall!(ret::PyObject, o::Union{PyObject,PyPtr}, args,
     end
     check_pyargsptr(nargs)
     pyargsptr = pyarg_tuples[nargs+1]
-    return _pycall!(ret, pyargsptr, o, args, nargs, kw) #::PyObject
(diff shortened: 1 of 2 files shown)


stevengj:
Even this fix is too complicated, and I missed some corner cases (e.g. #536).   Better for now to just get rid of the whole tuple-caching machinery: 8d300021792b8146483e0df42afd11ccd665fd99

PR #540, opened 2018-08-26
Segfault with Python 3.7

tkf:
Running tests with Python 3.7 segfaults. I try to reproduce the issue on Travis in this PR.

The minimal code I found to cause the segfault was:

```julia
using PyCall, Dates; convert(PyObject, DateTime(2013))
```

Added: closes #544

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,32 +1,32 @@
 environment:
   matrix:
   - julia_version: 0.6
-    PYTHON: "C:\\Python27\\python.exe"
+    python_version: "27"
   - julia_version: 0.6
-    PYTHON: "C:\\Python34\\python.exe"
+    python_version: "34"
   - julia_version: 0.6
-    PYTHON: "Conda"
+    python_version: "Conda"
 
   - julia_version: 0.7
-    PYTHON: "C:\\Python27\\python.exe"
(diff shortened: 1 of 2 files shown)


tkf:
Just applying the change in `PyDateTime_CAPI` struct from https://github.com/python/cpython/pull/5032/files#diff-fbac1e7a0d4bf4957e460fa4560277f4 fixes the segfault.

I didn't touch anything else to add full support of the new C APIs: https://docs.python.org/3/whatsnew/3.7.html#c-api-changes

tkf:
Actually, `TZInfoType` wasn't used either and stdlib `Dates` does not support timezone anyway.  So I guess this PR is good to go?

tkf:
Right, I agree that testing against the default Conda.jl setup is important.

I changed Python 3 version in appveyor.yml as you suggested.  If #543 CI *fails* and CI here succeeds, I think it's good to go.

tkf:
I rebased this PR onto #544 as it looks like previously tests in x64 AppVeyor was not run properly.  The tests were actually run against Python installed via conda since binary incompatibility was automatically detected.  Python 3.7 in both x86 and x64 correctly segfault.  So this PR should be good to go once the tests here all pass.

PR #543, opened 2018-08-27
[DON'T MERGE] Confirm that Python 3.7 segfaults in AppVeyor

tkf:
(to be merged into #540)

PR #544, opened 2018-08-27
Use correct Python version in x64 AppVeyor

tkf:
solves https://github.com/JuliaPy/PyCall.jl/pull/543#issuecomment-416404216

Closes #520

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,32 +1,32 @@
 environment:
   matrix:
   - julia_version: 0.6
-    PYTHON: "C:\\Python27\\python.exe"
+    python_version: "27"
   - julia_version: 0.6
-    PYTHON: "C:\\Python34\\python.exe"
+    python_version: "34"
   - julia_version: 0.6
-    PYTHON: "Conda"
+    python_version: "Conda"
 
   - julia_version: 0.7
-    PYTHON: "C:\\Python27\\python.exe"
+    python_version: "27"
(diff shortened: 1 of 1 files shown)


tkf:
I rebased this PR onto #544 as it looks like previously tests in x64 AppVeyor was not run properly.  The tests were actually run against Python installed via conda since binary incompatibility was automatically detected.  Python 3.7 in both x86 and x64 correctly segfault.  So this PR should be good to go once the tests here all pass.

tkf:
It's only in nightly, right? I saw a similar failure in PyJulia too: https://github.com/JuliaPy/pyjulia/pull/190#discussion_r213635490

stevengj:
Yes, looks like Travis nightly is failing too (e.g. https://travis-ci.org/JuliaPy/PyCall.jl/builds/423138539?utm_source=github_status&utm_medium=notification)

tkf:
Wouldn't it be better to have a separate repository (say, `JuliaPy/PyCall.jl-nightly`) that runs CIs against `julia` nightly using cron job?  The result would change at most once a day so it doesn't make sense to run it for every push and PRs.  Also, with a cron job we can trigger the CI without changes in PyCall.jl or PyJulia.

stevengj:
That seems like too much trouble to set up … would be easier to just change the travis and appveyor scripts so that they ignore failures on nightly.

PR #549, opened 2018-09-01
Libdl.dlsym_e now returns 'nothing' rather than 'C_NULL' when a symbo…

macd:
…l is not found.  Updated hassym


macd:
Just found out that the change to Libdl.dlsym_e might be reverted, so I am closing for now.

Issue #551, opened 2018-09-01
Passing PyObject through include/eval turns it into PyNULL (which causes segfaults)


tkf:
```julia
julia> using PyCall

julia> obj = pyimport("sys")  # get some PyObject
PyObject <module 'sys' (built-in)>

julia> py"""
       ns = {}
       def set(name):
           ns[name] = $include_string($Main, name)
       """

julia> py"set"("obj")

julia> obj
PyObject NULL

julia> py"lambda x: x"(obj)

signal (11): Segmentation fault
```

There are two problems: (1) `py"set"("obj")` turns `obj` to `PyNULL` (ref count issue?) and (2) passing `PyNULL` to Python function is not guarded enough.

Note that `py"set"` is modeling what we do in PyJulia all the time so it's not a contrived example.

PR #552, opened 2018-09-01
Guard against PyNULL in arguments

tkf:
fixes (the easier part of) #551

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -597,8 +597,16 @@ end
 function PyObject(d::AbstractDict)
     o = PyObject(@pycheckn ccall((@pysym :PyDict_New), PyPtr, ()))
     for k in keys(d)
+        pykey = PyObject(k)
+        pyval = PyObject(d[k])
+        if ispynull(pykey)
+            throw(ArgumentError("Dictionary keyword contains a PyNULL"))
+        end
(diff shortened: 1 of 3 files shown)


src/conversions.jl
```diff
@@ -597,8 +597,16 @@ end
 function PyObject(d::AbstractDict)
     o = PyObject(@pycheckn ccall((@pysym :PyDict_New), PyPtr, ()))
     for k in keys(d)
+        pykey = PyObject(k)
+        pyval = PyObject(d[k])
```
stevengj: Would be cleaner to have a function
```
function valid_pyobject(x)
    o = PyObject(x)
    ispynull(o) && error("invalid PyNULL object created from $x")
    return o
end
```
and call that in cases where we want to check for NULL results.

stevengj:
An unexpected NULL object will generally indicate a bug somewhere else, as is the case here (#553), so I'm not sure it's worthwhile to insert these argument checks everywhere.

tkf:
I saw `ispynull(o)` in `pystring` so I thought calling it makes sense. But I guess you have to use it there, as unless it's impossible to have `PyNULL` in REPL.

Anyway, I'm OK with not adding this support.  Closing it.

PR #553, opened 2018-09-01
Don't steal reference from existing Python object

### LINKED ISSUE #551: Passing PyObject through include/eval turns it into PyNULL (which causes segfaults)

tkf:
closes #551

I'm not very familiar with Python C API but I think this fixes the bug.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -133,6 +133,16 @@ function pystealref!(o::PyObject)
     return optr
 end
 
+"""
+    pyreturn(x) :: PyPtr
+
+Prepare `PyPtr` from `x` for passing it to Python.  If `x` is already
+a `PyObject`, the refcount is incremented.  Otherwise a `PyObject`
+wrapping/converted from `x` is created.
+"""
+pyreturn(x::Any) = pystealref!(PyObject(x))
+pyreturn(x::PyObject) = pyincref_(x.o)
+
 function Base.copy!(dest::PyObject, src::PyObject)
     pydecref(dest)
     dest.o = src.o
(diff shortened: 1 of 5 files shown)


stevengj:
Thanks, good catch!

stevengj:
Can you audit the other usages of `pystealref!` in case of similar problems?

It seems like it would be cleaner to define:
```jl
pyptr(x) = pystealref!(PyObject(x))
pyptr(x::PyObject) = pyincref_(x.o)
```
and then call `pyptr(x)` in all the cases where we would have previously called `pystealref!(PyObject(x))`

tkf:
I grepped `pystealref!` and found two more similar bugs which are fixed in the same way.

src/PyCall.jl
```diff
@@ -133,6 +133,16 @@ function pystealref!(o::PyObject)
...
 
+"""
+    pyreturn(x) :: PyPtr
+
+Prepare `PyPtr` from `x` for passing it to Python.  If `x` is already
+a `PyObject`, the refcount is incremented.  Otherwise a `PyObject`
+wrapping/converted from `x` is created.
+"""
+pyreturn(x::Any) = pystealref!(PyObject(x))
+pyreturn(x::PyObject) = pyincref_(x.o)
```
tkf: I named it `pyreturn` since I thought that's more intuitive.  Of course, it is super easy to replace it to `pyptr`.

stevengj:
Thanks!

PR #556, opened 2018-09-06
Use find_libpython.py in deps/build.jl

tkf:
as I suggested in https://github.com/JuliaPy/pyjulia/pull/190#issuecomment-418196179

<strike>
I think we can also remove this part (untouched ATM):

https://github.com/JuliaPy/PyCall.jl/blob/49029e0347a550ca15df06558fe5aad9ca80479a/deps/build.jl#L95-L134

`ctypes.util.find_library` used in `find_libpython.py` is probably good enough.  What do you think?
</strike>

(Edit: the fallback should stay: https://travis-ci.org/JuliaPy/PyCall.jl/builds/425108395)

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -78,8 +78,12 @@ function find_libpython(python::AbstractString)
     for lib in libs
         lib = splitext(lib)[1]
         try
-            return (Libdl.dlopen(lib, dlopen_flags),
-                    lib)
+            libpython = Libdl.dlopen(lib, dlopen_flags)
+            # Store the fullpath to libpython in deps.jl.  This makes
+            # it easier for users to investigate Python setup
+            # PyCall.jl trying to use.  It also helps PyJulia to
+            # compare libpython.
+            return (libpython, Libdl.dlpath(libpython))
         catch e
             push!(error_strings, string("dlopen($lib) ==> ", e))
         end
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -49,46 +49,26 @@ const dlprefix = Compat.Sys.iswindows() ? "" : "lib"
 
 # return libpython name, libpython pointer
 function find_libpython(python::AbstractString)
-    # it is ridiculous that it is this hard to find the name of libpython
-    v = pyconfigvar(python,"VERSION","")
-    libs = [ dlprefix*"python"*v, dlprefix*"python" ]
-    lib = pyconfigvar(python, "LIBRARY")
-    lib != "None" && pushfirst!(libs, splitext(lib)[1])
-    lib = pyconfigvar(python, "LDLIBRARY")
-    lib != "None" && pushfirst!(pushfirst!(libs, basename(lib)), lib)
-    libs = unique(libs)
-
-    # it is ridiculous that it is this hard to find the path of libpython
-    libpaths = [pyconfigvar(python, "LIBDIR"),
-                (Compat.Sys.iswindows() ? dirname(pysys(python, "executable")) : joinpath(dirname(dirname(pysys(python, "executable"))), "lib"))]
-    if Compat.Sys.isapple()
-        push!(libpaths, pyconfigvar(python, "PYTHONFRAMEWORKPREFIX"))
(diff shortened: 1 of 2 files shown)


deps/build.jl
```diff
@@ -49,46 +49,26 @@ const dlprefix = Compat.Sys.iswindows() ? "" : "lib"
...
+    cmd = `$python $(joinpath(@__DIR__, "find_libpython.py"))`
+    if "yes" == get(ENV, "PYCALL_DEBUG_BUILD", "no")
+        cmd = `$cmd --verbose`
+    end
+    if VERSION < v"0.7-"
+        pipe, process = open(cmd)
+    else
+        pipe = process = open(cmd)
+    end
+    libpy_name = read(pipe.out, String)
```
tkf: ATM `find_libpython.py` is run in the default "print the best one" mode. I can also run it with `--list-all` option, loop over all the candidate paths and use the first one that `Libdl.dlopen` succeeds. What do you think?

deps/find_libpython.py
```diff
@@ -0,0 +1,376 @@
+#!/usr/bin/env python
+
+"""
+Locate libpython associated with this Python executable.
+"""
+
+# License
```
tkf: I inlined the license so that it's easier to grab this file and put it in other projects.

deps/find_libpython.py
```diff
@@ -0,0 +1,376 @@
...
+
+    # List candidates for directories in which libpython may exist
+    lib_dirs = []
+    append_truthy(lib_dirs, sysconfig.get_config_var('LIBPL'))
+    append_truthy(lib_dirs, sysconfig.get_config_var('srcdir'))
+    append_truthy(lib_dirs, sysconfig.get_config_var("LIBDIR"))
+
+    # LIBPL seems to be the right config_var to use.  It is the one
+    # used in python-config when shared library is not enabled:
+    # https://github.com/python/cpython/blob/v3.7.0/Misc/python-config.in#L55-L57
```
tkf: So it seems adding `LIBPL` was the key (it was not in the PyJulia version; I'll back-port it later).  This script finds libpython in case the old julia function couldn't: https://github.com/JuliaPy/PyCall.jl/issues/565#issuecomment-421340493

tkf:
I think it's good to go.  CI on appveyor is [failing even on master](https://ci.appveyor.com/project/StevenGJohnson/pycall-jl-nu3aa/build/1.0.339) at:

https://github.com/JuliaPy/PyCall.jl/blob/0f685cf02b6214ffc40b0aba7ab657749d587f5a/test/runtests.jl#L327-L330

which is irrelevant to this PR (the build is already succeeded at this stage).

tkf:
ref: https://discourse.julialang.org/t/error-building-pycall-in-v0-7-conda-jl-did-not-install-python/15104

andreasnoack:
It would be great to get this one merged since it fixes https://github.com/JuliaPy/PyCall.jl/issues/565 which currently causes many failures on CIBot.

tkf:
Backported to https://github.com/JuliaPy/pyjulia/pull/203 and the tests pass there too.

deps/find_libpython.py
```diff
@@ -0,0 +1,394 @@
...
+    # For macOS:
+    append_truthy(lib_dirs, sysconfig.get_config_var("PYTHONFRAMEWORKPREFIX"))
+
+    lib_dirs.append(sys.exec_prefix)
+    lib_dirs.append(os.path.join(sys.exec_prefix, "lib"))
+
+    lib_basenames = list(candidate_names(suffix=suffix))
+
+    for directory in lib_dirs:
+        for basename in lib_basenames:
```
stevengj: Might want to guard this with `if directory:` in case `get_config_var` returned `None`

tkf: That's done with `append_truthy` utility function defined above. If my usage of utility function is confusing I can remove it and add explicit `if`s.

stevengj: No, I see.

stevengj:
I still wish we had a simpler solution, but I will merge this since it seems like it handles cases that our current code doesn't.

PR #558, opened 2018-09-06
Fix memory leak

### LINKED ISSUE #542: Memory leak with different libraries (numpy, pyTorch, OpenAI Gym)
I am currently working with OpenAI Gym using PyCall and the following code (randomly plays games of Atari 2600 Pong) uses up memory it never frees (on the order of 100MiB a second).

```julia
using PyCall

@pyimport gym
env = gym.make("Pong-v0")
env[:reset]()

for i in 1:100000
    obs, reward, done, info = env[:step](rand([3,4]))
    env[:render]()
    done && env[:reset]()
end

env[:close]()
```
To try you should `pip install gym[all]`

The same example in Python does not have this problem. Any idea how to fix this? Or how do I find out what exactly is happening here?

This seems to occur with other libraries (numpy, PyTorch) as well. See later comments on this issue.

tkf:
Since `PyObject_Call` returns a new reference, we don't need to increment the reference count.

https://docs.python.org/3/c-api/object.html#c.PyObject_Call

closes #542


stevengj:
Thanks!

PR #559, opened 2018-09-06
Make anaconda_conda work in PyJulia imported in python interpreter statically-linked to libpython.

tkf:


src/PyCall.jl
```diff
@@ -622,6 +622,9 @@ function anaconda_conda()
     if conda || !occursin("conda", unsafe_string(ccall(@pysym(:Py_GetVersion), Ptr{UInt8}, ())))
         return ""
     end
+    if isempty(pyprogramname)
+        return ""
+    end
```
stevengj: This means it will never find `conda` for statically linked libpython.  Is that what we want?

tkf: Yes. But, I think that's fine because it's called from a Python executable. I don't think Python packages don't usually install Python package dependencies automatically at runtime.

Having said that, `pyimport("sys")[:executable]` is *technically* a valid option here.  But I don't think that's the right thing to do.  Please see #560

PR #562, opened 2018-09-12
Add bin/python that mimics Python CLI

tkf:
This PR implements the tricky part (= libjulia initialization) of PyJulia "inside-out."  By that I just mean to provide a simple *Julia* script (= `bin/python`) that drops into Python interpreter via PyCall.  Inside this Python interpreter, setting up PyJulia is "free" and don't care if the Python executable is statically linked or not because it does not use the Python executable.

Now I can do

```console
$ bin/python -c 'from julia import Base; Base.banner()'
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.0.0 (2018-08-08)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |
```

even with Python from conda (whose `sys.executable` is statically linked to libpython).

### DIFF
--- a/lib/pseudo_python_cli.py
+++ b/lib/pseudo_python_cli.py
@@ -1,0 +1,132 @@
+'''
+Pseudo Python command line interface.
+
+It tries to mimic a subset of Python CLI:
+https://docs.python.org/3/using/cmdline.html
+'''
+
+import argparse
+import code
+import runpy
+import sys
+import traceback
+
+
+def python(module, command, script, script_args, interactive):
+    if command:
+        sys.argv[0] = "-c"
+    elif script:
+        sys.argv[0] = script
+    sys.argv[1:] = script_args
+
+    banner = ""
+    try:
+        if command:
+            scope = {}
+            exec(command, scope)
+        elif module:
+            scope = runpy.run_module(
+                module,
+                run_name="__main__",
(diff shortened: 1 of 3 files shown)


lib/pseudo_python_cli.py
```diff
@@ -0,0 +1,132 @@
+'''
+Pseudo Python command line interface.
+
+It tries to mimic a subset of Python CLI:
+https://docs.python.org/3/using/cmdline.html
+'''
```
stevengj: Can't this be done in Julia?

stevengj:
I'm unclear on how you plan to use this in pyjulia — can you clarify?

If this is mainly intended for use in pyjulia, why does it need to be in PyCall.jl?

tkf:
> If this is mainly intended for use in pyjulia, why does it need to be in PyCall.jl?

Ah, yes, that's a good point. Historically, I started writing this patch in the form that provides PyJulia-like Python package that does not require PyJulia so it made sense to have it in PyCall.  But then I totally agree that it should be in PyJulia.

PR #571, opened 2018-09-17
remove const declaration in `@pydef` which is deprecated in local scope

marius311:
`@pydef` declares the class `const` which if done in the local scope is deprecated in 0.7 / errors on 1.0. I just removed the `const`. I suppose it could still remain there in the global scope, which would offer slightly better type-stability / performance, but if you're building / calling into a Python class I would argue the overhead is likely to be insignificant in most cases.


stevengj:
I guess you can still do
```jl
const MyClass = PyCall.@pydef_object ....
```
but in a module setting I guess you couldn't define this as a `const` global anyway because it wouldn't be precompile-safe.

PR #591, opened 2018-10-13
Support closing the first (root) window in Tk

tkf:
Maybe fix https://github.com/JuliaPy/PyPlot.jl/issues/316

### DIFF
--- a/src/gui.jl
+++ b/src/gui.jl
@@ -195,13 +195,9 @@ end
 # based on https://github.com/ipython/ipython/blob/7.0.1/IPython/terminal/pt_inputhooks/tk.py
 function Tk_eventloop(sec::Real=50e-3)
     Tk = pyimport(tkinter_name())
-    if pyversion < v"3"
-        _tkinter = Tk[:tkinter]
-    else
-        _tkinter = pyimport("_tkinter")
-    end
+    _tkinter = pyimport("_tkinter")
     flag = _tkinter[:ALL_EVENTS] | _tkinter[:DONT_WAIT]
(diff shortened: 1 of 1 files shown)


tkf:
OK. Applied the fix.

stevengj:
Why the change from `root.update` to `dooneevent`?

tkf:
Right, I should have left comments. Calling `update` after closing first windows throws.  You can see it by

```julia
julia> using PyCall, PyPlot

julia> figure(1)
       plot([1,2,3])
       figure(2)
       plot([1,2,3])
       root = pyimport("tkinter")["_default_root"]
       close(1)

julia> pycall(root["update"], PyObject)
ERROR: PyError ($(Expr(:escape, :(ccall(#= /home/takafumi/.julia/dev/PyCall/src/pyfncall.jl:44 =# @pysym(:PyObject_Call), PyPtr, (PyPtr, PyPtr, PyPtr), o, pyargsptr, kw))))) <class '_tkinter.TclError'>
TclError('can\'t invoke "update" command: application has been destroyed')
  File "/usr/lib64/python3.7/tkinter/__init__.py", line 1174, in update
    self.tk.call('update')
```

I used `dooneevent` simply because IPython uses it: https://github.com/ipython/ipython/blob/7.0.1/IPython/terminal/pt_inputhooks/tk.py

It seems that `dooneevent` (which presumably wraps `Tcl_DoOneEvent`) acts on the whole eventloop: https://www.tcl.tk/man/tcl8.0/TclLib/DoOneEvent.htm

So I guess that's why it works even after the `root` window is closed.

Issue #593, opened 2018-10-16
Iterate `PyObject` without conversion


jw3126:
I am in a situation, where I want to iterate over a `PyObject`, without converting the items into julia objects.
My quick and dirty solution was to copy the `iterate` code and remove the conversion here:
https://github.com/JuliaPy/PyCall.jl/blob/ae9bbba3c54958d6762f90782c5630daa6386c54/src/pyiterator.jl#L37

Is there a better way? Would you like a PR that adds a lower level function without conversion (say pyiterate) and makes `iterate` a stub that just adds convert on top of this function?

tkf:
a dup of #153
see also #507

tkf:
closed by #594 (thanks!)

PR #594, opened 2018-10-16
add pyiterate

jw3126:
This is what I proposed in https://github.com/JuliaPy/PyCall.jl/issues/593

### DIFF
--- a/src/pyiterator.jl
+++ b/src/pyiterator.jl
@@ -29,14 +29,23 @@ end
 
     Base.done(po::PyObject, s) = ispynull(s[1])
 else
-    function Base.iterate(po::PyObject, s=_start(po))
+    function pyiterate(po::PyObject, s=_start(po))
         ispynull(s[1]) && return nothing
         sigatomic_begin()
         try
             nxt = PyObject(@pycheck ccall((@pysym :PyIter_Next), PyPtr, (PyPtr,), s[2]))
-            return (convert(PyAny, s[1]), (nxt, s[2]))
+            return (s[1], (nxt, s[2]))
         finally
             sigatomic_end()
+        end
+    end
+    function Base.iterate(po::PyObject, s=_start(po))
+        raw = pyiterate(po, s)
+        if raw == nothing
+            raw
+        else
+            pyitem, state = raw
+            convert(PyAny, pyitem), state
         end
     end
 end
--- a/src/pyiterator.jl
+++ b/src/pyiterator.jl
@@ -102,9 +102,7 @@ else
         # avoid the constructor that calls length
         # since that might be an expensive operation
         # even if length is cheap, this adds 10% performance
-
-        DummyNeverUsed = Base.SizeUnknown
-        piter = PyIterator{PyAny, DummyNeverUsed}(po)
+        piter = PyIterator{PyAny, Base.SizeUnknown}(po)
         iterate(piter, s)
     end
 end
--- a/src/pyiterator.jl
+++ b/src/pyiterator.jl
@@ -102,7 +102,9 @@ else
         # avoid the constructor that calls length
         # since that might be an expensive operation
         # even if length is cheap, this adds 10% performance
-        piter = PyIterator{PyAny, Base.IteratorSize}(po)
+
+        DummyNeverUsed = Base.SizeUnknown
+        piter = PyIterator{PyAny, DummyNeverUsed}(po)
         iterate(piter, s)
     end
 end
--- a/src/pyiterator.jl
+++ b/src/pyiterator.jl
@@ -66,6 +66,9 @@ else
             length(o)
             Base.HasLength()
         catch err
+            if !(err isa PyError && pyisinstance(err.val, @pyglobalobjptr :PyExc_TypeError))
+                rethrow()
+            end
             Base.SizeUnknown()
         end
     end
--- a/src/pyiterator.jl
+++ b/src/pyiterator.jl
@@ -84,7 +84,7 @@ else
     Base.eltype(::Type{<:PyIterator{PyAny}}) = Any
     Base.length(piter::PyIterator) = length(piter.o)
 
-    Base.IteratorSize(::Type{<: PyIterator{T,S}}) where {T,S} = S
+    Base.IteratorSize(::Type{<: PyIterator{T,S}}) where {T,S} = S()
 
     _start(piter::PyIterator) = _start(piter.o)
     
--- a/src/pyiterator.jl
+++ b/src/pyiterator.jl
@@ -64,7 +64,7 @@ else
     function _compute_IteratorSize(o::PyObject)
         S = try
             length(o)
-            Base.HasLength()
+            Base.HasLength
         catch err
             if !(err isa PyError && pyisinstance(err.val, @pyglobalobjptr :PyExc_TypeError))
                 rethrow()
--- a/src/pyiterator.jl
+++ b/src/pyiterator.jl
@@ -29,24 +29,56 @@ end
 
     Base.done(po::PyObject, s) = ispynull(s[1])
 else
-    function pyiterate(po::PyObject, s=_start(po))
+    """
+        PyIterator{T}(pyobject)
+
+    Wrap `pyobject::PyObject` into an iterator, that produces items of type `T`. To be more precise `convert(T, item)` is applied in each iteration. This can be useful to avoid automatic conversion of items into corresponding julia types.
+    ```jldoctest
+    julia> using PyCall
+
+    julia> l = PyObject([PyObject(1), PyObject(2)])
+    PyObject [1, 2]
+    
+    julia> piter = PyCall.PyIterator{PyAny}(l)
+    PyCall.PyIterator{PyAny}(PyObject [1, 2])
+    
+    julia> collect(piter)
+    2-element Array{Any,1}:
+     1
+     2
+    
+    julia> piter = PyCall.PyIterator{PyObject}(l)
+    PyCall.PyIterator{PyObject}(PyObject [1, 2])
+    
+    julia> collect(piter)
+    2-element Array{PyObject,1}:
+     PyObject 1
+     PyObject 2
+     ```
+    """
+    struct PyIterator{T}
+        o::PyObject
+    end
+    
+    Base.eltype(::Type{PyIterator{T}}) where T = T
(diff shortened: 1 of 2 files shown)


tkf:
I'd suggest `pyiterate` to behave like `pycall(o::PyObject, returntype::TypeTuple)`, i.e., `pyiterate(o::PyObject, eltype::DataType)` which returns a Julia iterator with conversion to the given `eltype`.  The default `Base.iterate` then would be defined as

```julia
Base.iterate(o::PyObject) = Base.iterate(pyiterate(o, PyAny))
```

Iteration without auto-conversion can be done with `pyiterate(o, PyObject)`:

```julia
for e in pyiterate(o, PyObject)
    e :: PyObject
end
```

We would need a helper sturct for this. Something like

```julia
struct PyIterator{T}
    o::PyObject
end

pyiterate(o::PyObject, eltype::DataType) = Pyiterator{eltype}(o)

Base.eltype(::Type{PyIterator{T}}) where T = T
Base.eltype(::Type{PyIterator{PyAny}}) = Any

Base.iterate(iter::PyIterator) = ...
```

(Maybe we don't need the helper/factory function `pyiterate` and should expose `PyIterator` as an API.)

@stevengj What do you think?

jw3126:
@tkf I like this proposal very much, and implemented it (without the helper function).

src/pyiterator.jl
```diff
@@ -29,16 +29,57 @@ end
...
+     ```
+    """
+    struct PyIterator{T}
+        o::PyObject
+    end
+    
+    Base.eltype(::Type{PyIterator{T}}) where T = T
+    Base.eltype(::Type{PyIterator{PyAny}}) = Any
+    Base.length(piter::PyIterator) = length(piter.o)
+
```
tkf: Since `HasLength()` is the default `IteratorSize`, I think we need something like:

```suggestion
    Base.IteratorSize(::Type{<: PyIterator}) = Base.SizeUnknown()

```

see: https://docs.julialang.org/en/latest/manual/interfaces/

(Though it looks like `Base.iterate` has been working well in PyCall without it. I wonder why...)

In principle, we can analyse `PyObject` in the constructor (i.e., run Python code `len(o)` and see if it throws a `TypeError`. We can then add it to the second type parameter of `PyIterator` (say). But I'm not sure how useful it is. Any optimisation based on iterator traits would be irrelevant if you are calling Python.

tkf: Ah, I get it. Code like `collect(zip(py"iter([1, 2, 3])", 1:3))` fails in the current PyCall.  Defining

```julia
Base.IteratorSize(::Type{PyObject}) = Base.SizeUnknown()
```

makes it work.  So, I think `Base.IteratorSize` should be defined for both `PyObject` and `PyIterator`.

tkf:
> getitem and getattr

For getitem, you can already do

```julia
julia> get(py"{'a': 1}"o, PyObject, "a")
PyObject 1
```

(which I just learnt recently).  For getattr, you can kind of do it already as well:

```julia
julia> pycall(pybuiltin("getattr"), PyObject, py"1"o, "__add__")
PyObject <method-wrapper '__add__' of int object at 0x7f8a4d3583e0>
```

though actually it's shorter to write

```julia
julia> pycall(py"lambda x: x.__add__", PyObject, py"1"o)
PyObject <method-wrapper '__add__' of int object at 0x7f8a4d3583e0>
```

jw3126:
Thanks for these tips! I was not aware of `get` either, nice!

src/pyiterator.jl
```diff
@@ -29,16 +30,59 @@ end
...
+    julia> piter = PyCall.PyIterator{PyAny}(l)
+    PyCall.PyIterator{PyAny}(PyObject [1, 2])
+    
+    julia> collect(piter)
+    2-element Array{Any,1}:
+     1
+     2
+    
+    julia> piter = PyCall.PyIterator{PyObject}(l)
+    PyCall.PyIterator{PyObject}(PyObject [1, 2])
```
stevengj: Probably this should be the default: there should be a `PyIterator(o::PyObject) = PyIterator{PyObject}(o)` constructor.

src/pyiterator.jl
stevengj: The advantage of analyzing the `PyObject` in the constructor to see whether it supports a length might be that there may be some functions acting on iterators that will throw an error for `SizeUnknown` iterators.  We don't want such functions to fail on Python objects if we can help it.

jw3126: I like that the current implementation is dirt simple. I think we should keep it for now and only change it, if we have real code, where it causes trouble.

src/pyiterator.jl
jw3126: Good idea, I added it.

test/runtests.jl
```diff
@@ -666,4 +666,36 @@ def try_call(f):
...
+        piter = PyCall.PyIterator(l)
+        @test length(piter) == 2
+        @test length(collect(piter)) == 2
+        r1, r2 = collect(piter)
+        @test r1.o === i1.o
+        @test r2.o  === i2.o
+    end
+
+    # 594
+    @test collect(zip(py"iter([1, 2, 3])", 1:3)) == [(1, 1), (2, 2), (3, 3)]
```
tkf: Let's add

```julia
@test collect(zip(PyCall.PyIterator{Int}(py"iter([1, 2, 3])"), 1:3)) == [(1, 1), (2, 2), (3, 3)]
```

to check `Base.IteratorSize(::Type{<: PyIterator})`.  Let's also add

```julia
@test collect(zip(PyCall.PyIterator(py"[1, 2, 3]"o), 1:3)) == [(1, 1), (2, 2), (3, 3)]
```

once introspection-based `Base.IteratorSize` is implemented.

src/pyiterator.jl
```diff
@@ -29,16 +30,75 @@ end
...
+    """
+    struct PyIterator{T,S}
+        o::PyObject
+    end
+
+    function _compute_IteratorSize(o::PyObject)
+        S = try
+            length(o)
+            Base.HasLength()
+        catch err
```
tkf:
```suggestion
        catch err
            if !(err isa PyError && pyisinstance(err.val, @pyglobalobjptr :PyExc_TypeError))
                rethrow()
            end
```

would be safer.

(Note: we can't rely on the existence of `__len__` since the Python object may be a wrapper/proxy object.  i.e., it could be doing something like https://github.com/GrahamDumpleton/wrapt/blob/a4aeb4c7924ca18f4e9afa97172c95b1dad1e925/src/wrapt/wrappers.py#L381)

src/pyiterator.jl
jw3126: I misunderstood something, its not that complicated.

src/pyiterator.jl
```diff
@@ -29,16 +30,78 @@ end
...
+            if !(err isa PyError && pyisinstance(err.val, @pyglobalobjptr :PyExc_TypeError))
+                rethrow()
+            end
+            Base.SizeUnknown()
```
tkf: Why storing value instead of type? Wouldn't storing type be more useful as you can do `PyIterator{T, <:HasShape}` (if that's implemented)?

```suggestion
            Base.SizeUnknown
```

src/pyiterator.jl
```diff
@@ -29,16 +30,78 @@ end
...
+    ```
+    """
+    struct PyIterator{T,S}
+        o::PyObject
+    end
+
+    function _compute_IteratorSize(o::PyObject)
+        S = try
+            length(o)
+            Base.HasLength()
```
tkf:
```suggestion
            Base.HasLength
```

src/pyiterator.jl
```diff
@@ -29,16 +30,78 @@ end
...
+    function (::Type{PyIterator{T}})(o::PyObject) where {T}
+        S = _compute_IteratorSize(o)
+        PyIterator{T,S}(o)
+    end
+    
+    Base.eltype(::Type{<:PyIterator{T}}) where T = T
+    Base.eltype(::Type{<:PyIterator{PyAny}}) = Any
+    Base.length(piter::PyIterator) = length(piter.o)
+
+    Base.IteratorSize(::Type{<: PyIterator{T,S}}) where {T,S} = S
```
tkf:
```suggestion
    Base.IteratorSize(::Type{<: PyIterator{T,S}}) where {T,S} = S()
```

src/pyiterator.jl
jw3126: I agree, storing the type is better.

src/pyiterator.jl
```diff
@@ -29,16 +30,81 @@ end
...
+            return (convert(T,s[1]), (nxt, s[2]))
         finally
             sigatomic_end()
         end
     end
+    function Base.iterate(po::PyObject, s=_start(po))
+        # avoid the constructor that calls length
+        # since that might be an expensive operation
+        # even if length is cheap, this adds 10% performance
+        piter = PyIterator{PyAny, Base.IteratorSize}(po)
```
tkf: Good catch :)

stevengj: Do you mean to pass `Base.SizeUnkown` as the second parameter?

jw3126: Not really, the value of the second parameter is completely irrelevant.

stevengj: Isn't the second parameter used as the return value for `Base.IteratorSize(::Type{<: PyIterator{T,S}})`?

jw3126: In general sure. But the specific object we are talking about here only lives during a single `iterate` call and hence there is no chance to call `Base.IteratorSize` on it.

tkf: @jw3126 I thought `Base.IteratorSize` was OK since I realized that this type parameter won't be used anywhere.  But I think there was a few seconds of what??? moment.  So maybe just use `Base.SizeUnkown` and comment that this type parameter will not be used?  Using `Base.SizeUnkown` could also do a super micro-optimization (nano-optimization?) to reduce number of methods (nah, I guess not :smiley: ).

jw3126: I did the change, I also think it is clearer now. Also consider, that `Base.SizeUnkown` has less characters, so we can save lots of parse time in addition :smile:.

jw3126:
So can this be merged? Or is there any remaining problem in this PR?

Issue #592, opened 2018-10-16
Building pyCalll fails after update from 1.18.4 to 1.18.5 on Julia 1.0.1


kurbkid:
output of package update:
```
(v1.0) pkg> update
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
 Resolving package versions...
 Installed Compat ───────── v1.3.0
 Installed BinaryProvider ─ v0.5.2
 Installed PyCall ───────── v1.18.5
 Installed Conda ────────── v1.1.1
 Installed CMake ────────── v1.1.0
  Updating `~/.julia/environments/v1.0/Project.toml`
  [8f4d0f93] ↑ Conda v1.0.2 ⇒ v1.1.1
  [438e738f] ↑ PyCall v1.18.4 ⇒ v1.18.5
  Updating `~/.julia/environments/v1.0/Manifest.toml`
  [b99e7846] ↑ BinaryProvider v0.5.1 ⇒ v0.5.2
  [631607c0] ↑ CMake v1.0.2 ⇒ v1.1.0
  [34da2185] ↑ Compat v1.2.0 ⇒ v1.3.0
  [8f4d0f93] ↑ Conda v1.0.2 ⇒ v1.1.1
  [438e738f] ↑ PyCall v1.18.4 ⇒ v1.18.5
  Building Conda ─→ `~/.julia/packages/Conda/uQitS/deps/build.log`
  Building PyCall → `~/.julia/packages/PyCall/0jMpb/deps/build.log`
┌ Error: Error building `PyCall`:
│ ERROR: LoadError: UndefVarError: v not defined
│ Stacktrace:
│  [1] find_libpython(::String) at /home/dicbro/.julia/packages/PyCall/0jMpb/deps/build.jl:101
│  [2] top-level scope at logging.jl:313
│  [3] top-level scope at /home/dicbro/.julia/packages/PyCall/0jMpb/deps/build.jl:170
│  [4] include at ./boot.jl:317 [inlined]
│  [5] include_relative(::Module, ::String) at ./loading.jl:1041
│  [6] include(::Module, ::String) at ./sysimg.jl:29
│  [7] include(::String) at ./client.jl:388
│  [8] top-level scope at none:0
│ in expression starting at /home/dicbro/.julia/packages/PyCall/0jMpb/deps/build.jl:144
[...]

PR #578, opened 2018-09-21
Support virtual environment

tkf:
This PR grew out of a small feature addition to a patch that includes a few more. Here is a new summary:

* Main feature addition: Control Python environment at run-time by the environment variables `PYCALL_JL_RUNTIME_PYTHON` (and `PYCALL_JL_RUNTIME_PYTHONHOME`).
* fix #410: Support building Pycall against a Python executable generated by `python -m venv`.
* Refactoring: I merged how `Py_SetPythonHome` and `Py_SetProgramName` are `ccall`ed with and without `PYCALL_JL_RUNTIME_PYTHON`.  `wpyprogramname` and `wPYTHONHOME` are removed.

---

Original summary:

I suggest to let users control Python environment at runtime by (say) `ENV["PYCALL_JL_RUNTIME_PYTHON"]` (and `ENV["PYCALL_JL_RUNTIME_PYTHONHOME"]`).  This can be used to make sure that Julia and Python environments for a project are fully reproducible.  Combining `Pkg3.jl` and `pipenv`, a typical usage would be very simple:

```console
$ ls
activate.jl  Manifest.toml  Pipfile  Pipfile.lock  Project.toml
$ cat activate.jl
import Pkg
Pkg.activate(@__DIR__)
ENV["PYCALL_JL_RUNTIME_PYTHON"] = strip(read(setenv(`pipenv --py`, dir=@__DIR__), String))
$ julia -i activate.jl
julia> using PyCall
julia> pyimport("numpy")  # loads numpy specified by Pipfile
```

PR #595, opened 2018-10-17
Test find_libpython

tkf:
TODO:

- [x] Wait for #578 and then rebase.

---

I think `deps/build.jl` is complex enough to require some tests.  I should have added some tests in #556.

This PR fixes a yet another bug introduced in #556.  From the first version of the comment https://github.com/JuliaPy/PyCall.jl/issues/592#issuecomment-430517684:

```
│ ERROR: LoadError: UndefVarError: libpath_lib not defined
│ Stacktrace:
│  [1] show_dlopen_error(::ErrorException) at /home/dicbro/.julia/packages/PyCall/0jMpb/deps/build.jl:63
│  [2] find_libpython(::String) at /home/dicbro/.julia/packages/PyCall/0jMpb/deps/build.jl:78
│  [3] top-level scope at logging.jl:313
│  [4] top-level scope at /home/dicbro/.julia/packages/PyCall/0jMpb/deps/build.jl:170
│  [5] include at ./boot.jl:317 [inlined]
│  [6] include_relative(::Module, ::String) at ./loading.jl:1041
│  [7] include(::Module, ::String) at ./sysimg.jl:29
│  [8] include(::String) at ./client.jl:388
│  [9] top-level scope at none:0
│ in expression starting at /home/dicbro/.julia/packages/PyCall/0jMpb/deps/build.jl:144
└ @ Pkg.Operations /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Pkg/src/Operations.jl:1069
```

The error was produced here:

https://github.com/JuliaPy/PyCall.jl/blob/ae9bbba3c54958d6762f90782c5630daa6386c54/deps/build.jl#L61-L63

closes #596

---

I opened an alternative fix-only version here: #596


deps/buildutils.jl
```diff
@@ -0,0 +1,102 @@
...
+import Conda, Compat.Libdl
+
+# Fix the environment for running `python`, and setts IO encoding to UTF-8.
+# If cmd is the Conda python, then additionally removes all PYTHON* and
+# CONDA* environment variables.
+function pythonenv(cmd::Cmd)
+    env = copy(ENV)
+    if dirname(cmd.exec[1]) == abspath(Conda.PYTHONDIR)
+        pythonvars = String[]
+        for var in keys(env)
```
stevengj: shouldn't we just call `Conda._set_conda_env`?

tkf: Yeah, this may be required for Conda with Python >= 3.7.  https://github.com/JuliaPy/Conda.jl/issues/125#issuecomment-423755365

It's probably better to define `Conda.run` which is like `Base.run` but with appropriate environment variables. I think you'll need it in IJulia as well to let Python running Jupyter server to load appropriate DLL in Windows.  Using "private" function starting with underscore is a bit awkward.

PR #596, opened 2018-10-17
Fix show_dlopen_error

tkf:
In case the testing "infrastructure" I introduced in #595 was too much, this is a quick fix version.

closes #595


stevengj:
Merging the quick fix; we can merge the longer testing PR later.

PR #602, opened 2018-10-23
Test also with Python 3 in macOS

tkf:
based on https://github.com/pytest-dev/pytest/blob/799b72cf6f35c4c4c73797303d3f78c12a795f77/.travis.yml#L38-L52

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,9 +45,6 @@ matrix:
 # https://github.com/pytest-dev/pytest/blob/799b72cf6f35c4c4c73797303d3f78c12a795f77/.travis.yml#L38-L52
 # https://github.com/pytest-dev/pytest/pull/3893
 
-before_install:
-    - test "x$TRAVIS_OS_NAME" = xosx -o "x$PYTHON" = xConda || (sudo apt-get -qq update && sudo apt-get install -y $PYTHON)
-
 after_success:
     - julia -e 'Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
 notifications:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,11 +18,32 @@ env:
...
+    - &test-macos-py2
+      language: julia
+      os: osx
+      env: PYTHON=python
+      julia: 0.6
+    - &test-macos-py3
+      <<: *test-macos-py2
+      before_install:
+        - brew update
+        - brew upgrade python || echo "Ignoring errors..."
```
tkf: Strangely, python3 is correctly installed https://travis-ci.org/JuliaPy/PyCall.jl/jobs/444940334#L1356 even though the command fails https://travis-ci.org/JuliaPy/PyCall.jl/jobs/444940334#L1244

tkf: Actually, we've been using the same trick in PyJulia: https://github.com/JuliaPy/pyjulia/blob/f7a552d2eabbb5045c7182ba056d0867c05830f9/.travis.yml#L49

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: julia
 os:
     - linux
-    - osx
```
stevengj: why is this line deleted?

tkf: `osx` tests are added via `matrix.include`: https://github.com/JuliaPy/PyCall.jl/pull/602/files#diff-354f30a63fb0907d4ad57269548329e3R20

I think there would be duplicates if we left `osx` here.

PR #603, opened 2018-10-28
Finalize Python runtime while shutting down Julia

tkf:
closes https://github.com/JuliaPy/pyjulia/issues/208

It's a bit scary to play with the shutdown process of two GCs. But it seems to work in simple tests...

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -98,11 +98,12 @@ it is equivalent to a `PyNULL()` object.
 """
 ispynull(o::PyObject) = o.o == PyPtr_NULL
 
-function pydecref_(o::Union{PyPtr,PyObject}) :: Nothing
+function pydecref_(o::Union{PyPtr,PyObject})
     if _finalized[]
-        return
+        return o
     end
     ccall(@pysym(:Py_DecRef), Cvoid, (PyPtr,), o)
+    return o
 end
 
 function pydecref(o::PyObject)
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -98,9 +98,11 @@ it is equivalent to a `PyNULL()` object.
 """
 ispynull(o::PyObject) = o.o == PyPtr_NULL
 
-function pydecref_(o::Union{PyPtr,PyObject})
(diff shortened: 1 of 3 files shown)


src/PyCall.jl
```diff
@@ -98,9 +98,11 @@ it is equivalent to a `PyNULL()` object.
 """
 ispynull(o::PyObject) = o.o == PyPtr_NULL
 
-function pydecref_(o::Union{PyPtr,PyObject})
+function pydecref_(o::Union{PyPtr,PyObject}) :: Nothing
```
tkf: We don't use the return value anywhere. So why not return `nothing` always?

tkf:
There is a segfault in Julia 0.7... https://travis-ci.org/JuliaPy/PyCall.jl/jobs/447533492#L767

tkf:
...which is apparently solved by going to non-closure (non-`$`) form of `@cfunction` (just a by-product of 0.6 support).

https://travis-ci.org/JuliaPy/PyCall.jl/builds/447551472

src/PyCall.jl
stevengj: It seems better to return something potentially useful (`o`), so that the result can be chained with some other function.  Also, this way it mirrors `pyincref`.   Why not just return `o` always?

tkf:
Re https://github.com/JuliaPy/PyCall.jl/pull/603#discussion_r229056706: I changed the return type of `pydecref_` back to the original. (I was uncertain about returning an "unsafe" value from it. But I guess usage like `pydecref_(pyobj).o = PyPtr_NULL` could be useful and decref'ed value was already unsafe anyway before the PR.)

crstnbr:
While resolving merge conflicts for #517, I found that the "atexit" tests from this PR are failing due to me having a `startup.jl` (which loads `BenchmarkTools`).

stevengj:
How does your startup.jl file affect the test?

tkf:
I should have used  `$(Base.julia_cmd()) --startup-file=no -e $script`

crstnbr:
Should I simply fix it as part of #517?

crstnbr:
Done.

PR #605, opened 2018-10-29
Allow @pydef to define classes with class variables

### LINKED ISSUE #513: Cannot use `@pydef` for classes with static attributes
It seems that `@pydef` does not work for classes with static attributes:

```julia
julia> @pydef mutable struct Foo
           class_attr = 1
           function __init__(self)
               println(self[:class_attr])
           end
       end

ERROR: AssertionError: Not a function definition: class_attr = 1
```

galenlynch:
Before this commit, `@pydef` and `@pydef_object` did not support the creation of
python classes with [class variables][1]. I have expanded `@pydef` to support
class variables with the following syntax:

```julia
@pydef mutable struct ObjectCounter
    obj_count = 0 # Class variable
    function __init__(::PyObject)
        ObjectCounter[:obj_count] += 1
    end
end
```

closes #513

[1]: https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables

### DIFF
--- a/src/pyclass.jl
+++ b/src/pyclass.jl
@@ -73,9 +73,9 @@ function parse_pydef(expr)
     method_syms = Dict{Any, Symbol}() # see below
     class_vars = Dict{Symbol, Any}()
     for line in lines
-        if line isa LineNumberNode || line isa Symbol || line.head == :line
-            continue
-        end
+        line isa LineNumberNode && continue
+        line isa Expr || error("Malformed line: $line")
+        line.head == :line && continue
         if isfunction(line)
             def_dict = splitdef(line)
             py_f = def_dict[:name]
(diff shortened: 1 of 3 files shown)


src/pyclass.jl
```diff
@@ -66,8 +71,12 @@ function parse_pydef(expr)
     getter_dict = Dict{Any, Symbol}() # python_var => jl_getter_name
     setter_dict = Dict{Any, Symbol}()
     method_syms = Dict{Any, Symbol}() # see below
+    class_vars = Dict{Symbol, Any}()
     for line in lines
-        if !isa(line, LineNumberNode) && line.head != :line # need to skip those
+        if line isa LineNumberNode || line isa Symbol || line.head == :line
```
stevengj: Wouldn't a `Symbol` line be an error?

galenlynch: Ah yes it would, I'll fix that!

galenlynch: Ah I remember why I had that now: `Symbol` does not have a `head`, which would cause `parse_pydef` to error prior to this PR. Silently ignoring the error is worse, however, so I'll figure something out.

galenlynch:
This PR makes it much easier to work with PyQt from Julia, btw. My main motivation was to get Qt slots and signals to work for Julia. PyQt requires that custom signals be assigned to class variables when the class is defined (dynamically added class variables do not work). This PR allows custom Qt signals to be added to classes defined in Julia.

PR #606, opened 2018-10-30
Add weakref support for pyjlwrap types

### LINKED ISSUE #158: support weak references to pyjlwrap types
Complicated Julia objects (like functions) are passed to Python by creating a new Python type that wraps around them and implements the appropriate interfaces (see [pytype.jl](https://github.com/stevengj/PyCall.jl/blob/master/src/pytype.jl)).  However, such wrapped objects currently don't support [weak references](https://docs.python.org/2/library/weakref.html) in Python.

To support weak references, [the PyTypeObject documentation](https://docs.python.org/2/c-api/typeobj.html) indicates that we need to set the `Py_TPFLAGS_HAVE_WEAKREFS` bit and define the `tp_weaklistoffset` and `tp_weaklist` fields.

Probably this is straightforward, but requires some careful digging through the documentation.

galenlynch:
I have added [weak reference support][1] for `pyjlwrap` types, following the
documentation for Python 3+. Adding weak reference support for Python 2.x seems
much the same as in 3.x, however 3.x does not use Py_TPFLAGS_HAVE_WEAKREFS,
which also does not seem necessary for 2.x. Simple testing suggests that this is
working as expected.

Closes JuliaPy/PyPlot.jl#21
Closes #158

[1]: https://docs.python.org/3/extending/newtypes.html#weak-reference-support

### DIFF
--- a/src/pyinit.jl
+++ b/src/pyinit.jl
@@ -27,7 +27,7 @@ const pyxrange = Ref{PyPtr}(0)
 function pyjlwrap_init()
     # PyMemberDef stores explicit pointers, hence must be initialized at runtime
     push!(pyjlwrap_members, PyMemberDef(pyjlwrap_membername,
-                                        T_PYSSIZET, sizeof_PyObject_HEAD, READONLY,
+                                        T_PYSSIZET, sizeof_pyjlwrap_head, READONLY,
                                         pyjlwrap_doc),
                             PyMemberDef(C_NULL,0,0,0,C_NULL))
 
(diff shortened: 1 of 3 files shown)


stevengj:
The HAVE_WEAKREFS flag is not needed since it is [part of https://github.com/JuliaPy/PyCall.jl/blob/ca792edda782d506607b1e96231f830b3cd8a7f4/src/pytype.jl#L299](Py_TPFLAGS_DEFAULT).  Apparently this flag became [obsolete in Python 2.1](https://github.com/python/cpython/commit/4e262a963190b201578b0b5d972035b1637f6bd5) when types became responsible for initializing the weakref field to `NULL` themselves.

PR #608, opened 2018-10-31
Make Julia exceptions wrapped in Python Exceptions more informative

### LINKED ISSUE #607: Is it possible to get a julia stack trace from PyCall functions?
I have been using Julia code wrapped with PyCall for a PyQt GUI. If there is a problem with my Julia code, it often causes the Julia process to terminate. The uninformative nature of the Julia exceptions wrapped in Python exceptions makes it very difficult to figure out where the problem lies. Is there any way to make these wrapped Julia exceptions more informative, e.g. with some form of Julia back trace?

An example error below:

```
Traceback (most recent call last):
  File "PyCall", line 1, in <lambda>
ArithmeticError: Julia exception: InexactError: trunc(Int32, -2.989629686e9)
Traceback (most recent call last):
  File "PyCall", line 1, in <lambda>
ArithmeticError: Julia exception: InexactError: trunc(Int32, -2.989629686e9)

signal (6): Aborted
in expression starting at no file:0
gsignal at /lib64/libc.so.6 (unknown line)
abort at /lib64/libc.so.6 (unknown line)
_ZNK14QMessageLogger5fatalEPKcz at /lib64/libQt5Core.so.5 (unknown line)
unknown function (ip: 0x7fb28b376456)
unknown function (ip: 0x7fb28b4d1634)
unknown function (ip: 0x7fb28b4d20d6)
_ZN11QMetaObject8activateEP7QObjectiiPPv at /lib64/libQt5Core.so.5 (unknown line)
unknown function (ip: 0x7fb28b4cd89d)
_PyMethodDef_RawFastCallKeywords at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyMethodDescr_FastCallKeywords at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalFrameDefault at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalCodeWithName at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyFunction_FastCallKeywords at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalFrameDefault at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalCodeWithName at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyFunction_FastCallKeywords at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalFrameDefault at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalCodeWithName at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyFunction_FastCallKeywords at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalFrameDefault at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalCodeWithName at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyFunction_FastCallKeywords at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalFrameDefault at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalCodeWithName at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyFunction_FastCallKeywords at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalFrameDefault at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalCodeWithName at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyFunction_FastCallDict at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyObject_Call_Prepend at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
PyObject_Call at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
unknown function (ip: 0x7fb28eab9c9a)
unknown function (ip: 0x7fb28eab9d82)
unknown function (ip: 0x7fb28b076b80)
unknown function (ip: 0x7fb28aed721e)
_ZN14QGraphicsScene5eventEP6QEvent at /lib64/libQt5Widgets.so.5 (unknown line)
unknown function (ip: 0x7fb28aed8d7a)
_ZN19QApplicationPrivate13notify_helperEP7QObjectP6QEvent at /lib64/libQt5Widgets.so.5 (unknown line)
_ZN12QApplication6notifyEP7QObjectP6QEvent at /lib64/libQt5Widgets.so.5 (unknown line)
unknown function (ip: 0x7fb28ae1723d)
_ZN16QCoreApplication15notifyInternal2EP7QObjectP6QEvent at /lib64/libQt5Core.so.5 (unknown line)
_ZN20QGraphicsViewPrivate21mouseMoveEventHandlerEP11QMouseEvent at /lib64/libQt5Widgets.so.5 (unknown line)
unknown function (ip: 0x7fb28aeec84e)
_PyMethodDef_RawFastCallKeywords at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyCFunction_FastCallKeywords at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyEval_EvalFrameDefault at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyFunction_FastCallDict at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
_PyObject_Call_Prepend at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
PyObject_Call at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
unknown function (ip: 0x7fb28eab9c9a)
unknown function (ip: 0x7fb28eab9d82)
unknown function (ip: 0x7fb28b07cbf0)
unknown function (ip: 0x7fb28aeec6ae)
_ZN7QWidget5eventEP6QEvent at /lib64/libQt5Widgets.so.5 (unknown line)
_ZN6QFrame5eventEP6QEvent at /lib64/libQt5Widgets.so.5 (unknown line)
_ZN13QGraphicsView13viewportEventEP6QEvent at /lib64/libQt5Widgets.so.5 (unknown line)
unknown function (ip: 0x7fb28aeef35a)
_ZN23QCoreApplicationPrivate29sendThroughObjectEventFiltersEP7QObjectP6QEvent at /lib64/libQt5Core.so.5 (unknown line)
_ZN19QApplicationPrivate13notify_helperEP7QObjectP6QEvent at /lib64/libQt5Widgets.so.5 (unknown line)
_ZN12QApplication6notifyEP7QObjectP6QEvent at /lib64/libQt5Widgets.so.5 (unknown line)
unknown function (ip: 0x7fb28ae1723d)
_ZN16QCoreApplication15notifyInternal2EP7QObjectP6QEvent at /lib64/libQt5Core.so.5 (unknown line)
_ZN19QApplicationPrivate14sendMouseEventEP7QWidgetP11QMouseEventS1_S1_PS1_R8QPointerIS0_Eb at /lib64/libQt5Widgets.so.5 (unknown line)
unknown function (ip: 0x7fb28a7b99f7)
unknown function (ip: 0x7fb28a7bc59d)
_ZN19QApplicationPrivate13notify_helperEP7QObjectP6QEvent at /lib64/libQt5Widgets.so.5 (unknown line)
_ZN12QApplication6notifyEP7QObjectP6QEvent at /lib64/libQt5Widgets.so.5 (unknown line)
unknown function (ip: 0x7fb28ae1723d)
_ZN16QCoreApplication15notifyInternal2EP7QObjectP6QEvent at /lib64/libQt5Core.so.5 (unknown line)
_ZN22QGuiApplicationPrivate17processMouseEventEPN29QWindowSystemInterfacePrivate10MouseEventE at /lib64/libQt5Gui.so.5 (unknown line)
_ZN22QGuiApplicationPrivate24processWindowSystemEventEPN29QWindowSystemInterfacePrivate17WindowSystemEventE at /lib64/libQt5Gui.so.5 (unknown line)
_ZN22QWindowSystemInterface22sendWindowSystemEventsE6QFlagsIN10QEventLoop17ProcessEventsFlagEE at /lib64/libQt5Gui.so.5 (unknown line)
unknown function (ip: 0x7fb2e3065d9e)
_ZN16QCoreApplication13processEventsE6QFlagsIN10QEventLoop17ProcessEventsFlagEEi at /lib64/libQt5Core.so.5 (unknown line)
unknown function (ip: 0x7fb28b3a48e7)
PyCFunction_Call at /usr/lib64/libpython3.7m.so.1.0 (unknown line)
macro expansion at /home/glynch/.julia/dev/PyCall/src/exception.jl:81 [inlined]
__pycall! at /home/glynch/.julia/dev/PyCall/src/pyfncall.jl:44
_pycall! at /home/glynch/.julia/dev/PyCall/src/pyfncall.jl:22
#pycall#89 at /home/glynch/.julia/dev/PyCall/src/pyfncall.jl:11 [inlined]
pycall at /home/glynch/.julia/dev/PyCall/src/pyfncall.jl:83 [inlined]
#3 at /home/glynch/.julia/dev/PyCall/src/gui.jl:153
#435 at ./event.jl:475
jl_fptr_trampoline at /buildworker/worker/package_linux64/build/src/gf.c:1831
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2184
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1537 [inlined]
start_task at /buildworker/worker/package_linux64/build/src/task.c:268
unknown function (ip: 0xffffffffffffffff)
Allocations: 153495583 (Pool: 153445452; Big: 50131); GC: 376
[1]    5614 abort (core dumped)  julia1
```

galenlynch:
I have added a back-trace to the description of exceptions coming from Julia
that are subsequently wrapped in Python exceptions.

This makes it much easier to find the source of errors in Julia code being
called from Python functions.

Closes #607

### DIFF
--- a/src/callback.jl
+++ b/src/callback.jl
@@ -39,7 +39,7 @@ function _pyjlwrap_call(f, args_::PyPtr, kw_::PyPtr)
 
         return pyreturn(ret)
     catch e
-        pyraise(e)
+        pyraise(e, catch_backtrace())
     finally
         args.o = PyPtr_NULL # don't decref
     end
--- a/src/callback.jl
+++ b/src/callback.jl
@@ -39,7 +39,7 @@ function _pyjlwrap_call(f, args_::PyPtr, kw_::PyPtr)
 
         return pyreturn(ret)
     catch e
-        pyraise(e, catch_backtrace())
+        @pyraise e
     finally
         args.o = PyPtr_NULL # don't decref
     end
--- a/src/exception.jl
+++ b/src/exception.jl
@@ -123,16 +123,21 @@ function pyexc_initialize()
     pyexc[PyIOError] = @pyglobalobjptr :PyExc_IOError
 end
 
+_showerror_string(io::IO, e, ::Nothing) = showerror(io, e)
+_showerror_string(io::IO, e, bt) = showerror(io, e, bt)
+
+# bt argument defaults to nothing, to delay dispatching on the presence of a
+# backtrace until after the try-catch block
 """
     showerror_string(e) :: String
 
 Convert output of `showerror` to a `String`.  Since this function may
 be called via Python C-API, it tries to not throw at all cost.
 """
-function showerror_string(e::T) where {T}
+function showerror_string(e::T, bt = nothing) where {T}
     try
         io = IOBuffer()
-        showerror(io, e, catch_backtrace())
+        _showerror_string(io, e, bt)
         return String(take!(io))
     catch
         try
@@ -163,14 +168,15 @@ function showerror_string(e::T) where {T}
     end
 end
 
-function pyraise(e)
+function pyraise(e, bt = nothing)
     eT = typeof(e)
     pyeT = haskey(pyexc::Dict, eT) ? pyexc[eT] : pyexc[Exception]
     ccall((@pysym :PyErr_SetString), Cvoid, (PyPtr, Cstring),
-          pyeT, string("Julia exception: ", showerror_string(e)))
+          pyeT, string("Julia exception: ", showerror_string(e, bt)))
 end
 
-function pyraise(e::PyError)
+# Second argument allows for backtraces passed to `pyraise` to be ignored.
+function pyraise(e::PyError, ::Vector = [])
     ccall((@pysym :PyErr_Restore), Cvoid, (PyPtr, PyPtr, PyPtr),
           e.T, e.val, e.traceback)
     e.T.o = e.val.o = e.traceback.o = C_NULL # refs were stolen
--- a/src/exception.jl
+++ b/src/exception.jl
@@ -181,3 +181,13 @@ function pyraise(e::PyError, ::Vector = [])
           e.T, e.val, e.traceback)
     e.T.o = e.val.o = e.traceback.o = C_NULL # refs were stolen
 end
+
+"""
+    @pyraise e
+
+Throw the exception `e` to Python, attaching a backtrace.  This macro should only be
+used in a `catch` block so that `catch_backtrace()` is valid.
+"""
+macro pyraise(e)
+    :(pyraise($(esc(e)), catch_backtrace()))
+end
--- a/src/exception.jl
+++ b/src/exception.jl
@@ -132,7 +132,7 @@ be called via Python C-API, it tries to not throw at all cost.
 function showerror_string(e::T) where {T}
     try
         io = IOBuffer()
-        showerror(io, e)
+        showerror(io, e, catch_backtrace())
         return String(take!(io))
     catch
         try
--- a/src/io.jl
+++ b/src/io.jl
@@ -8,13 +8,13 @@
 # IOBase methods:
 
 # IO objects should raise IOError for unsupported operations or failed IO
-function ioraise(e)
+function ioraise(e, bt = nothing)
     if isa(e, MethodError) || isa(e, ErrorException)
         ccall((@pysym :PyErr_SetString), Cvoid, (PyPtr, Cstring),
               (pyexc::Dict)[PyIOError],
-              string("Julia exception: ", e))
+              showerror_string(e, bt))
     else
-        pyraise(e)
+        pyraise(e, bt)
     end
 end
 
@@ -22,7 +22,7 @@ macro with_ioraise(expr)
     :(try
         $(esc(expr))
       catch e
-        ioraise(e)
+        ioraise(e, catch_backtrace())
       end)
 end
 
--- a/src/pyiterator.jl
+++ b/src/pyiterator.jl
@@ -134,7 +134,7 @@ const jlWrapIteratorType = PyTypeObject()
                 return pyreturn(item)
             end
         catch e
-            pyraise(e, catch_backtrace())
+            @pyraise e
         end
         return PyPtr_NULL
     end
@@ -149,7 +149,7 @@ else
                 return pyreturn(item)
             end
         catch e
-            pyraise(e, catch_backtrace())
+            @pyraise e
         end
         return PyPtr_NULL
     end
@@ -162,7 +162,7 @@ function pyjlwrap_getiter(self_::PyPtr)
         self = unsafe_pyjlwrap_to_objref(self_)
         return pystealref!(jlwrap_iterator(self))
     catch e
-        pyraise(e, catch_backtrace())
+        @pyraise e
     end
     return PyPtr_NULL
 end
--- a/src/pyiterator.jl
+++ b/src/pyiterator.jl
@@ -134,7 +134,7 @@ const jlWrapIteratorType = PyTypeObject()
                 return pyreturn(item)
             end
         catch e
-            pyraise(e)
+            pyraise(e, catch_backtrace())
         end
         return PyPtr_NULL
     end
@@ -149,7 +149,7 @@ else
                 return pyreturn(item)
             end
         catch e
-            pyraise(e)
+            pyraise(e, catch_backtrace())
         end
         return PyPtr_NULL
     end
@@ -162,7 +162,7 @@ function pyjlwrap_getiter(self_::PyPtr)
         self = unsafe_pyjlwrap_to_objref(self_)
         return pystealref!(jlwrap_iterator(self))
     catch e
-        pyraise(e)
+        pyraise(e, catch_backtrace())
     end
     return PyPtr_NULL
 end
--- a/src/pytype.jl
+++ b/src/pytype.jl
@@ -348,7 +348,7 @@ function pyjlwrap_repr(o::PyPtr)
         return pyreturn(o != C_NULL ? string("<PyCall.jlwrap ",unsafe_pyjlwrap_to_objref(o),">")
                         : "<PyCall.jlwrap NULL>")
     catch e
-        pyraise(e, catch_backtrace())
+        @pyraise e
         return PyPtr_NULL
     end
 end
@@ -395,7 +395,7 @@ function pyjlwrap_getattr(self_::PyPtr, attr__::PyPtr)
             end
         end
     catch e
-        pyraise(e, catch_backtrace())
+        @pyraise e
     finally
         attr_.o = PyPtr_NULL # don't decref
     end
--- a/src/pytype.jl
+++ b/src/pytype.jl
@@ -348,7 +348,7 @@ function pyjlwrap_repr(o::PyPtr)
         return pyreturn(o != C_NULL ? string("<PyCall.jlwrap ",unsafe_pyjlwrap_to_objref(o),">")
                         : "<PyCall.jlwrap NULL>")
     catch e
-        pyraise(e)
+        pyraise(e, catch_backtrace())
         return PyPtr_NULL
     end
 end
@@ -395,7 +395,7 @@ function pyjlwrap_getattr(self_::PyPtr, attr__::PyPtr)
             end
         end
     catch e
-        pyraise(e)
+        pyraise(e, catch_backtrace())
     finally
         attr_.o = PyPtr_NULL # don't decref
     end


tkf:
Julia's backtrace can be quite long sometimes so I'm not sure constructing a string all the time is a good idea.  I think the right approach is to wrap exception lazily and then generate the exception message in Python side on demand:

```python
class PyCallException(Exception):
    def __init__(self, exception, backtrace):
        self.exception = exception
        self.backtrace = backtrace

   def __str__(self):
       return f"Julia exception: {showerror_string(self.exception)}"

   def print_backtrace(self):
       print(sprint(showerror, self.exception, self.backtrace))


class PyCallKeyError(PyCallException, KeyError):
   ...
```

where `print_backtrace` is a non-standard method you need to call manually (e.g., via a debugger or logger).

One thing I'm not sure about my approach is the lifetime of the backtrace returned by `catch_backtrace()`.  It's just a `Vector{Union{Ptr{Nothing}, InterpreterIP}}` and I wonder how long those pointers are valid.

Another much more simple-minded approach is to just put 

```julia
@debug "Throwing error in Python" exception=(e, bt)
```

in `pyraise(e)`.  This way, you can just `ENV["JULIA_DEBUG"] = "PyCall"` to look at the stacktraces.

tkf:
I didn't mean that there is a length limit for the string generated from Python exceptions.  I'm not sure if such limit exists (other than RAM).

> generating a string would a small performance hit compared to exception handling

When the exception includes objects with a lot of type parameters (like when using ForwardDiff.jl), I find that stacktrace to be huge and very slow to print everything.  I don't think you'd want to pay that cost simply because a Julia function throws (say) a `KeyError`.  Furthermore, many Python APIs use exception type as the part of the interface (at least more than Julia, I think).  So, the exception message may be ignored and generating it can be totally wasteful.  That's why I don't think converting stacktrace to a string all the time is optimal.

> I find the error messages so terse that they're seldom helpful.

Yeah, I want to improve it, too.

> Is there some other mechanism to include the Julia stack in the python exception?

AFIK `Exception.__str__` is the only generic way to do this.  Actually, including stacktrace in it (like you do in this PR) is not so bad since it means that it is likely that Python is printing its own stacktrace.

galenlynch:
Thanks for thinking through this in so much detail!

If I understand you correctly, your solution requires making a whole bunch of new exception classes in python and putting the Julia stacktrace in the `__str__` attribute. While this may be the right way of solving the problem, I think it's beyond the scope of what I need and I won't implement it in this PR.

Since your suggestion of adding `@debug "Throwing error in Python" exception=(e, bt)` in `pyraise(e)` would satisfy my need of determining where exceptions in Julia are coming from when I am developing some code, then I would be happy to implement it and see how it works as part of this PR. If not I'm happy to close the PR.

src/exception.jl
```diff
@@ -132,7 +132,7 @@ be called via Python C-API, it tries to not throw at all cost.
 function showerror_string(e::T) where {T}
     try
         io = IOBuffer()
-        showerror(io, e)
+        showerror(io, e, catch_backtrace())
```
stevengj: I'm worried that this isn't valid  We should really only call `catch_backtrace()` in a `catch` block, otherwise its value may be unpredictable.

Can the caller to `showerror_string` pass in the backtrace if it is available?

galenlynch:
@stevengj I tried to address your concern, though the solution I came up with isn't the prettiest.

I have changed `pyraise` to accept an optional backtrace, which will be passed to `showerror_string`. Since `pyraise` only calls `showerror_string` for non-`PyError` exceptions, the `pyraise` method accepting `PyError`s has a dummy argument so backtraces can be passed to either method, though they will ultimately be ignored if they are paired with a `PyError` exception.

I also added `catch_backtrace()` to all calls of `pyraise` inside a try-catch block in PyCall.jl.

I think these changes result in more verbose and informative error messages from exceptions originating in Julia code, while not calling `catch_backtrace` in inappropriate contexts. Additionally, if users want to supply a backtrace to `pyraise`, they will now be able to do so.

I understand that these changes obscure the flow of logic, so I would totally understand if you don't like them. However, my goal has already been achieved with the original commit, which I agree is not a general solution as `pyraise` is exported and can be called outside of a try-catch block where there may not be a current exception. As such, I might just tweak PyCall's code locally to show a stack trace when I need one, instead of developing a more generally robust solution than the one I've added here.

src/callback.jl
```diff
@@ -39,7 +39,7 @@ function _pyjlwrap_call(f, args_::PyPtr, kw_::PyPtr)
 
         return pyreturn(ret)
     catch e
-        pyraise(e)
+        pyraise(e, catch_backtrace())
```
stevengj: Might be cleaner to have a `@pyraise e` macro that inserted the `catch_backtrace()` call:
```jl
macro pyraise(e)
    :(pyraise($(esc(e)), catch_backtrace())
end
```

galenlynch: Done

stevengj:
Overall, this looks good to me.

tkf:
@stevengj I'm a bit worried about the overhead for this https://github.com/JuliaPy/PyCall.jl/pull/608#issuecomment-434915444, since exceptions in Python are considered "lightweight" and rendering backtrace can be time-consuming in Julia sometimes.  But we can test if that's the case in the wild by just releasing it... :smiling_imp:

galenlynch:
Given the choice between fast exceptions that are hard to locate, and slower exceptions that tell you exactly where the problem is, I would personally choose the latter. For my use case, hunting down the exception would have been nearly impossible without a backtrace. However, @tkf is right that converting the backtrace into a string takes time:

```julia
julia> bt = backtrace()
16-element Array{Union{Ptr{Nothing}, InterpreterIP},1}:
 Ptr{Nothing} @0x00007f0af9dcb176                                                                                                                                                                         
 Ptr{Nothing} @0x00007f0af9f34800                                                                                                                                                                         
 Ptr{Nothing} @0x00007f0af9f34529                                                                                                                                                                         
 Ptr{Nothing} @0x00007f0af9f34eac                                                                                                                                                                         
 Ptr{Nothing} @0x00007f0af9f3562f                                                                                                                                                                         
 Base.InterpreterIP(CodeInfo(
 1 ─ %1 = backtrace()                                                       │
 │        bt = %1                                                           │
 └──      return %1                                                         │
), 0x0000000000000000)
 Ptr{Nothing} @0x00007f0af9f360ad                                                                                                                                                                         
 Ptr{Nothing} @0x00007f0af9e012fc                                                                                                                                                                         
 Ptr{Nothing} @0x00007f0af9ddc590                                                                                                                                                                         
 Ptr{Nothing} @0x00007f0ae16ea992                                                                                                                                                                         
 Ptr{Nothing} @0x00007f0af9dcb176                                                                                                                                                                         
 Ptr{Nothing} @0x00007f0ae1d59ac3                                                                                                                                                                         
 Ptr{Nothing} @0x00007f0ae1d59d24                                                                                                                                                                         
 Ptr{Nothing} @0x00007f0af9dcb176                                                                                                                                                                         
 Ptr{Nothing} @0x00007f0af9de6afb                                                                                                                                                                         
 Ptr{Nothing} @0x0000000000000000

julia> @benchmark PyCall.showerror_string(ErrorException("foo"))
BenchmarkTools.Trial: 
  memory estimate:  336 bytes
  allocs estimate:  8
  --------------
  minimum time:     339.516 ns (0.00% GC)
  median time:      345.489 ns (0.00% GC)
  mean time:        399.394 ns (6.74% GC)
  maximum time:     11.166 μs (93.70% GC)
  --------------
  samples:          10000
  evals/sample:     219

julia> @benchmark PyCall.showerror_string(ErrorException("foo"), bt)
BenchmarkTools.Trial: 
  memory estimate:  23.47 KiB
  allocs estimate:  519
  --------------
  minimum time:     9.299 ms (0.00% GC)
  median time:      9.652 ms (0.00% GC)
  mean time:        9.734 ms (0.00% GC)
  maximum time:     14.666 ms (0.00% GC)
  --------------
  samples:          514
  evals/sample:     1
```
I don't have a sense for how common it is for python code to rely on exception handling for non-exceptional situations, which would be necessary to determine the real world performance impact of this PR.

Merging this PR does not preclude a future lazy implementation, like @tkf suggested. Right now, passing a back trace to `pyraise` is optional, and calls to `pyraise` within PyCall.jl supply that back trace. If and when a better implementation is made, then it would be trivial to change the default behaviour within PyCall.jl.

tkf:
I was rather talking about the case that printing backtrace takes a few tens of seconds.  I remember it happens in case there are a lot of type parameters.  I couldn't find the example now but here is a rather contrived example that takes about a half second:

```julia
using ForwardDiff

nthderiv(f, n, x = 0.0) =
    if n == 1
        ForwardDiff.derivative(f, x)
    else
        ForwardDiff.derivative(z -> nthderiv(f, n - 1, z), x)
    end

for n in 1:100
    f = Symbol("f$n")
    g = Symbol("f$(n-1)")
    @eval @noinline $f(x, bang) = $g(x, bang) + 1
end
function f0(n, bang)
    bang && error("bang!")
    return 0
end

ex = try
    nthderiv(x -> f100(x, true), 10)
catch e
    e, catch_backtrace()
end
```

```julia
julia> sizeof(@time sprint(showerror, ex...)) / 2^20
  0.465707 seconds (3.96 M allocations: 225.571 MiB, 6.76% gc time)
11.887667655944824
```


But this generates a 11 MiB string so maybe there is no wonder why it is slow.  Actually, this is much faster than printing it to the terminal (via `tmux`) which takes more than a minute (at which time I terminated the process).  So maybe the slowness I remembered was due to the bottleneck in the terminal.  (Although I remember that sometimes it looked like Julia was JIT'ing something)

tkf:
FYI Python manual says:

> EAFP
> Easier to ask for forgiveness than permission. This **common Python coding style** assumes the existence of valid keys or attributes and catches exceptions if the assumption proves false.
> --- https://docs.python.org/3/glossary.html#term-eafp

stevengj:
It seems like showerror should be fixed in Base to truncate the backtrace if it is longer than a couple kB. 

I’m not so concerned about performance here at the moment, though. I doubt that people will go to the trouble of calling Julia code from Python except for fairly expensive functions—not for cheap functions like dictionary lookups where the cost of constructing a backtrace is a concern.

PR #613, opened 2018-11-08
RFC: Add an API to safely add conda packages

tkf:
If a package installed via `Conda.add` is not available for installed
Python version, `conda` may downgrade Python:
https://discourse.julialang.org/t/help-understanding-pycall-related-travis-failure/15957

This brakes PyCall because it stores the path to libpython etc.  This
patch provides a workaround to the issue by adding an API that wraps
`Conda.add` and re-build PyCall if Python version is changed during
installation of conda packages.

Proposed API:

> ```julia
> conda_add(packages::AbstractVector{<: AbstractString})
> ```
>
> Install conda `pakcages` _if_ PyCall is configured to use Conda.jl; otherwise this function does nothing.  It is intended to be used from `deps/build.jl` of the packages depending on PyCall, for safely installing Python packages via conda.  More precisely, this function detects if Python executable is re-installed during installing `packages` and then re-build PyCall if it happens.

fix https://github.com/JuliaPy/Conda.jl/issues/127


src/PyCall.jl
```diff
@@ -686,6 +687,33 @@ function pyimport_conda(modulename::AbstractString, condapkg::AbstractString,
...
+"""
+function conda_add(packages::AbstractVector{<: AbstractString})
+    if !conda
+        @debug "PyCall is not configured to use Conda.jl. Skip installing $packages."
+        return
+    end
+
+    before = Conda.version("python")
+    for pkg in packages
+        Conda.add(pkg)
```
tkf: TODO: switch to `Conda.add(packages)` once the Conda.jl version containing  https://github.com/JuliaPy/Conda.jl/pull/130 is released.

PR #615, opened 2018-11-15
Fix Segfault during finalization

JobJob:
Incorporates change introduced in #603


JobJob:
@tkf does this make sense? I didn't really pay attention, just copied what was done for `pydecref_(::Union{PyPtr, PyObject})` in #603 and it fixed the segfault locally on 0.7

tkf:
Awesome!  Thanks a lot for tracking down the issue super quickly.  Yes, it looks like the right thing to do.  Once CPython runtime is finalized, we can't call any CPython API.

I guess this the only function that is touched in #487 and could be called via `finalizer`?

stevengj:
LGTM; AppVeyor failure is an unrelated network glitch.

PR #619, opened 2018-11-22
Julia 0.6+ is supported and 0.7/1.0 really works

PallHaraldsson:
It seems to me 0.7 works according to: https://travis-ci.org/JuliaPy/PyCall.jl/jobs/456635014

The badge for 0.7, showing 1.16.1 failing isn't really helpful? Nor really the other for newer 1.17.1, but still old version, rather than showing for newest?

I'm not sure how to fix the badges. Is it better to say 0.5 is supported? Including the README on purpose in case you use an older PyCall.jl?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,6 @@
 
 [![PyCall](http://pkg.julialang.org/badges/PyCall_0.5.svg)](http://pkg.julialang.org/?pkg=PyCall&ver=0.5)
 [![PyCall](http://pkg.julialang.org/badges/PyCall_0.6.svg)](http://pkg.julialang.org/?pkg=PyCall&ver=0.6)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,7 +6,6 @@
 
 [![PyCall](http://pkg.julialang.org/badges/PyCall_0.5.svg)](http://pkg.julialang.org/?pkg=PyCall&ver=0.5)
```
stevengj: Did you mean to delete this line and keep the 0.7 line?

PallHaraldsson: Actually, I meant to drop the line showing 0.7 failing, as it gives the impression 0.7 and 1.0 is failing... while I believe it to be false. I'm not sure if how can be modified to show 0.7 (and/or at least 1.0 succeeding).

I don't care too much about showing 0.5 not failing, it's your call to drop that one or not. My understanding is that if you up the requirements, it shows the last Julia version working for 0.5, and I can go either way, keep it or not.

stevengj:
Obviated by #628

PR #621, opened 2018-12-06
Use native sizes by default when parsing buffer format string

JobJob:
Ahh oops. So, apparently you're supposed to use native sizes by default when parsing the buffer protocol format strings. I did the opposite in #487.

Current master
```
using PyCall
py"""
import numpy as np
"""
nparr = py"np.array([1,1,1,1])"o
nparr["dtype"]  # PyObject dtype('int64')
convert(PyAny, nparr)
```
returns `Int32[1,0,1,0]` \*
It should just be `Int[1,1,1,1]`

This actually caused me a bug in some relatively large distributed code, that I found very hard to track down. So karmically, pretty interesting.

At least it didn't make it to a release.

\* The [1, 0, 1, 0] is also in part due to a bug in f_contiguous for 1d arrays when `stride != sizeof(T)` - I will post a PR with a fix for that next week, I need to make some other changes to get the tests right for that. This is the minor positive to come out of this.

### DIFF
--- a/src/pybuffer.jl
+++ b/src/pybuffer.jl
@@ -239,7 +239,7 @@ function array_format(pybuf::PyBuffer)
     if length(fmt_str) > 1
         type_start_idx = 2
         if fmt_str[1] == '@' || fmt_str[1] == '^'
-            # defaults to native_byteorder: true, is_standard_size: false
+            # defaults to native_byteorder: true, use_native_sizes: true
         elseif fmt_str[1] == '<'
             native_byteorder = ENDIAN_BOM == 0x04030201
             use_native_sizes = false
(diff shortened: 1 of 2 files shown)


stevengj:
What is the AppVeyor failure?

JobJob:
Just an fyi: with the [appveyor failures before](https://ci.appveyor.com/project/StevenGJohnson/pycall-jl-nu3aa/builds/20824400), it seems like the conda x64 environments might not have actually been 64-bit builds of Python. The tests were failing because the dtype of a `np.array([1,2,...,10])` in python was int32, not int64, but in julia Int was Int64 on those environments.

PR #631, opened 2019-01-23
avoid stealing refs in pyreturn

### LINKED ISSUE #567: Reference to `PyPlot.Figure` is still stolen
The same bug I reported in #551 still happens with `PyPlot.Figure` (and probably with `SymPy.jl`, `Pandas.jl`, etc.):

```julia
julia> using PyCall, PyPlot

julia> fig = gcf()
Figure(PyObject <Figure size 666x1002 with 1 Axes>)

julia> py"""
       ns = {}
       def set(name):
           ns[name] = $include_string($Main, name)
       """

julia> py"set"("fig")

julia> fig
Figure(PyObject NULL)
```

The fix to #551 in #553 was

https://github.com/JuliaPy/PyCall.jl/blob/0f685cf02b6214ffc40b0aba7ab657749d587f5a/src/PyCall.jl#L143-L144

but `PyPlot` "smuggles" the refcount to Python via `pyreturn(x::Any)` path since [`PyObject(f::Figure)` just unwraps the internal `PyObject`](https://github.com/JuliaPy/PyPlot.jl/blob/daadae458ecb696e29c45636d91a9e7b7eddc246/src/PyPlot.jl#L56):

```julia
PyObject(f::Figure) = f.o
```

I wonder what's the best approach.  It would be bad if all `JuliaPy/*.jl` packages have to care about refcounts...  Maybe define an `AbstractPyObject` type and let people define "wrappers" by inheriting from it?

```julia
# --- PyCall.jl
abstract type AbstractPyObject end
pyreturn(x::AbstractPyObject) = pyincref_(PyObject(x))

# --- PyPlot.jl
mutable struct Figure <: AbstractPyObject
    o::PyObject
end
PyObject(f::Figure) = f.o
```

It still feels a bit dangerous that `PyObject` may create a new reference or returning existing one.  Maybe add an interface method (say) `pyunwrap` to be overloaded for `AbstractPyObject` subtypes which mixes-in a safer `PyObject` constructor?

```julia
# --- PyCall.jl
function pyunwrap end
abstract type AbstractPyObject end

PyObject(x::AbstractPyObject) = pyincref(pyunwrap(x).o)

pyreturn(x::AbstractPyObject) = pyincref_(pyunwrap(x))

# --- PyPlot.jl
mutable struct Figure <: AbstractPyObject
    o::PyObject
end
pyunwrap(f::Figure) = f.o  # define pyunwrap instead of PyObject
```

This sounds right since `invoke(pyreturn, Tuple{Any}, f)` where `f::Figure` still works (I think).

stevengj:
Fixes #567, cc @tkf.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -79,7 +79,7 @@ PyPtr(o::PyObject) = getfield(o, :o)
 """
     ≛(x, y)
 
-`PyPtr` based comparison of `x` and `y`, which can be of type `PyObject` or `PyPtr`. 
+`PyPtr` based comparison of `x` and `y`, which can be of type `PyObject` or `PyPtr`.
 """
 ≛(o1::Union{PyObject,PyPtr}, o2::Union{PyObject,PyPtr}) = PyPtr(o1) == PyPtr(o2)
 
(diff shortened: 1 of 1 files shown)


stevengj:
CI failures are unrelated glitches.

PR #632, opened 2019-01-23
deprecate o[integer] so that we can transition to 0-based access

stevengj:
Now that `o.foo` mirrors Python, we also want `o[foo]` to mirror Python.  Unfortunately, PyCall currently makes `o[integer]` in Julia act like `o[integer-1]` in Python, to emulate 1-based access.   This PR deprecates 1-based indexing of `PyObject` so that in PyCall 2.0 we can switch to 0-based indexing.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -79,7 +79,7 @@ PyPtr(o::PyObject) = getfield(o, :o)
 """
     ≛(x, y)
 
-`PyPtr` based comparison of `x` and `y`, which can be of type `PyObject` or `PyPtr`. 
+`PyPtr` based comparison of `x` and `y`, which can be of type `PyObject` or `PyPtr`.
 """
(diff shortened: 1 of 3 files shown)


crstnbr:
Naively, to have to mentally switch between 0 and 1-based indexing depending on whether I'm accessing a Julia or Python object seems like a big inconvenience. Maybe I misunderstand though. Could you elaborate why you want to switch to 0-based indexing? Isn't consistent 1-based indexing just as good - and more user friendly?

stevengj:
Because o[foo] will soon be equivalent to o[foo] in Python for all foo, not just integers. Imagine a Python dictionary, for example. Once we do that, it will be weird to modify foo only if foo is an integer. 

Auto-conversion will still translate a Python list to a 1-based Julia array (or PyVector). But if you have a raw PyObject then I think it should mimic Python.

PR #634, opened 2019-01-24
RFC: Warn top-level pyimport during pre-compilation

tkf:
This PR adds a warning from `pyimport` when a user tried to import Python modules at precompile time at top-level scope.  Example output:

```
┌ Error: Python module `sys` imported at top-level scope in /home/takafumi/.julia/dev/PyPlot/src/plot3d.jl:4.
│
│ This is very likely to be a miss-use of PyCall API.  Importing Python
│ modules _MUST_ be done in `__init__` function.  Please read:
│ https://github.com/JuliaPy/PyCall.jl#using-pycall-from-julia-modules
│
│ If importing Python code at import time cannot be avoided, use
│ `disable_warn_pyimport_at_toplevel` to disable this warning within a
│ restricted code block.
└ @ PyCall ~/.julia/dev/PyCall/src/PyCall.jl:506
```

This solution probably sounds like too "pythonic" because it needs to look at `stacktrace()` and dynamically determine the scope.  But `jl_generating_output` is the only overhead at run-time so I guess it's fine?  Other concern is that I'm probably using a lot of non-public interface.  But if it makes sense to do this in PyCall maybe we can request the public APIs to implement this PR?

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -472,11 +472,72 @@ function pyimport_e(name::AbstractString)
 end
 
 """
+    called_from_toplevel() :: Bool
+
+Check if this function is called at module top-level scope than via
+`__init__`.
+"""
+function called_from_toplevel()
+    for frame in stacktrace()
+        frame.func === Symbol("top-level scope") && return true
+        frame.func === :__init__ && return false
+    end
+    return true
+end
+
+const _disable_warn_pyimport_at_toplevel = Ref(false)
+
+function warn_pyimport_at_toplevel(name)
+    ccall(:jl_generating_output, Cint, ()) == 0 && return
+    _disable_warn_pyimport_at_toplevel[] && return
+    called_from_toplevel() || return
+    frames = stacktrace()
+    if length(frames) >= 3
+        # frames[1]: this function
+        # frames[2]: pyimport
+        # frames[3]: user's code
+        file = string(frames[3].file)
+        line = frames[3].line
+        location = " in $file:$line"
+    else
+        # something is terribly wrong, but try not fail
+        location = ""
+    end
+    @error """
+Python module `$name` imported at top-level scope$location.
+
+This is very likely to be a miss-use of PyCall API.  Importing Python
+modules _MUST_ be done in `__init__` function.  Please read:
+https://github.com/JuliaPy/PyCall.jl#using-pycall-from-julia-modules
+
+If importing Python code at import time cannot be avoided, use
+`disable_warn_pyimport_at_toplevel` to disable this warning within a
+restricted code block.
+"""
+end
+
+"""
+    disable_warn_pyimport_at_toplevel(f)
+
+Run function `f` while disabling warning about top-level Python
+imports.
+"""
+function disable_warn_pyimport_at_toplevel(f)
+    try
+        _disable_warn_pyimport_at_toplevel[] = true
+        f()
+    finally
+        _disable_warn_pyimport_at_toplevel[] = false
+    end
+end
+
+"""
     pyimport(s::AbstractString)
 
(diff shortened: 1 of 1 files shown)


src/PyCall.jl
```diff
@@ -471,12 +471,73 @@ function pyimport_e(name::AbstractString)
...
+        # frames[3]: user's code
+        file = string(frames[3].file)
+        line = frames[3].line
+        location = " in $file:$line"
+    else
+        # something is terribly wrong, but try not fail
+        location = ""
+    end
+    @error """
+Python module `$name` imported at top-level scope$location.
```
tkf: I just realized I can do what `Base.depwarn` does here and pass `_file` and `_line` to `@logmsg`. I'll do it if this entire patch makes sense.

src/PyCall.jl
```diff
@@ -471,12 +471,73 @@ function pyimport_e(name::AbstractString)
...
+"""
+end
+
+"""
+    disable_warn_pyimport_at_toplevel(f)
+
+Run function `f` while disabling warning about top-level Python
+imports.
+"""
+function disable_warn_pyimport_at_toplevel(f)
```
tkf: It makes sense to warn from `py"..."` etc. So this function should be renamed. Maybe `disable_toplevel_warning`.

stevengj:
While this is attractive in principle, it's perfectly legitimate to call `pyimport` in top-level code as long as the usage is transient.  Conversely, you would have the same problem for `const foo = PyObject(0)` in top-level code, or any code that tries to store a `PyObject` in the precompiled image — not just `pyimport`.

The right thing would be if we can issue a warning or error when *any* `PyObject` is serialized into the precompiled image.  I'm not sure if this is available?

(Ideally, we could just define `Base.precompile_isunsafe(::PyObject) = true` and it would issue a warning.  This would be a nice feature to add to Base.)

stevengj:
Hmm, I guess `Base.precompile_isunsafe(o::PyObject) = ispynull(o)` wouldn't actually work, because precompilation runs `__init__` and actually does serialize the `PyPtr` — it's just that well-written code would ignore the serialized value by overwriting it in `__init__`.

Alternatively, if we could somehow get precompilation to use our pickle serialization, then we could maybe allow precompilation with Python objects.

tkf:
> it's perfectly legitimate to call `pyimport` in top-level code as long as the usage is transient.

Yeah, so that's why I included `disable_warn_pyimport_at_toplevel` which probably should be called `disable_toplevel_warning`.  It can be used like

```julia
const pypi = PyCall.disable_toplevel_warning() do
    pyimport("math").pi
end
```

> Conversely, you would have the same problem for `const foo = PyObject(0)` in top-level code, or any code that tries to store a `PyObject` in the precompiled image — not just `pyimport`.

I was thinking to do this also with `py"..."` as well.  But there are probably too many entry points to cover.  At the same time, I think `pyimport` and `py"..."` cover most of the newbie traps.

> if we could somehow get precompilation to use our pickle serialization

Wow.  It would be great if we can support this.

stevengj:
@vtjnash, is there some way to hook into how precompilation serializes an object?  I looked at `src/dump.c` and it seems quite difficult to alter from Julia code, but maybe I'm missing something.

tkf:
> > if we could somehow get precompilation to use our pickle serialization
> 
> Wow. It would be great if we can support this.

Actually, let me take it back.  A Python package may be updated after a Julia package using it is precompiled.  It then could easily break pickled object when some relevant internal of the updated Python package is changed (e.g., class property is renamed).  So, I think it's much safer to disallow it.

vtjnash:
@stevengj no, there are generally not any hooks in there

PR #635, opened 2019-01-24
Make ?obj.property work for obj::PyObject

tkf:
Before:

```julia
julia> os = pyimport("os")
PyObject <module 'os' from '/home/takafumi/.virtualenvs/julia/lib/python3.7/os.py'>

help?> os.path.join
ERROR: MethodError: no method matching Base.Docs.Binding(::PyObject, ::Symbol)
...
```

After:

```julia
help?> os.path.join
Join two or more pathname components, inserting '/' as needed.
    If any component is an absolute path, all previous path components
    will be discarded.  An empty last part will result in a path that
    ends with a separator.
```

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -896,6 +896,7 @@ end
 # Expose Python docstrings to the Julia doc system
 
 Docs.getdoc(o::PyObject) = Text(convert(String, o."__doc__"))
+Docs.Binding(o::PyObject, s::Symbol) = getproperty(o, s)
 
 #########################################################################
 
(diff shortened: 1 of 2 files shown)


stevengj:
It seems like this is a bug in `Base.Docs`, to be honest.  Shouldn't it have a fallback method:

```jl
Binding(x, s::Symbol) = getproperty(x, s)
```

?

Also, does help work for `os."path"."join"` as well?  Do we need a `Binding` method for `AbstractString`?

tkf:
Right. I should probably send a PR to `Base`.  Supporting `?os."path"."join"` also makes sense.

PR #637, opened 2019-01-24
warn if Julia & numpy are linked with MKL

stevengj:
Alert users to #433.


PallHaraldsson:
>Warn

Did you mean to link to somewhere else than 433? It or same number at JuliaLang, seem not relevant.

Issue #348, opened 2017-01-12
return value of last expression from multi-line py"""..."""


stevengj:
Currenty,
```jl
py"""
....Python code
"""
```
returns nothing, because that is what [`Py_CompileString` does for `Py_file_input`](https://docs.python.org/2/c-api/veryhigh.html).

It would be nice to, instead, return the result of the last expression evaluated, similar to what IPython does when evaluating multi-line input.   As discussed in #166, IPython does this by parsing the AST and evaluating one expression at a time, and in principle we could do the same thing.

NHDaly:
This might be a silly question, but could we simply call out to IPython itself, to share this implementation, rather than reimplementing it ourselves?

Or can we not rely on `ipython` being installed? Thanks!

NHDaly:
(Sorry to re-post this, I noticed that I never got an email notification about my comment, I think because of the Github outage last week. Reposting in case no one saw it. 😊)

------


This might be a silly question, but could we simply call out to IPython itself, to share this implementation, rather than reimplementing it ourselves?

Or can we not rely on `ipython` being installed? Thanks!

PR #638, opened 2019-01-25
Use separate Python namespace in py"" for each Julia module

### LINKED ISSUE #599: Better Python eval/exec?
I think `py"..."` has two major limitations:

### No private namespace

As `py"..."` does not support passing namespace dictionaries, you are forced to use `__main__` all the time.  It means that `py"..."` cannot be used in any libraries without interfering with other libraries.  There are things which are easier to do in Python syntax.  Some special syntaxes like `async`/`await` are not supported in PyCall.  But if `eval`/`exec` can't be done in a private namespace, `py"..."` is not an option.

Of course, you can create a custom Python module and `pyimport` it after tweaking `sys.path`.  But this is rather cumbersome to prepare if you just want to define a few Python functions.

### Multi-line eval

Python expression can span multiple lines.  Following code does not return the array as one would expect.

```julia
py"[
    1,
    2,
    3,
]"
```

I suppose current Julia can't distinguish `py"..."` and `py"""..."""` since `_mstr` is gone?

### Ideas

Here are a few ideas to fix the problem:

1. Extend `py_str` by adding options?

    - `py"..."eval(globals, locals)`
    - `py"..."exec(globals, locals)`
    - The idea is to return a callable from `py"..."eval` and `py"..."exec`.

2. Bring back `pyeval` and add `pyexec`?

    - `pyeval(source, [globals, [locals]]; returntype=PyAny)`
    - `pyexec(source, [globals, [locals]])`

3. Add macros `@pyeval` and `@pyexec`?

    - Actually, I suggest to implement `pyeval` and `pyexec` as macors, to extract file name and line number and pass them to Python.  This way, Python stack trace can show the code in Julia source.
    - Also, interpolation mechanism can be used here.
    - Implementation: https://gist.github.com/tkf/93a6d76f5ca2aa15bae2b1b72b162db9

I also suggest to add a `@pyload` macro (or `pyload` function) which is equivalent to:

```julia
locals = PyDict{Symbol, PyObject}()
@pyexec("...", locals)
return locals
```

This way, defining helper Python functions can easily be done:

```julia
ns = @pyload("""
def f():
    ...

def g():
    ...
""")
ns[:f]()
ns[:g]()
```

A better option probably is to return a `types.SimpleNamespace` so that you can do `ns.f` instead of `ns[:f]` once `getproperty` is implemented.

tkf:
fixes #599 (the other part of #599 is already tracked by #348)

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,11 +304,11 @@ dictionary.
 #### PyTextIO
 
 Julia `IO` streams are converted into Python objects implementing the
-[RawIOBase](http://docs.python.org/2/library/io.html#io.RawIOBase)
+[RawIOBase](http://docs.python.org/3/library/io.html#io.RawIOBase)
 interface, so they can be used for binary I/O in Python.  However,
 some Python code (notably unpickling) expects a stream implementing
 the
-[TextIOBase](http://docs.python.org/2/library/io.html#io.TextIOBase)
+[TextIOBase](http://docs.python.org/3/library/io.html#io.TextIOBase)
 interface, which differs from RawIOBase mainly in that `read` an
 `readall` functions return strings rather than byte arrays.  If you
 need to pass an `IO` stream as a text-IO object, just call
@@ -353,7 +353,7 @@ and also by providing more type information to the Julia compiler.
   `pycall(function,returntype,args...)`.
 
 * `py"..."` evaluates `"..."` as a Python code, equivalent to
-  Python's [`eval`](https://docs.python.org/2/library/functions.html#eval) function, and returns the result
+  Python's [`eval`](https://docs.python.org/3/library/functions.html#eval) function, and returns the result
   converted to `PyAny`.  Alternatively, `py"..."o` returns the raw `PyObject`
   (which can then be manually converted if desired).   You can interpolate
   Julia variables and other expressions into the Python code with `$`,
@@ -369,7 +369,9 @@ and also by providing more type information to the Julia compiler.
   If you use `py"""..."""` to pass a *multi-line* string, the string can
   contain arbitrary Python code (not just a single expression) to be evaluated,
   but the return value is `nothing`; this is useful e.g. to define pure-Python
-  functions, and is equivalent to Python's [`exec`](https://docs.python.org/2/reference/simple_stmts.html#exec) statement.  (If you define a Python global `g` in a multiline `py"""..."""`
+  functions, and is equivalent to Python's
+  [`exec`](https://docs.python.org/3/library/functions.html#exec) function.
+  (If you define a Python global `g` in a multiline `py"""..."""`
   string, you can retrieve it in Julia by subsequently evaluating `py"g"`.)
 
   When `py"..."` is used inside a Julia module, it uses a Python namespace
--- a/README.md
+++ b/README.md
@@ -141,6 +141,50 @@ conversion, use `f::PyObject`). Similarly, if the context manager returns a type
 which is automatically converted to a Julia type, you will have override this
 via `@pywith EXPR::PyObject ...`.
 
+If you are already familiar with Python, it perhaps is easier to use
+`py"..."` and `py"""..."""` which are equivalent to Python's
+[`eval`](https://docs.python.org/3/library/functions.html#eval) and
+[`exec`](https://docs.python.org/3/library/functions.html#exec),
+respectively:
+
+```julia
+py"""
+import numpy as np
+
+def sinpi(x):
+    return np.sin(np.pi * x)
+"""
+py"sinpi"(1)
+```
+
+When creating a Julia module, it is a useful pattern to define Python
+functions or classes in Julia's `__init__` and then use it in Julia
+function with `py"..."`.
+
+```julia
+module MyModule
+
+using PyCall
+
+function __init__()
+    py"""
+    import numpy as np
+
+    def one(x):
+        return np.sin(x) ** 2 + np.cos(x) ** 2
+    """
+end
+
+two(x) = py"one"(x) + py"one"(x)
+
+end
+```
+
+Note that Python code in `py"..."` of above example is evaluated in a
+Python namespace dedicated to `MyModule`.  Thus, Python function `one`
+cannot be accessed outside `MyModule`.
+
+
 ## Troubleshooting
 
 Here are solutions to some common problems:
@@ -308,7 +352,7 @@ and also by providing more type information to the Julia compiler.
   `@pycall function(args...)::returntype` into
   `pycall(function,returntype,args...)`.
 
-* `py"..."` evaluates `"..."` as a Python string, equivalent to
+* `py"..."` evaluates `"..."` as a Python code, equivalent to
   Python's [`eval`](https://docs.python.org/2/library/functions.html#eval) function, and returns the result
   converted to `PyAny`.  Alternatively, `py"..."o` returns the raw `PyObject`
   (which can then be manually converted if desired).   You can interpolate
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -308,8 +352,8 @@ and also by providing more type information to the Julia compiler.
   `@pycall function(args...)::returntype` into
   `pycall(function,returntype,args...)`.
 
-* `py"..."` evaluates `"..."` as a Python string, equivalent to
-  Python's [`eval`](https://docs.python.org/2/library/functions.html#eval) function, and returns the result
+* `py"..."` evaluates `"..."` as a Python code, equivalent to
```
stevengj: as Python code

(no "a")

README.md
```diff
@@ -325,9 +369,18 @@ and also by providing more type information to the Julia compiler.
...
+  [`exec`](https://docs.python.org/3/library/functions.html#exec) function.
+  (If you define a Python global `g` in a multiline `py"""..."""`
   string, you can retrieve it in Julia by subsequently evaluating `py"g"`.)
 
+  When `py"..."` is used inside a Julia module, it uses a Python namespace
+  dedicated to this Julia module.  Thus, you can define Python function
+  using `py"""...."""` in your module without worrying about name clash
+  with other Python code.  Note that Python functions _must_ be defined in
+  `__init__`.  Side-effect in Python occurred at top-level Julia scope
+  cannot be used at run-time.
```
stevengj: for precompiled modules

src/pyeval.jl
```diff
@@ -2,13 +2,25 @@ const Py_single_input = 256  # from Python.h
...
+    get!(_namespaces, m) do
+        if m === Main
+            return PyDict{String,PyObject,true}(pyincref(@pycheckn ccall((@pysym :PyModule_GetDict), PyPtr, (PyPtr,), pyimport("__main__"))))
+        else
+            ns = PyDict{String,PyObject}()
+            # In Python 2, it looks like `__builtin__` (w/o 's') must
+            # exist at module namespace.  See also:
+            # http://mail.python.org/pipermail/python-dev/2001-April/014068.html
+            # https://github.com/ipython/ipython/blob/512d47340c09d184e20811ca46aaa2f862bcbafe/IPython/core/interactiveshell.py#L1295-L1299
+            PyObject(ns).setdefault("__builtin__", builtin)
```
stevengj: Should this be only `if pyversion < v"3"` then?

src/pyeval.jl
```diff
@@ -2,13 +2,25 @@ const Py_single_input = 256  # from Python.h
...
+            PyObject(ns).setdefault("__builtins__", builtin)
+            return ns
+        end
     end
-    return _maindict
-end
+
+maindict() = pynamespace(Main)
```
stevengj: Do we still need this function?

src/pyeval.jl
tkf: I didn't put the conditional because current IPython (which supports only Python 3) is still have this hack https://github.com/ipython/ipython/blob/512d47340c09d184e20811ca46aaa2f862bcbafe/IPython/core/interactiveshell.py#L1295-L1299 But that was probably a historical effect in IPython codebase.

src/pyeval.jl
```diff
@@ -2,29 +2,38 @@ const Py_single_input = 256  # from Python.h
...
 # internal function evaluate a python string, returning PyObject, given
 # Python dictionaries of global and local variables to use in the expression,
 # and a current "file name" to use for stack traces
-function pyeval_(s::AbstractString, globals=maindict(), locals=maindict(), input_type=Py_eval_input, fname="PyCall")
-    sb = String(s) # use temp var to prevent gc before we are done with o
-    sigatomic_begin()
-    try
+function pyeval_(s::AbstractString, globals=pynamespace(Main), locals=pynamespace(Main),
+                 input_type=Py_eval_input, fname="PyCall")
+    GC.@preserve s globals locals fname disable_sigint() do
```
tkf: Don't we need `GC.@preserve` here?

stevengj: Why?  `ccall` automatically roots its arguments.

tkf: Ah, I see. I was caught up in the line `sb = String(s) # use temp var to prevent gc before we are done with o` in the old code.  But just removing this line and passing `s` to `ccall` is enough?  I suppose `ccall` would root the argument and then `cconvert` it?

tkf: Wait, no.  Converted string has to be preserved for _two_ `ccall`s.  So I guess it should be

```julia
    sb = String(s)
    GC.@preserve sb disable_sigint() do
        o = PyObject(@pycheckn ccall((@pysym :Py_CompileString), PyPtr,
                                     (Cstring, Cstring, Cint),
                                     sb, fname, input_type))
        return PyObject(@pycheckn ccall((@pysym :PyEval_EvalCode),
                                         PyPtr, (PyPtr, PyPtr, PyPtr),
                                         o, globals, locals))
    end
```

src/pyeval.jl
```diff
@@ -2,29 +2,39 @@ const Py_single_input = 256  # from Python.h
...
 # internal function evaluate a python string, returning PyObject, given
 # Python dictionaries of global and local variables to use in the expression,
 # and a current "file name" to use for stack traces
-function pyeval_(s::AbstractString, globals=maindict(), locals=maindict(), input_type=Py_eval_input, fname="PyCall")
+function pyeval_(s::AbstractString, globals=pynamespace(Main), locals=pynamespace(Main),
+                 input_type=Py_eval_input, fname="PyCall")
     sb = String(s) # use temp var to prevent gc before we are done with o
-    sigatomic_begin()
-    try
+    GC.@preserve sb disable_sigint() do
```
tkf: Implemented https://github.com/JuliaPy/PyCall.jl/pull/638#discussion_r251158240

src/pyeval.jl
stevengj: No, `sb` does not have to be preserved after the first `ccall`.

`Py_CompileString` creates its own parsed representation of the input string, and does not need the caller's string to persist after the call.

stevengj: Let me dig into when that comment was added; maybe it is an error.

stevengj: Oh, you added the comment yourself in 4d577452c80f39fdcf4a38725bebff0bf252790c … I think this is an error.

tkf: Actually, it looks like tokentized object seems to be destroyed before `Py_CompileString` returns.  But I'm not sure the if `char *str` is passed to the code object returned by `Py_CompileString`.  The [documentation](https://docs.python.org/3/c-api/veryhigh.html#c.Py_CompileString) does not mention anything.

src/pyeval.jl
```diff
@@ -2,24 +2,36 @@ const Py_single_input = 256  # from Python.h
...
+            return PyDict{String,PyObject,true}(pyincref(@pycheckn ccall((@pysym :PyModule_GetDict), PyPtr, (PyPtr,), pyimport("__main__"))))
+        else
+            ns = PyDict{String,PyObject}()
+            # In Python 2, it looks like `__builtin__` (w/o 's') must
+            # exist at module namespace.  See also:
+            # http://mail.python.org/pipermail/python-dev/2001-April/014068.html
+            # https://github.com/ipython/ipython/blob/512d47340c09d184e20811ca46aaa2f862bcbafe/IPython/core/interactiveshell.py#L1295-L1299
+            if pyversion < v"3"
+                PyObject(ns).setdefault("__builtin__", builtin)
+                PyObject(ns).setdefault("__builtins__", builtin)
```
stevengj: We should still have `__builtins__` in Python 3,  think?

stevengj: See https://docs.python.org/3/library/builtins.html

tkf: Good point, I changed it.

src/pyeval.jl
```diff
@@ -15,13 +15,13 @@ pynamespace(m::Module) =
             # http://mail.python.org/pipermail/python-dev/2001-April/014068.html
             # https://github.com/ipython/ipython/blob/512d47340c09d184e20811ca46aaa2f862bcbafe/IPython/core/interactiveshell.py#L1295-L1299
             if pyversion < v"3"
-                PyObject(ns).setdefault("__builtin__", builtin)
+                ns["__builtin__"] = builtin
```
tkf: Quoting my commit message in 5ceca2ab95ef43c97d0f5a1fe7d0db60c07218c0:

> Unlike IPython, we are sure that we have an empty dictionary.  So, using setindex! here is cleaner.

stevengj:
Thanks!

Issue #502, opened 2018-05-15
Problem importing python file from current directory


dompas:
I have tried the solution suggested on #48 , but it is not working. I am using Julia 0.6.2 on Windows with Python 2.7 and the program is 

```
using PyCall
unshift!(PyVector(pyimport("sys")["path"]),"")
@pyimport Leitura_Sistema_REDS as ls
```

and I get the error message 

```
ERROR: LoadError: PyError (ccall(@pysym(:PyImport_ImportModule), PyPtr, (Cstring,), name)

The Python package Leitura_Sistema_REDS could not be found by pyimport. Usually this means
that you did not install Leitura_Sistema_REDS in the Python version being used by PyCall.
```

YP-Ye:
I get the same error. unshift!(PyVector(pyimport("sys")["path"]),"") does not work.

stevengj:
What is your current directory?  i.e. what is `pwd()`?  Are you sure that you are in the directory with the module that you want?

YP-Ye:
Sorry, I found that my problem is caused by a python module named gurobipy. The problem is solved after reinstalling the module.

tkf:
Adding an empty string to `sys.path` is not really a good practice.  I think the right approach is `pushfirst!(PyVector(pyimport("sys")["path"]), @__DIR__)`.  This way, running Julia script works in any directory.  This is close to what Python interpreter does when executing a script (it does _not_ add `""` to `sys.path`).  Here is an example:

```console
$ cat script.jl
using PyCall
pushfirst!(PyVector(pyimport("sys")["path"]), @__DIR__)
pyimport("mymodule")[:main]()

$ cat mymodule.py
def main():
    import sys
    print(*sys.path, sep="\n")


if __name__ == "__main__":
    main()

$ cd /

$ julia /PATH/TO/DIR/script.jl
/PATH/TO/DIR
/usr/lib/python37.zip
/usr/lib/python3.7
/usr/lib/python3.7/lib-dynload
/usr/lib/python3.7/site-packages

$ python /PATH/TO/DIR/mymodule.py  # prints the same paths
```

Of course, `pyimport` has to be done in `__init__` if you are creating a Julia module.

(We probably should add this to README)

MaxandreJ:
[...]

PR #641, opened 2019-01-31
Fix syntax for adding current directory to search path

dangirsh:
pyimport("sys")("path") should be pyimport("sys")[:path]


dangirsh:
Fixes  #502

dangirsh:
Ignore this. I was mistaken.

PR #642, opened 2019-02-03
Do not throw KeyError from property access

tkf:
We don't use `setindex!` etc. anymore so it's a bit strange to throw `KeyError`.  Julia has no AttributeError equivalent so throwing PyError seems to be an obvious straightforward solution.

However, maybe we should wait until PyCall 2.0 as we still only deprecated indexing API and haven't completely removed it.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -292,11 +292,7 @@ function getproperty(o::PyObject, s::AbstractString)
     if ispynull(o)
         throw(ArgumentError("ref of NULL PyObject"))
     end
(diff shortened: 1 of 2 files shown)


stevengj:
I don't know, `KeyError` still seems like the best available exception here.  The attributes are a kind of dictionary after all, even if the syntax is different.

tkf:
> The attributes are a kind of dictionary after all, even if the syntax is different.

Python also has `__slots__` so I think this is an implementation detail.

But I don't have very strong opinion about it at the moment.  Relying on exception type is not super Julian approach so it probably does not matter much.

PR #649, opened 2019-02-19
Add PySide2 support to pygui

cherichy:
Add the PySide2 for :qt and :qt_pyside to pygui()

### DIFF
--- a/src/gui.jl
+++ b/src/gui.jl
@@ -26,9 +26,9 @@ pygui_works(gui::Symbol) = gui == :default ||
      (gui == :tk && pyexists(tkinter_name())) ||
      ((gui == :qt5 || gui == :qt_pyqt5) && pyexists("PyQt5")) ||
      (gui == :qt_pyqt4 && pyexists("PyQt4")) ||
-     (gui == :qt_pyside && pyexists("PySide")) ||
+     (gui == :qt_pyside && (pyexists("PySide") || pyexists("PySide2"))) ||
      (gui == :qt4 && (pyexists("PyQt4") || pyexists("PySide"))) ||
-     (gui == :qt && (pyexists("PyQt5") || pyexists("PyQt4") || pyexists("PySide"))))
+     (gui == :qt && (pyexists("PyQt5") || pyexists("PyQt4") || pyexists("PySide") || pyexists("PySide2"))))
 
 # get or set the default GUI; doesn't affect running GUI
 """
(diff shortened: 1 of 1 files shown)


src/gui.jl
```diff
@@ -232,7 +232,11 @@ function pygui_start(gui::Symbol=pygui(), sec::Real=50e-3)
         elseif gui == :qt_pyqt4
             eventloops[gui] = qt_eventloop("PyQt4", sec)
         elseif gui == :qt_pyside
-            eventloops[gui] = qt_eventloop("PySide", sec)
+            try
+                eventloops[gui] = qt_eventloop("PySide", sec)
+            catch
+                eventloops[gui] = qt_eventloop("PySide2", sec)
```
tkf: Shouldn't this be in `elseif gui == :qt_pyside2` branch? Also, I think `try`-`catch` in a newly created `elseif gui == :qt5` branch makes sense (just to reflect what is already done for `:qt4`).

> Q: Why PySide2 and not just PySide?
> A: Since PySide was developed for Qt4, when the port was made to support Qt5, the name was changed to PySide2 to imply that it was a newer version.
> --- https://wiki.qt.io/Qt_for_Python

cherichy: Yeah, it should be in :qt5, but its name is PySide2, maybe users will consider :qt_pyside first. Anyway, anyone use qt backend would consider :qt as the first choice. I will revise it to :qt5 if it is preferred.

tkf: What I meant was something like

```julia
elseif gui == :qt_pyqt4
    eventloops[gui] = qt_eventloop("PyQt4", sec)
elseif gui == :qt_pyqt5
    eventloops[gui] = qt_eventloop("PyQt5", sec)
elseif gui == :qt_pyside
    eventloops[gui] = qt_eventloop("PySide", sec)
elseif gui == :qt_pyside2
    eventloops[gui] = qt_eventloop("PySide2", sec)
elseif gui == :qt4
    try
        eventloops[gui] = qt_eventloop("PyQt4", sec)
    catch
        eventloops[gui] = qt_eventloop("PySide", sec)
    end
elseif gui == :qt5
    try
        eventloops[gui] = qt_eventloop("PyQt5", sec)
    catch
        eventloops[gui] = qt_eventloop("PySide2", sec)
    end
```

But additionally falling back to `"PySide2"` in `:qt_pyside` branch also makes sense.  I don't have a strong opinion on this.  Maybe @stevengj has a better idea.

stevengj:
Thanks!

PR #651, opened 2019-02-25
Make PyCall.jl AOT-compilable

tkf:
This PR makes PyCall.jl compatible with PackageCompiler.jl and add some tests.

Quick usage:

```console
$ aot/compile.jl --color=yes
...

$ aot/julia.sh --compiled-modules=no --startup-file=no --banner=no
julia> @time using PyCall
  0.001537 seconds (309 allocations: 16.688 KiB)
```

Slightly more detailed docs in: [aot/README.md](https://github.com/tkf/PyCall.jl/blob/aot/aot/README.md)

### DIFF
--- a/aot/compile.jl
+++ b/aot/compile.jl
@@ -1,0 +1,17 @@
+using Pkg
+Pkg.activate(@__DIR__)
+Pkg.develop(PackageSpec(path=dirname(@__DIR__)))
+Pkg.activate()
+
+using PackageCompiler
+sysout, _curr_syso = compile_incremental(
+    joinpath(@__DIR__, "Project.toml"),
+    joinpath(@__DIR__, "precompile.jl"),
+)
+
+pysysout = joinpath(@__DIR__, basename(sysout))
+cp(sysout, pysysout, force=true)
+
+write(joinpath(@__DIR__, "_julia_path"), Base.julia_cmd().exec[1])
+
+@info "System image: $pysysout"
--- a/aot/compile.jl
+++ b/aot/compile.jl
@@ -1,3 +1,9 @@
+#!/bin/bash
+# -*- mode: julia -*-
+#=
+exec "${JULIA:-julia}" "$@" ${BASH_SOURCE[0]}
+=#
+
 using Pkg
 Pkg.activate(@__DIR__)
 Pkg.develop(PackageSpec(path=dirname(@__DIR__)))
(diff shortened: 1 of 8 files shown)


src/pyinit.jl
```diff
@@ -125,6 +126,17 @@ function Py_Finalize()
...
+    # Clear out global states.  This is required only for PyCall
+    # AOT-compiled into system image.
+    _finalized[] = false
+    empty!(_namespaces)
+    empty!(eventloops)
+    empty!(npy_api)
+    empty!(pycall_gc)
+    empty!(pyexc)
+    empty!(pytype_queries)
+    empty!(permanent_strings)
```
stevengj: Why is this required?

aot/compile.jl
```diff
@@ -0,0 +1,25 @@
...
+    joinpath(@__DIR__, "Project.toml"),
+    joinpath(@__DIR__, "precompile.jl"),
+)
+
+pysysout = joinpath(@__DIR__, basename(sysout))
+cp(sysout, pysysout, force=true)
+
+write(joinpath(@__DIR__, "_julia_path"), Base.julia_cmd().exec[1])
+
+@info "System image: $pysysout"
```
stevengj: This all seems like boilerplate — why does PyCall.jl need a custom `compile.jl` file?

src/pyinit.jl
tkf: The changes in https://github.com/JuliaPy/PyCall.jl/pull/651/commits/b0e131ac8c89b6247a436ae60f270e65991955a3 are the only ones required for making PyCall compilable with PackageCompiler.  The ones in https://github.com/JuliaPy/PyCall.jl/pull/651/commits/a9e9419a40192d3397d6d9f6b4c6a805af02b728 are not required but added for consistency and possible similar problems.  It seems that at least adding them does not break PyCall.

aot/compile.jl
tkf: Snoop in PackageCompiler does not work due to https://github.com/JuliaLang/julia/issues/31156 (a bug in Julia?).  There is another "latent" issue https://github.com/JuliaLang/PackageCompiler.jl/issues/192.  I think PackageCompiler is still rather a moving target so using lower-level API with a bit of boilerplate makes sense.  Another upside is that we can control exactly what functions are included in system image by tweaking code in `aot/precompile.jl` (although it is rather sparse and random ATM).

tkf:
Rebuilt after #671: https://travis-ci.org/JuliaPy/PyCall.jl/builds/511789604 (all green)

PR #664, opened 2019-03-12
Clarify readme on python exe path

musm:
I was struggling with  PyCall as it wasn't picking up my python installation since I thought the `PYTHON` env variable should point to the python **directory** as opposed the  python executable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ can change the Python version by setting the `PYTHON` environment variable
 to the path of the `python` (or `python3` etc.) executable and then re-running `Pkg.build("PyCall")`.
 In Julia:
 
-    ENV["PYTHON"] = "... path of the python program you want ..." # e.g. "C:\\Miniconda3\\python.exe"
+    ENV["PYTHON"] = "... path of the python executable ..." # e.g. "C:\\Miniconda3\\python.exe"
     Pkg.build("PyCall")
 
 Note also that you will need to re-run `Pkg.build("PyCall")` if your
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,7 +49,7 @@ can change the Python version by setting the `PYTHON` environment variable
 to the path of the `python` (or `python3` etc.) executable and then re-running `Pkg.build("PyCall")`.
 In Julia:
 
-    ENV["PYTHON"] = "... path of the python program you want ..."
+    ENV["PYTHON"] = "... path of the python executable ..." # e.g. "C:\\Miniconda3\\python.exe"
```
tkf: Including `"C:\\Miniconda3\\python.exe"` may confuse people who are using Conda.jl and may make them think they have to configure it manually.  Also, maybe it's better to have examples both for Windows and *nix.

So, how about

```suggestion
    ENV["PYTHON"] = "... path of the python executable ..."
    # ENV["PYTHON"] = raw"C:\Python37-x64\python.exe"  # example for Windows
    # ENV["PYTHON"] = "/usr/bin/python3.7"             # example for *nix
```

musm: Hmm, but Conda.jl decides to *download a complete python distribution* as a new to a new conda enc,  taking up massive amount of storage.....

tkf: I'm not suggesting for people to switch to Conda.jl.  In fact, my suggestion does not mention `conda` at all.  I'm trying to make it less confusing for people using Conda.jl.

musm: I see, sorry I misunderstood your reply. I guess nowadays most people I know use anaconda, which is why I didn't realize what you originally meant.

PR #668, opened 2019-03-19
various fixes to array conversions

stevengj:
Add bounds-checking, GC rooting, and `CartesianIndex` support in `PyArray`. Fixes conversion for arrays of pointers/pyptrs and 0-dimensional arrays.

Fixes JuliaPy/PyPlot.jl#431

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -369,6 +369,7 @@ function pyarray_dims(o::PyObject, forcelist=true)
         return () # too many non-List types can pretend to be sequences
     end
     len = ccall((@pysym :PySequence_Size), Int, (PyPtr,), o)
+    len < 0 && error("not a PySequence object")
     if len == 0
         return (0,)
     end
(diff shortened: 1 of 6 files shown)


stevengj:
So far I'm not able to reproduce the Travis failure on nightly locally.  @tkf, does it work for you?

tkf:
I checked this branch with 1.2.0-DEV.339 at-colon/00f257d603 and 1.2.0-DEV.521 b076d04bf4. I couldn't reproduce the failure in Travis.

stevengj:
I tried restarting Travis and the problem seems to have gone away!  So maybe it was a temporary bug that I was triggering in nightly that got fixed.

Now I'm seeing an unrelated failure:
```
The following packages have unmet dependencies:
 python3.4-venv : Depends: python-setuptools-whl but it is not installable
                  Depends: python-pip-whl but it is not installable
E: Unable to correct problems, you have held broken packages.
```
I will restart Travis again in case this is a temporary Travis hiccup, since I wasn't seeing this yesterday.

stevengj:
Yay, it's green now!

Issue #670, opened 2019-03-25
Suspicious segfault since #668


tkf:
We thought the segmentation faults in #668's CI were some temporary bug in Julia nightly.  However, I just noticed that it happens with Julia 1.0.3 as well:

```
signal (11): Segmentation fault
in expression starting at /home/travis/build/JuliaPy/PyCall.jl/test/runtests.jl:29
_ZNK4llvm4Type18isSizedDerivedTypeEPNS_15SmallPtrSetImplIPS0_EE at /home/travis/julia/bin/../lib/julia/libLLVM-6.0.so (unknown line)
_ZN4llvm12InstCombiner12visitBitCastERNS_11BitCastInstE at /home/travis/julia/bin/../lib/julia/libLLVM-6.0.so (unknown line)
...
```

--- https://travis-ci.org/JuliaPy/PyCall.jl/jobs/510151180#L804 from `v1.91.0` tag
(also https://travis-ci.org/JuliaPy/PyCall.jl/jobs/510149760#L793 from `master` branch)

stevengj:
Those errors went away in #668 after I added a `GC.@protect` in the `PyArray{PyPtr} -> Array{PyObject}` conversion, IIRC?

tkf:
The CI results I linked above are from the build for `v1.91.0` and `master`.  Both for the commit 33f07ebd6ad3d8b911a3447fadf48b08cceaa9b7.

stevengj:
I committed your workaround from JuliaLang/julia#31473

stevengj:
Tagged 1.91.1 with fix.

PR #671, opened 2019-03-26
avoid jl_value_ptr

stevengj:
A cleaner solution to JuliaLang/julia#31473 and #670.


yuyichao:
Technically this is dangerous since the argument to `unsafe_convert` is not guaranteed rooted by the caller here. The current implementation my or may not take advantage of this but it is technically allowed to assume that and give wrong result.. `pointer_from_objref` should be fine.

stevengj:
@yuyichao, the `jlWrapType` variable is a `const` precisely because we can't allow it to be garbage-collected as long as we are using the `tp_base` pointer.

So, for the same reason, I don't think rooting is an issue?

stevengj:
I guess in theory the `RefValue` object needs to be rooted too, so I would need a `GC.@preserve` here.

yuyichao:
Correct, it's the

>  argument to `unsafe_convert`

that's the issue.

PR #674, opened 2019-04-04
Add Project.toml

tkf:
We'll need a `Project.toml` for the next release https://discourse.julialang.org/t/22677

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,0 +1,25 @@
+name = "PyCall"
+uuid = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
+authors = ["Steven G. Johnson <stevenj@mit.edu>", "Yichao Yu <yyc1992@gmail.com>", "Takafumi Arakaki <aka.tkf@gmail.com>", "Simon Kornblith <simon@simonster.com>", "Páll Haraldsson <Pall.Haraldsson@gmail.com>", "Jon Malmaud <malmaud@gmail.com>", "Jake Bolewski <jakebolewski@gmail.com>", "Keno Fischer <keno@alumni.harvard.edu>", "Joel Mason <jobba1@hotmail.com>", "Jameson Nash <vtjnash@gmail.com>", "The JuliaPy development team"]
+version = "1.91.2-DEV"
+
+[deps]
+Conda = "8f4d0f93-b110-5947-807f-2305c1781a2d"
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
+Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
+VersionParsing = "81def892-9a0e-5fdd-b105-ffc91e053289"
+
+[compat]
+Conda = "1.0"
+MacroTools = "≥ 0.4"
+VersionParsing = "1.0"
+julia = "0.7, 1.0"
+
+[extras]
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[targets]
+test = ["Test"]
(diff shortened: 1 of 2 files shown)


Project.toml
```diff
@@ -0,0 +1,15 @@
+name = "PyCall"
+uuid = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
+authors = ["Steven G. Johnson <stevenj@mit.edu>", "Yichao Yu <yyc1992@gmail.com>", "Takafumi Arakaki <aka.tkf@gmail.com>", "Simon Kornblith <simon@simonster.com>", "Páll Haraldsson <Pall.Haraldsson@gmail.com>", "Jon Malmaud <malmaud@gmail.com>", "Jake Bolewski <jakebolewski@gmail.com>", "Keno Fischer <keno@alumni.harvard.edu>", "Joel Mason <jobba1@hotmail.com>", "Jameson Nash <vtjnash@gmail.com>", "The JuliaPy development team"]
```
tkf: Not sure what to do here. List of authors are taken from top 10 `git log --format='%aN <%aE>' | sort | uniq -c | sort -nr`. I then appended `"The JuliaPy development team"`.

Project.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "v1.91.2-DEV"
```
tkf: `version` is somewhat random, too

Project.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[deps]
+Conda = "8f4d0f93-b110-5947-807f-2305c1781a2d"
+MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
+VersionParsing = "81def892-9a0e-5fdd-b105-ffc91e053289"
+
+[compat]
+Conda = "1.0"
+MacroTools = "≥ 0.4"
```
tkf: It seems MacroTools 0.4 worked with Julia 0.7 nightly https://travis-ci.org/MikeInnes/MacroTools.jl/builds/288579177. So if we say `julia = "0.7, 1.0"` I think it makes sense to allow MacroTools 0.4.  Alternatively, we can do `MacroTools = "0.4, 0.5"` but it then makes this version of PyCall incompatible with MacroTools 0.6 and above.

Project.toml
```diff
@@ -0,0 +1,19 @@
+name = "PyCall"
+uuid = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
+authors = ["Steven G. Johnson <stevenj@mit.edu>", "Yichao Yu <yyc1992@gmail.com>", "Takafumi Arakaki <aka.tkf@gmail.com>", "Simon Kornblith <simon@simonster.com>", "Páll Haraldsson <Pall.Haraldsson@gmail.com>", "Jon Malmaud <malmaud@gmail.com>", "Jake Bolewski <jakebolewski@gmail.com>", "Keno Fischer <keno@alumni.harvard.edu>", "Joel Mason <jobba1@hotmail.com>", "Jameson Nash <vtjnash@gmail.com>", "The JuliaPy development team"]
+version = "1.91.2-DEV"
```
tkf: `version` is somewhat random, too

Project.toml
stevengj: that's fine, I guess

Project.toml
stevengj: I'm confused by the release process with a Project.toml.   Do we update the `version` and then tag a release?  In which case the `version` here should be the last released/tagged version?

tkf: I think this is a good quick summary https://discourse.julialang.org/t/22677/32

But first you need to install the Registrator app https://github.com/apps/registratortest/installations/new (see https://github.com/JuliaComputing/Registrator.jl)

> Do we update the `version` and then tag a release?

I think we update `version` in Project.toml and comment ```@JuliaRegistrator `register()```` in github.  [Adding a Git tag is optional](https://discourse.julialang.org/t/switching-package-registration-systems-on-monday/22677/20) but I think it's better to do that.

> In which case the `version` here should be the last released/tagged version?

Yeah, I think that's a valid option too.  I'll change it.

stevengj:
I guess we should delete REQUIRE now?

PR #677, opened 2019-04-11
Remove REQUIRE

tkf:
I forgot to do this in #674


tkf:
Actually, can we release 1.91.2 before doing this? I think we can release it as usual via METADATA after merging #678.

Fixing AOT support in PyCall is the blocker of https://github.com/JuliaPy/pyjulia/pull/256 at the moment.

PR #678, opened 2019-04-11
Bump to 1.91.2

tkf:
Can we release 1.91.2 before removing REQUIRE #677?


stevengj:
Is REQUIRE/attobot still working?  I'm confused about when the changeover is happening.

tkf:
I think so. https://github.com/JuliaLang/METADATA.jl/pulls has some PRs from some hours ago and https://github.com/JuliaRegistries/General/commits/master has `automatic sync with METADATA` commits at a few hours ago.

Stefan Karpinski says:

> Since the hardware failure has taken the sync script offline at a bad time and thereby delayed the switchover process, and since people seem to be actively contributing to the script that generates project files, it seems like it may be better to wait a bit before pulling the trigger on switching over to the new registration process, although I could be convinced otherwise.

--- https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/108

tkf:
Thanks!

PR #712, opened 2019-09-17
Add Julia 1.2 and 1.3 in CI; drop 0.7 and 1.1

tkf:


tkf:
Travis failure is due to AOT testing with Julia nightly.

(I think we just need to call `Random.__init___` somewhere to fix this. Though it may need to happen in PackageCompiler.jl.)

PR #713, opened 2019-09-22
Call Random.__init__() manually to fix AOT with Julia 1.3

tkf:
(Trying to) Fix the error from AOT compilation test:

```
ERROR: LoadError: LoadError: LoadError: AssertionError: 0 < tid <= length(THREAD_RNGs)
Stacktrace:
 [1] default_rng(::Int64) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Random/src/RNGs.jl:298
 [2] default_rng at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Random/src/RNGs.jl:296 [inlined]
 [3] shuffle!(::Array{Symbol,1}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Random/src/misc.jl:220
 [4] include at ./boot.jl:328 [inlined]
...
```

--- https://travis-ci.org/JuliaPy/PyCall.jl/jobs/585892454#L803-L808

PR #724, opened 2019-11-06
Test on Python 3.8

PallHaraldsson:


PallHaraldsson:
https://www.anaconda.com/distribution/
Latest 19.10 version seems stuck on Python 3.7. 

There's some new interesting stuff, possibly to exploit?
https://www.python.org/downloads/release/python-380/

PEP 590, Vectorcall: a fast calling protocol for CPython

https://www.infoworld.com/article/3400640/the-best-new-features-in-python-38.html
the walrus operator ( := )

You should be able to call all Python code, including with this new syntax, assuming you call 3.8. I think no support for it needed in PyCall.

PallHaraldsson:
One build was stalled (false alarm?) and the other, for AOT, I'm not sure.

We should probably test on 3.8 (or 3.8.1) and 3.9 is out in alpha, and it has intriguing GC-stuff: https://bugs.python.org/issue38379

PallHaraldsson:
#732

tkf:
> One build was stalled (false alarm?) and the other, for AOT, I'm not sure.

macOS has been very flaky... I think it might be better to switch GitHub Actions although it has another problem https://github.com/JuliaPy/pyjulia/pull/353#issuecomment-586063738

AOT is tracked by #713 so I suggest to ignore it here.

PallHaraldsson:
I'm not sure if "AppVeyor build failed" is outdated, or how to rerun.

PallHaraldsson:
For travis, the one failing is for nightly (I guess 1.5 back then) and python (while same for python3 is ok). Maybe a combination to just drop?

For appveyor, there are fails, maybe Python 3.8 not installed, retry?

PallHaraldsson:
How about skip testing some combinations, at least? E.g. the one that fails python[2] and nightly? Python3 works. Otherwise:

ERROR: SystemError: opening file "/tmp/jl_RjdHn1": No such file or directory

tkf:
Thanks for the PR. We test with Python 3.8 as of #825.

Issue #255, opened 2016-04-06
Do you have plan to add exec?


colinfang:
`eval` only evaluate expressions.

Is it possible to have `exec` as well so we can define python functions in julia.

stevengj:
I've thought about it.  It would be nice to define a custom string macro, too, so that you could do:
```
py"""
....Python code goes here...
"""
```
to evaluate Python code.

Ideally it would work like evaluating an IPython notebook cell -- i.e. it would return the result of the last expression evaluated.  Would love to see someone take a crack at this.

novatena2015:
With the below macro any python code can be executed from julia. But I don't know how to get the result of the last expression.
```
macro py_str(ex)
       pybuiltin("eval")(pybuiltin("compile")(ex,"<string>","exec"),Dict())
end
```

For example:

py"""
def myfunc(x):
    return x*x
x = 5
print(myfunc(x))
"""
will print 25 but returns Void to julia

stevengj:
@Carreau, how does IPython evaluate code cells so that it (a) gets the value of the last expression and (b) defines variables in the global scope?

Carreau:
We manually [get the ast](https://github.com/ipython/ipython/blob/54d60477b4c420bd6a1279a5569b5f1ffcd94d9f/IPython/core/interactiveshell.py#L2690) and [run each runnable bloc independently](https://github.com/ipython/ipython/blob/54d60477b4c420bd6a1279a5569b5f1ffcd94d9f/IPython/core/interactiveshell.py#L2817-L2828), Python provide natively an [`exec`](https://docs.python.org/3/library/functions.html#exec) that takes a namespace for both globals and the user one , we just the [use it](https://github.com/ipython/ipython/blob/54d60477b4c420bd6a1279a5569b5f1ffcd94d9f/IPython/core/interactiveshell.py#L2881). So technically we are not evaluating in the global scope, but in a subscope (which appears as the main one for the user but has a few [escape hatches](https://github.com/ipython/ipython/blob/54d60477b4c420bd6a1279a5569b5f1ffcd94d9f/IPython/core/interactiveshell.py#L1150-L1208)). Does that helps ?

stevengj:
Thanks @Carreau!

stevengj:
Multi-line input via `py"""...."""` is now supported, ala `exec`, so I'm closing this issue. 

It would be nice to be able to return the result of the last expression, similar to Python, and I opened a new issue #348 for that.

PR #726, opened 2019-11-11
Return the last value in a multiline py"" macrocall by evaling the last expression

### LINKED ISSUE #348: return value of last expression from multi-line py"""..."""

NHDaly:
Fixes #348.

Following the explanation given in https://github.com/JuliaPy/PyCall.jl/issues/255#issuecomment-212842832, I've gone ahead and changed the `py""` macro to do basically the same thing that IPython does:
1. Parse the string into a `Module` AST (by calling `compile()` with the `PyCF_ONLY_AST` flag, and _always_ asking for "exec").
2. Compile as `"exec"` and run all the nodes in the AST _except the last Expression_.
3. For the last Expression node, compile it with `"eval"` mode and run it to extract its result.

Note: This differs a bit from IPython, which always compiles the last node using the `"single"` mode, which _prints out_ the result. We want to use `"eval"` to _return_ the last result. But the `"single"` mode is more flexible, handling any type of node, and will only print the result `iff` the node was an Expression. Instead, since the `"eval"` mode _requires_ its input to be an expression, we only `eval` the last node `iff` if it is an Expression, otherwise we keep it with the `"exec"` nodes.

This abstracts away the weird python distinction between `exec` and `eval` so that callers don't need to know about it or think about it when building `py""` calls! :) This now lets you do nice things like single-line imports and definitions, and multiline expressions; and of course it lets you mix definitions and expressions like an IPython cell. :)

```julia
julia> py"def foo(): return 1"  # Used to be a syntax error

julia> py"""[
         1,
         2,
         3,
       ]"""   # Used to return nothing
3-element Array{Int64,1}:
 1
 2
 3

julia> foo = py"""
         def foo(x): return x+1
         foo"""  # Used to return nothing
PyObject <function foo at 0x7f8fc0a81488>

julia> foo(1)
2
```

:)

--------

TODO:
- [ ] Update the documentation to eliminate notes about dealing with this distinction and explain that `py""` blocks work exactly like IPython cells: they return the last expression if there is one.


NHDaly:
(sorry, i forgot to set the title on the first PR)

NHDaly:
Some implementation notes:
- To keep the old behavior of `py""` not setting local variables if only evaling an expression, I added a sort of wonky output argument to track whether any statements were `exec`'d. There might be a better way to do this, or I'm not sure if we want to keep this behavior anyway? It seems probably useful though.
    - One nice aspect of this is that it's a generalization of the old behavior: everything that was true only for expressions is still true only for expressions, it's just that expressions can now be multi-line, and vice-versa.

TODO:
- [ ] Update the documentation to eliminate notes about dealing with this distinction and explain that `py""` blocks work exactly like IPython cells: they return the last expression if there is one.

NHDaly:
- I had to switch to using they python `compile` builtin instead of the C `Py_CompileString` function, since the C API only accepts strings, but we need to recompile the AST nodes after dividing them into `exec` and `eval`, so I needed to (I think) call the `compile` builtin, which also accepts AST objects.

PR #753, opened 2020-03-09
Improving find libpython

### LINKED ISSUE #752: Python 3.8 with Julia 1.3
I couldn't build `PyCall` with Python 3.8.0 and Julia 1.3.0. This happens with PyCall v1.91.4
```
julia> Pkg.build("PyCall")
  Building Conda ─→ `~/.julia/packages/Conda/3rPhK/deps/build.log`
  Building PyCall → `~/.julia/packages/PyCall/zqDXB/deps/build.log`
┌ Error: Error building `PyCall`: 
│ ERROR: LoadError: Couldn't find libpython; check your PYTHON environment variable.
│ 
│ The python executable we tried was /home/darren/miniconda3/bin/python3.8 (= version 3.7).
│ Re-building with
│     ENV["PYCALL_DEBUG_BUILD"] = "yes"
│ may provide extra information for why it failed.
│ 
│ Stacktrace:
│  [1] error(::String) at ./error.jl:33
│  [2] #find_libpython#3(::typeof(Libdl.dlopen), ::typeof(find_libpython), ::String) at /home/darren/.julia/packages/PyCall/zqDXB/deps/buildutils.jl:72
│  [3] find_libpython(::String) at /home/darren/.julia/packages/PyCall/zqDXB/deps/buildutils.jl:41
│  [4] top-level scope at /home/darren/.julia/packages/PyCall/zqDXB/deps/build.jl:87
│  [5] include at ./boot.jl:328 [inlined]
│  [6] include_relative(::Module, ::String) at ./loading.jl:1105
│  [7] include(::Module, ::String) at ./Base.jl:31
│  [8] include(::String) at ./client.jl:424
│  [9] top-level scope at none:5
│ in expression starting at /home/darren/.julia/packages/PyCall/zqDXB/deps/build.jl:43
└ @ Pkg.Operations /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Pkg/src/backwards_compatible_isolation.jl:649
false
```
Can confirm the Python path is correct:
```
(base) darren@darren-GK5CN5Z-APEX-15:~$ /home/darren/miniconda3/bin/python3.8
Python 3.8.0 (default, Oct 24 2019, 15:04:43) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
```
Saw [this](https://github.com/JuliaPy/PyCall.jl/issues/732) issue, so PyCall seems to support Python 3.8 already? If so, why am I getting this? Is there any workaround?

Thanks!

darrencl:
Use `sysconfig.get_system_var("VERSION")` as last resort as there is a possibility that it is not updated.

```python
>>> sys.version_info
sys.version_info(major=3, minor=8, micro=0, releaselevel='final', serial=0)
>>> sysconfig.get_config_var("VERSION")
'3.7'
>>> sysconfig.get_python_version()
'3.8'
```
Closes #752

### DIFF
--- a/deps/find_libpython.py
+++ b/deps/find_libpython.py
@@ -182,8 +182,9 @@ def candidate_names(suffix=SHLIB_SUFFIX):
     sysdata = dict(
         v=sys.version_info,
         # VERSION is X.Y in Linux/macOS and XY in Windows:
-        VERSION=(sysconfig.get_config_var("VERSION") or
-                 "{v.major}.{v.minor}".format(v=sys.version_info)),
+        VERSION=(sysconfig.get_python_version() or
+                 "{v.major}.{v.minor}".format(v=sys.version_info) or
+                 sysconfig.get_config_var("VERSION")),
         ABIFLAGS=(sysconfig.get_config_var("ABIFLAGS") or
                   sysconfig.get_config_var("abiflags") or ""),
     )


stevengj:
Thanks, LGTM.

deps/find_libpython.py
```diff
@@ -182,8 +182,9 @@ def candidate_names(suffix=SHLIB_SUFFIX):
     sysdata = dict(
         v=sys.version_info,
         # VERSION is X.Y in Linux/macOS and XY in Windows:
-        VERSION=(sysconfig.get_config_var("VERSION") or
-                 "{v.major}.{v.minor}".format(v=sys.version_info)),
+        VERSION=(sysconfig.get_python_version() or
+                 "{v.major}.{v.minor}".format(v=sys.version_info) or
+                 sysconfig.get_config_var("VERSION")),
```
tkf: `sysconfig.get_config_var("VERSION")` never gets used since `"{v.major}.{v.minor}".format(v=sys.version_info)` is always truthy (non-empty string).

Also, reading https://docs.python.org/3/library/sysconfig.html#sysconfig.get_python_version, it looks like it's always X.Y while we need XY in Windows (guessing from the comment). I guess we just should have formatted this directly from `sys.version_info`.

darrencl: Hmm yeah it seems the comment says so, but how come the Windows CI passes?  Seems we tested this script in `test_build.jl`.

tkf: `find_libpython.py` contains a bunch of fallbacks to make it as robust as possible. So the CI in PyCall may not be using this code path.

Also, since `find_libpython.py` is mainly maintained in PyJulia, it has a bit more tests there. So, in general, if you want to improve `find_libpython.py`, it's better to do this against PyJulia repository. I don't think this case is caught by the CI, though.

PR #758, opened 2020-04-03
GC safety

stevengj:
Updates some old GC-unsafe code that pre-dates the modern `Ref`.


stevengj:
Unrelated travis failures.

Issue #645, opened 2019-02-07
detect ModuleNotFoundError in pyimport


stevengj:
Currently, we print `The Python package foo could not be found by pyimport` when *any* exception is thrown by `pyimport`.  We should only print this on a `ModuleNotFoundError`.

tkf:
An exception does not tell us if a module is missing or its dependency is missing (at least directly).  I think a better way to do it is to use [`importlib.util.find_spec`](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec) (or [`imp.find_module`](https://docs.python.org/3/library/imp.html#imp.find_module) for supporting Python < 3.4).

goretkin:
Just ran into this, and github helpfully pointed me to this issue when I was writing it. My details here won't help contribute the issue, but it might help someone searching later.
---

I'm running into an issue loading `matplotlib` on my machine (not the subject of this issue). Calling `pyimport` manually shows:

```
julia> PyPlot.PyCall.pyimport("matplotlib")
ERROR: PyError (PyImport_ImportModule

The Python package matplotlib could not be found by pyimport. Usually this means
that you did not install matplotlib in the Python version being used by PyCall.

PyCall is currently configured to use the Julia-specific Python distribution
installed by the Conda.jl package.  To install the matplotlib module, you can
use `pyimport_conda("matplotlib", PKG)`, where PKG is the Anaconda
package the contains the module matplotlib, or alternatively you can use the
Conda package directly (via `using Conda` followed by `Conda.add` etcetera).

Alternatively, if you want to use a different Python distribution on your
system, such as a system-wide Python (as opposed to the Julia-specific Python),
you can re-configure PyCall with that Python.   As explained in the PyCall
documentation, set ENV["PYTHON"] to the path/name of the python executable
you want to use, run Pkg.build("PyCall"), and re-launch Julia.

) <class 'ImportError'>
ImportError('dlopen(/Users/goretkin/.julia/conda/3/lib/python3.7/site-packages/matplotlib/ft2font.cpython-37m-darwin.so, 2): Library not loaded: @rpath/libpng16.16.dylib\n  Referenced from: /Users/goretkin/.julia/conda/3/lib/libfreetype.6.dylib\n  Reason: Incompatible library version: libfreetype.6.dylib requires version 51.0.0 or later, but libpng16.16.37.0.dylib provides version 16.0.0')
[...]

PR #760, opened 2020-04-07
Emit more accurate error message

goretkin:
Until issue #645 is resolved, this is a  measure to not mislead the user.


goretkin:
The CI failures appear unrelated to this change.

PR #764, opened 2020-04-11
set optimization level to 1 for faster load time

stevengj:
Similar to JuliaPlots/Plots.jl#2544 … this is a new feature of Julia 1.4.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -1,5 +1,9 @@
 module PyCall
 
+if isdefined(Base, :Experimental) && isdefined(Base.Experimental, Symbol("@optlevel"))
+    @eval Base.Experimental.@optlevel 1
+end
+
 using VersionParsing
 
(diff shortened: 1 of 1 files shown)


stevengj:
Unrelated Travis failures.

PallHaraldsson:
FYI: It's new in Julia 1.5 (and I don't know it to be backported to 1.4, or that we should), in case read your commit message.

And:

Startup can be much faster (not quite the 3.7x faster I've seen elsewhere):

```
$ ~/julia-1.6.0-DEV-8f512f3f6d/bin/julia -O0 --compile=min --startup-file=no
julia> @time using PyCall
  0.815904 seconds (495.25 k allocations: 30.702 MiB)

julia> @time using PyCall
  0.003530 seconds (3.00 k allocations: 191.172 KiB)

julia> @time using PyCall
  0.001573 seconds (608 allocations: 31.234 KiB)
```

PR #777, opened 2020-05-12
refactor py_str macro to allow passing custom global/locals

marius311:
This should have no change to functionality, but the new `@_py_str` macro can be used by other functions to pass custom globals/locals dicts into Python expression while still getting the benefit of `@py_str`'s variable interpolation. 

`@_py_str` is kept a macro rather than function because it makes downstream macro hygene a bit easier. 

I'm using this here https://github.com/JuliaPy/pyjulia/pulls/tbd, where this can maybe be discussed more.


src/pyeval.jl
```diff
@@ -202,32 +202,37 @@ pasted into the Python code.   This allows you to evaluate code
 where the code itself is generated by a Julia expression.
 """
 macro py_str(code, options...)
+    m = :(pynamespace($__module__))
+    esc(macroexpand(__module__, :(PyCall.@_py_str($m, $m, $code, $(options...)))))
+end
+
+macro _py_str(pyglobals, pylocals, code, options...)
```
stevengj, line 209: Should this be a macro or just a function?

stevengj:
Could you clarify why this should be a macro?

src/pyeval.jl
```diff
@@ -202,32 +202,37 @@ pasted into the Python code.   This allows you to evaluate code
 where the code itself is generated by a Julia expression.
 """
 macro py_str(code, options...)
+    m = :(pynamespace($__module__))
+    esc(macroexpand(__module__, :(PyCall.@_py_str($m, $m, $code, $(options...)))))
```
tkf, line 206: If it were a macro, I don't think we need to call `macroexpand` manually here?

```suggestion
    esc(:($PyCall.@_py_str($m, $m, $code, $(options...))))
```

src/pyeval.jl
tkf, line 209: If we are going to use a function for this, it should take a named tuple `(__module__=__module__, __source__=__source__)` as the first argument so that all the information available within a usual macro is available.  I think it's better to make it a single first argument rather than two arguments because future versions of `julia` may introduce some more information within a macro definition but it'd be bad to change the calling convention of this function.

src/pyeval.jl
```diff
@@ -202,32 +202,37 @@ pasted into the Python code.   This allows you to evaluate code
...
+end
+
+macro _py_str(pyglobals, pylocals, code, options...)
     T = length(options) == 1 && 'o' in options[1] ? PyObject : PyAny
     code, locals = interpolate_pycode(code)
     input_type = '\n' in code ? Py_file_input : Py_eval_input
     fname = make_fname(@__FILE__)
```
tkf, line 213: Should it use `__source__.file`, BTW?

src/pyeval.jl
tkf: Also, we should use `options` rather than `options...`.

src/pyeval.jl
marius311: Yea, I think so. Before I think it was broken since `@__FILE__` would always just return `pycall.jl`. I also think the the `make_fname` stuff is unnecessary, `__source__.file` never seems to return `nothing`. Will add this change.

marius311:
> Could you clarify why this should be a macro?

So it stems from the fact that inside `@prepare_for_pyjulia_call` we do a complicated recursive walk through the whole expression to transform it. Its really nice to just be able to stick an `esc` around the whole transformed thing at the end [here](https://github.com/JuliaPy/pyjulia/pull/381/files#diff-edf2cf9175b9b67d834d1b52cd15a1dfL81), otherwise you have to think (hard) about where to stick in `esc`'s throughout that walk. Maybe there's a simple way, but all my playing with it it got really complex. 

Given that you want to `esc` the whole thing, then whatever you expression you get from calling into `PyCall` can't have bare symbols like `PyObject` or `pyeval_` in it, they need to be turned into `PyCall.PyObject` and `PyCall.pyeval_`. You can do this by modifing things on the PyCall side, but it seems easier to just make `@_py_str` a macro and then hygene does this for you. That's basically my motivation. 

That said, I realize now all the `macroexpand(__module__, ...` stuff is unnecessary so I'll get rid of it in both PRs.

src/pyeval.jl
```diff
@@ -202,32 +198,37 @@ pasted into the Python code.   This allows you to evaluate code
 where the code itself is generated by a Julia expression.
 """
 macro py_str(code, options...)
+    m = :(pynamespace($__module__))
+    fname = String(__source__.file)
+    esc(:(PyCall.@_py_str($m, $m, $fname, $code, $(options...))))
```
stevengj, line 203: Why is this call to `esc` necessary, since there are `esc` calls within `@_py_str`? 

Indeed, the call to `esc` here seems like it might cause problems, because it prevents the names `pylocals`, `pyglobals`, and `ret` from being hygienized.

marius311: No, `pylocals` etc... get hygenized by `@_py_str`, the `esc` here makes it so the interpolated variables don't also get hygenized, e.g. with the current code: 

```julia
julia> @macroexpand py"1+$i"
quote
    #= /home/marius/src/pyjulia/dev/PyCall/src/pyeval.jl:229 =#
    (var"#18#pyglobals", var"#19#pylocals") = (PyCall.pynamespace(Main), PyCall.pynamespace(Main))
    #= /home/marius/src/pyjulia/dev/PyCall/src/pyeval.jl:230 =#
    begin
        var"#19#pylocals"["__julia_localvar_2_1"] = PyCall.PyObject(i)
    end
    #= /home/marius/src/pyjulia/dev/PyCall/src/pyeval.jl:231 =#
    var"#20#ret" = (PyAny)(PyCall.pyeval_((string)("1+__julia_localvar_2_1"), var"#18#pyglobals", var"#19#pylocals", 258, "none"))
    #= /home/marius/src/pyjulia/dev/PyCall/src/pyeval.jl:232 =#
    begin
        PyCall.delete!(var"#19#pylocals", "__julia_localvar_2_1")
    end
    #= /home/marius/src/pyjulia/dev/PyCall/src/pyeval.jl:233 =#
    var"#20#ret"
end
```

tkf, line 203:
```suggestion
    esc(:($PyCall.@_py_str($m, $m, $fname, $code, $(options...))))
```

I think you need this?

I think we better test `py""` with something like

```julia
baremodule TestPyStr
using PyCall: @py_str
pystr(x) = py"$$x"
end
@test TestPyStr.pystr("1+1") == 2
```

tkf: Maybe I should've asked this before but why [is `insert_pyevals` run at run-time](https://github.com/JuliaPy/pyjulia/pull/381/files#diff-edf2cf9175b9b67d834d1b52cd15a1dfL83)?  It looks like this doesn't depend on the run-time value of `globals` and `locals` (or maybe I'm missing something)?  If we expand everything at macro-expansion time, perhaps `esc` of `@py_str` here and `@prepare_for_pyjulia_call` there can be removed? Then maybe `_py_str` doesn't have to be a macro?

Issue #776, opened 2020-05-10
Cannot build PyCall using existing Conda env; 100s of dependency conflicts


z-q-y:
Hello, one of the modules in my project relies on calls to dozens of Python modules, so I thought I can use an existing Conda environment, which I built from a .yml file given by someone else. I added PyCall with `Pkg.add()`, then set `ENV["PYTHON"] = "/Users/qingyanz/opt/anaconda3/envs/pymatg/bin/python3"`, then used `Pkg.build("PyCall")`.

PyCall refused to build due to 100s of dependency conflicts, and the error log is given below. I then realized that PyCall only supports virtual environments created with `venv` and `virtualenv`, and NOT `conda`, and tried to revert to Julia's default vanilla Python distribution by setting `ENV["PYTHON"] = ""`, but the error message remained exactly the same. I tried reinstalling PyCall, deleting all relevant directories, and browsed all relevant question threads, but nothing seems to work. Therefore, I will deeply appreciate your input.

```
PREFIX=/Users/qingyanz/opt/anaconda3/envs/pymatg

#= 
# Hundreds of Python package conflict info redacted here
#= 

Package matplotlib-base conflicts for:
pymatgen -> matplotlib-base[version='>=1.5']
matplotlib-base
ase -> matplotlib -> matplotlib-base==3.1.3=py38h11da6c2_0

Package pydispatcher conflicts for:
pymatgen -> pydispatcher[version='>=2.0.5']
pydispatcher

Package tabulate conflicts for:
pymatgen -> tabulate
tabulate

Package traitlets conflicts for:
ipython -> traitlets[version='>=4.2']
traitlets

┌ Info: Using the Python distribution in the Conda package by default.
└ To use a different Python version, set ENV["PYTHON"]="pythoncommand" and re-run Pkg.build("PyCall").
[ Info: Downloading miniconda installer ...
[ Info: Installing miniconda ...
ERROR: LoadError: failed process: Process(`/Users/qingyanz/opt/anaconda3/envs/pymatg/installer.sh -b -f -p /Users/qingyanz/opt/anaconda3/envs/pymatg`, ProcessExited(1)) [1]

Stacktrace:
 [1] run(::Cmd; wait::Bool) at /Applications/JuliaPro-1.4.1-1.app/Contents/Resources/julia/Contents/Resources/julia/lib/julia/sys.dylib:?
 [2] run at ./process.jl:438 [inlined]
[...]

PR #778, opened 2020-05-14
Add a more helpful error message when precompiling fails due to Python build change

nishaChandramoorthy:
Upon a Python update, running e.g. `using PyPlot` throws this error, using Julia 1.4.1 on ArchLinux (not a Conda install unlike #776 ):

```
julia> using PyPlot
[ Info: Precompiling PyPlot [d330b81b-6aea-500a-939a-2ce795aea3ee]
/usr/lib/libpython3.7m.so.1.0
ERROR: LoadError: LoadError: could not load library "/usr/lib/libpython3.7m.so.1.0"
/usr/lib/libpython3.7m.so.1.0: cannot open shared object file: No such file or directory
Stacktrace:
 [1] dlopen(::String, ::UInt32) at /build/julia/src/julia-1.4.1/usr/share/julia/stdlib/v1.4/Libdl/src/Libdl.jl:109
 [2] top-level scope at /home/nishac/.julia/packages/PyCall/zqDXB/src/startup.jl:48
 [3] include(::String) at /home/nishac/.julia/packages/PyCall/zqDXB/src/PyCall.jl:1
 [4] top-level scope at /home/nishac/.julia/packages/PyCall/zqDXB/src/PyCall.jl:34
 [5] top-level scope at none:2
 [6] eval at ./boot.jl:331 [inlined]
in expression starting at /home/nishac/.julia/packages/PyCall/zqDXB/src/startup.jl:41
in expression starting at /home/nishac/.julia/packages/PyCall/zqDXB/src/PyCall.jl:34
ERROR: LoadError: Failed to precompile PyCall [438e738f-606a-5dbb-bf0a-cddfbfd45ab0] to /home/nishac/.julia/compiled/v1.4/PyCall/GkzkC_4xJNN.ji.
Stacktrace:
 [1] top-level scope at none:2
 [2] eval at ./boot.jl:331 [inlined]
in expression starting at /home/nishac/.julia/packages/PyPlot/XHEG0/src/PyPlot.jl:4
```

This error goes away on re-building `PyCall` because `deps/deps.jl` is rewritten with the correct Python environment variables. However, updating the packages or `Pkg.add`-ing PyPlot or even PyCall will not fix the issue. This may not be obvious to someone however (although, when using Conda.jl, that rebuilding PyCall helps is mentioned in the [PyPlot docs](https://github.com/JuliaPy/PyPlot.jl/blob/master/README.md#automated-matplotlib-installation)). Perhaps we can explicitly say that PyCall needs to be rebuilt along with this error message. In the location mentioned below, there does not seem to be any other scenario when `libpython` holds the wrong library from `deps/deps.jl`, so it seems likely that the error message to rebuild PyCall would help others.

Thanks for your time!


src/startup.jl
```diff
@@ -44,7 +44,13 @@ if !symbols_present
     isfile(depfile) || error("PyCall not properly installed. Please run Pkg.build(\"PyCall\")")
     include(depfile) # generated by Pkg.build("PyCall")
     # Only to be used at top-level - pointer will be invalid after reload
-    libpy_handle = Libdl.dlopen(libpython, Libdl.RTLD_LAZY|Libdl.RTLD_DEEPBIND|Libdl.RTLD_GLOBAL)
+    try 
+		Libdl.dlopen(libpython, Libdl.RTLD_LAZY|Libdl.RTLD_DEEPBIND|Libdl.RTLD_GLOBAL)
+	catch err
+		error(err.msg, ". Please run `Pkg.build(\"PyCall\")` if your Python build has changed")
```
stevengj, line 50: This will fail if `err` doesn't have a `msg` field.  Maybe
```jl
if err isa ErrorException
    error(err.msg, ...)
else
   rethrow(err)
end
```

nishaChandramoorthy: Thanks @stevengj ! Added this if in the next commit...

src/startup.jl
```diff
@@ -44,15 +44,20 @@ if !symbols_present
...
     include(depfile) # generated by Pkg.build("PyCall")
     # Only to be used at top-level - pointer will be invalid after reload
-    try 
-		Libdl.dlopen(libpython, Libdl.RTLD_LAZY|Libdl.RTLD_DEEPBIND|Libdl.RTLD_GLOBAL)
+	try
+    	Libdl.dlopen(libpython, Libdl.RTLD_LAZY|Libdl.RTLD_DEEPBIND|Libdl.RTLD_GLOBAL)
 	catch err
-		error(err.msg, ". Please run `Pkg.build(\"PyCall\")` if your Python build has changed")
+		if err isa ErrorException
+			error(err.msg, ". Please run `Pkg.build(\"PyCall\")` if your Python build has changed")
```
stevengj, line 51: Use 4 spaces to indent

nishaChandramoorthy: Sorry, fixed.

src/startup.jl
```diff
@@ -44,8 +44,19 @@ if !symbols_present
...
     include(depfile) # generated by Pkg.build("PyCall")
     # Only to be used at top-level - pointer will be invalid after reload
-    libpy_handle = Libdl.dlopen(libpython, Libdl.RTLD_LAZY|Libdl.RTLD_DEEPBIND|Libdl.RTLD_GLOBAL)
+    try
+        Libdl.dlopen(libpython, Libdl.RTLD_LAZY|Libdl.RTLD_DEEPBIND|Libdl.RTLD_GLOBAL)
+    catch err
+        if err isa ErrorException
+            error(err.msg, ". Please run `Pkg.build(\"PyCall\")` if your Python build has changed")
+        else
+            error("Please run `Pkg.build(\"PyCall\")` if your Python build has changed")
```
stevengj, line 53: No `error` here.  Just `rethrow`.

stevengj: (If the libpython file is missing, it will throw an `ErrorException`.)

nishaChandramoorthy: Ok, removed. But, if the only two possibilities are an `ErrorException` object or a `Ptr` to nothing (if the file is found and the `catch` block is not executed) then, do we even need the `if`?

src/startup.jl
```diff
@@ -44,8 +44,18 @@ if !symbols_present
...
+        Libdl.dlopen(libpython, Libdl.RTLD_LAZY|Libdl.RTLD_DEEPBIND|Libdl.RTLD_GLOBAL)
+    catch err
+        if err isa ErrorException
+            error(err.msg, ". Please run `Pkg.build(\"PyCall\")` if your Python build has changed")
+        else
+            rethrow(err)
+        end
+    end
     # need SetPythonHome to avoid warning, #299
+    libpy_handle = Libdl.dlopen(libpython, Libdl.RTLD_LAZY|Libdl.RTLD_DEEPBIND|Libdl.RTLD_GLOBAL)
```
stevengj, line 57: Instead of calling `dlopen` twice, you can just do `libpy_handle = try` above, i.e. the library is the return value of the `try` block.

nishaChandramoorthy: Thank you, fixed!

src/startup.jl
stevengj: I don't know that those are the only two possibilities.  I only know that if the exception is not an `ErrorException` then the problem is probably not the Python installation.

stevengj:
Thanks!

PR #779, opened 2020-05-17
Update readme

PallHaraldsson:
[skip ci]


PallHaraldsson:
1. I was doing

py"""import oct2py"""

thinking there was still something wrong with my installation (I was previously having unrelated issues, and didn't read the error message carefully).

2. With Python 3.7 no longer supported, and e.g. not installed on my machine, it seems ok to have just the major version as an example, and people will figure out, minor also ok.

Issue #751, opened 2020-02-28
Can't convert a subtype of a base Python type to a custom Julia type


sethaxen:
When a Python package extends a basic type like a `dict` to add some functionality, we can't use `pytype_mapping` to convert it to a custom Julia type. Instead, the default conversion (in this case to `Dict`) takes precedent.

Example:

```julia
using PyCall
import Base: convert

py"""
class MyDict(dict):
   pass
x = MyDict({"a": "b", "c": "d"})
"""

struct MyDictWrapper
   o::PyObject
end

Base.convert(::Type{MyDictWrapper}, obj::PyObject) = MyDictWrapper(obj)

julia> pytype_mapping(py"MyDict", MyDictWrapper)
1-element Array{Tuple{PyObject,Type},1}:
 (PyObject <class '__main__.MyDict'>, MyDictWrapper)

julia> py"x" # expected: MyDictWrapper(PyObject {'a': 'b', 'c': 'd'})
Dict{Any,Any} with 2 entries:
  "c" => "d"
  "a" => "b"
```

Issue #700, opened 2019-07-25
Can't automatically convert KeyedTuple (from sqlalchemy) to Julia's NamedTuple


manuelma:
I use PyCall to call sqlalchemy from Julia, which uses a class called `KeyedTuple` to represent rows resulting from a query and I want to convert those automatically to julia's `NamedTuple`. I expected something like this to work:

```julia
using PyCall

sa = pyimport("sqlalchemy")

pytype_mapping(sa."util"."KeyedTuple", NamedTuple)

Base.convert(::Type{NamedTuple}, o::PyObject) = NamedTuple{Symbol.(o._fields)}([getproperty(o, f) for f in Symbol.(o._fields)])

x = sa.util.KeyedTuple([1, 2, 3], labels=["one", "two", "three"]) # returns (1, 2, 3)

```

This is no surprise, since:

```julia
o = @pycall sa.util.KeyedTuple([1, 2, 3], labels=["one", "two", "three"])::PyObject

pytype_query(o)  # Returns `Tuple{Integer,Integer,Integer}`
```

I guess a simple way to make my thing work would be to alter the order in which `pytype_query` does things> I mean first looking for a match in the `pytype_queries` array and then proceed with the rest...?

stevengj:
In general, we're trying to move away from more automatic conversions and just make it easier to use the `PyObject` as-is.

On the other hand, explicitly calling `NamedTuple(o)` or `convert(NamedTuple, o)` could and should be made to work.

manuelma:
Ok, cool, thanks for the reply. I guess I can disable the automatic conversion for my `KeyedTuple` (I believe the correct way to do this is through `@pycall function_call::PyAny`) and then treat it on my own.

manuelma:
I guess what's not so nice is that one explicitly needs to disable the automatic conversion for the `KeyedTuple` otherwise it just lands as a (non-named) tuple in Julia. One could even argue that this is an error since the automatic conversion is *losing* information. In the same line of reasoning, one could say that PyCall fails at determining that this KeyedTuple is more than just a tuple. 

I don't know why this happens, clearly it has something to do with `sqlalachemy`'s `KeyedTuple` implementation which makes `pyisinstance(o, @pyglobalobj :PyTuple_Type)` evaluate to `true` in `PyCall.pysequence_query`. Maybe I can try and take a look at the details.

manuelma:
There's not much we can do actually:

```python
>>> import sqlalchemy as sa
[...]

PR #792, opened 2020-07-12
give pytype_mapping precendence over default mappings

marius311:
Trivial fix to give the user-specified `pytype_mapping`s precedence over the default built-ins, such that things which subclass Python `array`, `dict`, etc... can have custom mappings in PyCall. 

@stevengj mentions in https://github.com/JuliaPy/PyCall.jl/issues/700#issuecomment-516049934 that you're trying to move away from needing these conversions at all, but given that its a year later and we're still not quite there, I think this is still useful in the meantime. 

Closes https://github.com/JuliaPy/PyCall.jl/issues/751, https://github.com/JuliaPy/PyCall.jl/issues/700


marius311:
Hey guys, can you take a look at this? I think this is right, and the current behavior is problematic.

stevengj:
LGTM.

marius311:
Awesome, thank you!

PR #794, opened 2020-07-14
change `py_str` to `py_cmd`

### LINKED ISSUE #793: replace `py_str` with `py_cmd` literal?
I think, semantically it makes more sense to use custom command literals instead of string literals for code in other languages, since I think it is important to distinguish code from plain strings. The default shell is already a different language within Julia, so I think it is reasonable to extend this concept to interop with other languages. This could also then be used by editors to customize syntax highlighting for Python code inside Julia, as already proposed in https://github.com/JuliaEditorSupport/julia-emacs/issues/132. The current string syntax could still be kept around, of course, and perhaps deprecated eventually. I would be curious to hear others' opinions on this.

simeonschaub:
Do we already want to deprecate `@py_str`? I left it as-is for now, since a lot of code currently depends on PyCall, so throwing a deprecation warning might get annoying pretty quickly. I also still need to update the README and the docs.
closes #793


README.md
```diff
@@ -145,45 +145,45 @@ which is automatically converted to a Julia type, you will have override this
 via `@pywith EXPR::PyObject ...`.
 
 If you are already familiar with Python, it perhaps is easier to use
-`py"..."` and `py"""..."""` which are equivalent to Python's
+``py`...``` and ````py```...``````` which are equivalent to Python's
```
simeonschaub, line 148: The latter doesn't seem to work either. Does anyone know how to do that with GitHub's Markdown syntax?

stevengj: Just putting a space seems to work:
```
` py```...``` `
```

stevengj:
I agree that we should leave the string version without a deprecation for now.

README.md
simeonschaub: Ah thanks!

simeonschaub: That still seems to leave a trailing whitespace highlighted, but probably not that noticeable.

simeonschaub: Ah, wait. I misunderstood!

simeonschaub: Should be fixed now

tkf:
@simeonschaub Sorry to comment it after you made all the changes. But I think there are a couple of conceptual issues: https://github.com/JuliaPy/PyCall.jl/issues/793#issuecomment-659068451

Issue #815, opened 2020-09-07
AOT test failure in Julia 1.6-DEV


tkf:
> ```
> [ Info: PackageCompiler: creating system image object file, this might take a while...
> ERROR: ArgumentError: Package PyCall not found in current path:
> - Run `import Pkg; Pkg.add("PyCall")` to install the PyCall package.
> ```
>
> --- https://travis-ci.org/github/JuliaPy/PyCall.jl/jobs/722001094#L738

I think this is an upstream issue: https://github.com/JuliaLang/julia/issues/37441

I think this failure is observable only if you don't have PyCall in the default environment.

PR #823, opened 2020-09-21
AOT testing: Workaround upstream --project + --output-o issue

tkf:
This patch implements a workaround for https://github.com/JuliaLang/julia/issues/37441 (https://github.com/JuliaPy/PyCall.jl/issues/815); i.e., since the current project is ignored during system image build in Julia 1.6-DEV, we `dev` PyCall into the global environment.

It's a bit ugly but I think it's worth doing this since it would have helped us catching other regressions like https://github.com/JuliaLang/PackageCompiler.jl/issues/445.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,14 @@ matrix:
         - aot/assert_has_pycall.jl
         - aot/runtests.sh
       after_success: skip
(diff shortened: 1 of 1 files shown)


tkf:
This was included #825.

PR #825, opened 2020-09-22
Use GitHub Actions to run CI

tkf:
I've been using GitHub Actions in PyJulia for a while and it's been working nicely. The best part of it is that we don't need OS-specific configurations and also Python on macOS just works. It's also nice that we don't need to click through pages to find out the failing combination.

Since travis-ci.org will be closed down on December 31, 2020, maybe it's a good opportunity to move to GitHub Actions? (Although migration to travis-ci.com is another option.)

https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom?e=%5BUNIQID%5D
https://docs.travis-ci.com/user/migrate/open-source-repository-migration

Some notes:

* I separated out conda and AOT tests from the system Python tests. Although we can save ~20 lines if we combine system and conda Python tests, I think it's cleaner to separate them. Also, re-running test is less time-consuming this way (GitHub Actions only support re-running the entire matrix).
* This PR includes a workaround as in #823 (#815).

This PR does not disable Travis and AppVeyor yet. I think it's safe to run all of them for a while to make sure that this setup works. But I'm OK with removing them in this PR as well.

### DIFF
--- a/.github/workflows/aot.yml
+++ b/.github/workflows/aot.yml
@@ -38,7 +38,10 @@ jobs:
         with:
           version: ${{ matrix.julia-version }}
           arch: ${{ matrix.architecture }}
-      - run: julia -e 'using Pkg; pkg"add PackageCompiler@1"'
+
+      # Revert to `@v1` after this PR is merged:
+      # https://github.com/JuliaLang/PackageCompiler.jl/pull/443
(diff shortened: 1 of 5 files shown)


tkf:
I think this is ready to merge. AOT test with Julia nightly fails but it is an upstream issue (https://github.com/JuliaLang/PackageCompiler.jl/issues/445) tracked by #815.

stevengj:
Thanks!

PR #826, opened 2020-09-22
Test Julia 1.5, skip older

PallHaraldsson:


tkf:
Thanks for the PR. It should be fixed by #825 now.

PR #827, opened 2020-09-22
Test Julia 1.5, less for unsupported Python 2.7

PallHaraldsson:


PallHaraldsson:
I was getting false positive here:
https://github.com/JuliaPy/PyCall.jl/pull/724#issuecomment-683325073

tkf:
Hi, thanks for the PR. It should be fixed by #825 now.

PR #841, opened 2020-09-28
Fix get_real_method; don't use pyversion_build

tkf:
I should've used `pyversion` instead of `pyversion_build` so that it works in PyJulia mode.

close https://github.com/JuliaPy/pyjulia/issues/424

@stevengj Can we release 1.92.1 just after merging this PR?  I incremented the version in Project.toml.


stevengj:
LGTM.

Issue #843, opened 2020-10-02
Better error printing


jw3126:
I often get errors that look like this:
![image](https://user-images.githubusercontent.com/7261506/94902935-450e3b00-0499-11eb-948c-dd86bac54db7.png)

There are a couple of things that make this pretty hard to read. The most important one is probably that instead of an actual line break there appears a literal `\n`.

Could this be improved? And is it hard to do so?

PallHaraldsson:
I looked a bit into it, and did see this function: https://github.com/JuliaPy/PyCall.jl/blob/8e3abedbc402a8fc36cb2727f33a5b6cd1f0a1e9/src/exception.jl#L180

It's probably not too hard to change this. Still, not sure, it seems you send many lines to Python to evaluate (that's supported), maybe they should be broken up by you (or PyCall.jl) to many evaluations?

jw3126:
Not sure, what you mean by sending many lines to python, but I just call a single method of a python object.

jw3126:
The above error is from `tensorflow.keras` and it is representative of what errors from keras look like for me.

PallHaraldsson:
It seems I misread the cause (i.e. you're not using multi-line py""" vs py "), and rereading this looks like some stack-trace. It should be simple to take it from Python (Keras?) and change \n to an actual newline. There might be a side-effect that some replacements are not wanted... so I'm not sure what's best to do.

PR #844, opened 2020-10-02
Prettyerrors

jw3126:
On master the following snippet:
```julia
using PyCall
py"""
def raise_multi_line_error():
    raise Exception("1\n2\n3\n4\n")
"""
py"raise_multi_line_error"()
```
gives
```
ERROR: LoadError: PyError ($(Expr(:escape, :(ccall(#= /home/jan/.julia/dev/PyCall/src/pyfncall.jl:43
 =# @pysym(:PyObject_Call), PyPtr, (PyPtr, PyPtr, PyPtr), o, pyargsptr, kw))))) <class 'Exception'>
Exception('1\n2\n3\n4\n')
  File "/home/jan/.julia/dev/PyCall/src/pyeval.jl", line 2, in raise_multi_line_error
    const Py_file_input = 257
```
This is hard to read. See also #843 . With this PR it becomes:
```
ERROR: LoadError: PyError:
1
2
3
4

  File "/home/jan/.julia/dev/PyCall/src/pyeval.jl", line 2, in raise_multi_line_error
    const Py_file_input = 257
```

### DIFF
--- a/src/exception.jl
+++ b/src/exception.jl
@@ -41,6 +41,18 @@ function pystr_nofail(o::PyObject)
     end
 end
 
+function Base.showerror(io::IO, e::PyError)
+    println(io, "PyError:\n", pystr_nofail(e.val))
+    if !ispynull(e.traceback)
+        o = pycall(format_traceback, PyObject, e.traceback)
+        if !ispynull(o)
+            for s in PyVector{AbstractString}(o)
+                print(io, s)
+            end
+        end
+    end
+end
+
 function show(io::IO, e::PyError)
     print(io, "PyError",
           isempty(e.msg) ? e.msg : string(" (",e.msg,")"),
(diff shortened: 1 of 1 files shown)


src/exception.jl
```diff
@@ -28,11 +28,35 @@ struct PyError <: Exception
...
+        if (s == C_NULL)
+            pyerr_clear()
+            return string(PyPtr(o))
+        end
+        return convert(AbstractString, PyObject(s))
+    end
+end
+
+function Base.showerror(io::IO, e::PyError)
+    println(io, "PyError:\n", pystr_nofail(e.val))
```
jw3126, line 45: This line is crucial. `e.val` is the python exception object. And the question is how should we print it. Master does it calling `repr`. From looking at some tutorials, it seems that calling `str` is preffered. But I am by no means sure if it is the best analog of `showerror` in python. Or if it even depends on the python version, whats the preferred thing.

jw3126 (commented):
The code is a bit quick and dirty and there are no tests. I would like to have some feedback about how errors should look like and if this is the right approach, before I clean it up.

jw3126:
Is there interest in this PR?

jw3126:
@stevengj can you briefly comment if there is interest in this PR? So I know if I should polish / redesign / abandon it.

PR #851, opened 2020-10-16
Test on Python 3.9

PallHaraldsson:
FYI: It's out, not sure if premature to test and ok to override 3.8 test. [There's some new stuff in 3.9, regarding e.g. Dicts and they caught up with us/Unicode. Nothing that I think should be a problem.]

Another FYI: I noticed at pybind11 "Python 3.9.0 bug workaround implemented" (and better support for C++20)
https://github.com/pybind/pybind11/releases/tag/v2.6.0rc2

I guess this doesn't affect PyCall (or pyjulia?), those calling Python code shouldn't care, if a problem indirectly then pybind11 would take care of that.

### DIFF
--- a/.github/workflows/system.yml
+++ b/.github/workflows/system.yml
@@ -18,24 +18,24 @@ jobs:
           - macos-latest
           - windows-latest
         architecture: [x64]
-        python-version: ['2.7', '3.8']
+        python-version: ['2.7', '3.9']
         julia-version: ['1.0', '1.5', 'nightly']
         include:
           - os: windows-latest
             architecture: x86
-            python-version: '3.8'
+            python-version: '3.9'
             julia-version: '1.5'
           - os: ubuntu-latest
             architecture: x64
-            python-version: '3.7'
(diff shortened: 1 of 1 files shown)


PallHaraldsson:
At least I don't think I'm responsible for Python 2.7 macos failing. And I thought we were just going to drop testing Python 2.7? Should there be an exception for macOS? I see in the just released Python 3.9 docs: "Mac OS X 10.8 comes with Python 2.7 pre-installed by Apple." I knew they were behind on Python 2, and I'm not sure if they're still on latest macOS. Can they also use Conda, or do by default to get Python 3?

I realize now this was intentional, and now I have duplicates:
python-version: '3.7'

PallHaraldsson:
With Julia 1.6 just out, and Python 3.9 out for a while maybe that combination should be tested. I'm not sure why this failed at the time, maybe outdated reason, nor know how to retest (I'm not even sure I can do it).

tkf:
closed by #933

PR #877, opened 2021-01-12
Fix tests, Core.TypeofVararg for Julia 1.7 and AOT

mkitti:
I extracted this commit from the "transpose_and_slice" branch to focus on fixing the tests and AOT now that several issues have been resolved.

### DIFF
--- a/.github/workflows/aot.yml
+++ b/.github/workflows/aot.yml
@@ -41,13 +41,7 @@ jobs:
           arch: ${{ matrix.architecture }}
           show-versioninfo: true
 
-      # Revert to `@v1` after this PR is merged:
-      # https://github.com/JuliaLang/PackageCompiler.jl/pull/443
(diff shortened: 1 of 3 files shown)


mkitti:
Hi @stevengj and @tkf ,

This PR makes PyCall nightly (1.7) compatible and updates the AOT code now that several issues have been resolved.

I believe it will also run the AOT tests on code in a PR rather than just on master. There was a line that would dev PyCall, but this just ended up pulling down master.
https://github.com/JuliaPy/PyCall.jl/pull/877/files#diff-4b46bebfbbf4ad893abfcd8217a3e897e27cf16134a5970b31739320ef5c0894L50

I extracted this from #876 so we can see that the tests pass without the other changes and so that these changes can be merged as soon as possible.

stevengj:
Looks good, thanks.

Issue #881, opened 2021-02-19
Loading PyCall causes segfault on any subsequent Julia interrupt


marius311:
On this one particular system I'm using, whenever I load PyCall, any time after I try to Ctrl+C interrupt, I get a segfault. Here's e.g. a single session, the first interrupt is fine, then after loading PyCall its a segfault. 

```
julia> inv(rand(3000,3000))
^CERROR: InterruptException:
Stacktrace:
 [1] Array
   @ ./boot.jl:448 [inlined]
 [2] getri!(A::Matrix{Float64}, ipiv::Vector{Int64})
   @ LinearAlgebra.LAPACK /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/lapack.jl:1043
 [3] inv!
   @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/lu.jl:475 [inlined]
 [4] inv(A::Matrix{Float64})
   @ LinearAlgebra /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/dense.jl:812
 [5] top-level scope
   @ REPL[1]:1

julia> using PyCall

julia> inv(rand(3000,3000))
^CFatal Python error: Segmentation fault

Current thread 0x00002aaaaab3f380 (most recent call first):

signal (11): Segmentation fault
in expression starting at none:0
raise at /lib64/libpthread.so.0 (unknown line)
_IO_funlockfile at /lib64/libpthread.so.0 (unknown line)
maybe_collect at /buildworker/worker/package_linux64/build/src/julia_threads.h:313 [inlined]
jl_gc_pool_alloc at /buildworker/worker/package_linux64/build/src/gc.c:1176
push! at ./array.jl:936 [inlined]
```

This is 100% reproducible, and it segfaults 100% of the time after loading PyCall. This is with PyCall 1.92.2, Julia 1.6-beta1 installed from binaries with 1 thread and `Python 3.7.3 (default, Mar 27 2019, 22:11:17) [GCC 7.3.0] :: Anaconda, Inc. on linux`. 

Any suggestions for what I can do to try and debug this further / fix it?

marius311:
I've got an even simpler reproducer. I'm guessing its the same issue, the stack trace is identical and its also triggered by loading PyCall. Run the following with `julia --threads=2`:

```julia
using PyCall

[...]

PR #883, opened 2021-02-25
Make GC more thread-safe

### LINKED ISSUE #882: thread safety
If Julia's `Threads.nthreads() > 1`, we may want to do some more work to ensure thread safety.  In particular:

* Call `PyEval_InitThreads()` in `__init__` (this is only needed for Python ≤ 3.6, and is a no-op in Python ≥ 3.7).
• Acquire the GIL in the PyObject `finalizer`, by calling `PyGILState_Ensure` and `PyGILState_Release`.  This is because the Julia GC may be called from threads other than the main thread, so we need to ensure that we hold Python's GIL before decref-ing.

We might also expose an API to acquire the GIL, but in general I would recommend that user code should only call Python from the main thread.

marius311:
Here's my maybe naive way to fix #882.

Trying to aquire the GIL from the finalizer led to deadlocks. This is instead basically option 3 from [here](https://docs.julialang.org/en/v1/manual/multi-threading/#Safe-use-of-Finalizers). Tests pass locally and this fixes the original issue, but I'm not an expert so maybe this has other consequences I'm not thinking of. Comments welcome.

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -85,11 +85,11 @@ end
 # queue to try again later when we'll eventually be on the main thread.
 function _pydecref_eventually(po)
     if Threads.threadid()==1
-        return pydecref(po)
+        pydecref(po)
     else
         finalizer(_pydecref_eventually, po)
-        return nothing
     end
+    return nothing
 end
 
 PyPtr(o::PyObject) = getfield(o, :o)
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -76,8 +76,19 @@ mutable struct PyObject
     o::PyPtr # the actual PyObject*
     function PyObject(o::PyPtr)
         po = new(o)
-        finalizer(pydecref, po)
-        return po
+        finalizer(_pydecref_eventually, po)
+    end
+end
+
+# we can only call pydecref from the main thread. if we happen to hit
+# GC from a different thread, add the finalizer back to the finalization
+# queue to try again later when we'll eventually be on the main thread.
+function _pydecref_eventually(po)
+    if Threads.threadid()==1
+        return pydecref(po)
+    else
+        finalizer(_pydecref_eventually, po)
+        return nothing
     end
 end
 
--- a/src/pybuffer.jl
+++ b/src/pybuffer.jl
@@ -32,7 +32,7 @@ mutable struct PyBuffer
         b = new(Py_buffer(C_NULL, PyPtr_NULL, 0, 0,
                           0, 0, C_NULL, C_NULL, C_NULL, C_NULL,
                           C_NULL, C_NULL, C_NULL))
-        finalizer(pydecref, b)
+        finalizer(_pydecref_eventually, b)
         return b
     end
 end


stevengj:
See my comment here: https://github.com/JuliaPy/PyCall.jl/issues/882#issuecomment-786266352

I think we probably need to copy [the FFTW.jl solution](https://github.com/JuliaMath/FFTW.jl/blob/2c20abe7098a2380fa37fe021b3d21fc593059ec/src/fft.jl#L275-L325) here, which was rather painfully developed with the help of @vtjnash. 

(I wonder if we should refactor that solution out into a package so that both FFTW.jl and PyCall.jl can use the same code?  The logic is pretty subtle to get right.)

src/pyinit.jl
```diff
@@ -190,6 +190,24 @@ function __init__()
         Py_SetPythonHome(libpy_handle, pyversion, pyhome)
         Py_SetProgramName(libpy_handle, pyversion, current_python())
         ccall((@pysym :Py_InitializeEx), Cvoid, (Cint,), 0)
+
+        # _pydecref_queue used to pydecref only on the main thread:
```
stevengj, line 194: If someone calls Python from a different thread we are in trouble…

marius311: Thats right, but at least with this we can say "you can run with threads>1, just dont call Python from a different thread" whereas the current situation is "running with threads>1 and PyCall may segfault even if you only call Python from the main thread". Or is there a scenario where a user which is careful to only call from the main thread may trigger calls from other threads? Anyway, not trying to push for this PR exactly, just wanted to point that out.

src/pyinit.jl
```diff
@@ -190,6 +190,24 @@ function __init__()
...
+                if !isempty(_pydecref_queue)
+                    lock(_pydecref_queue_lock)
+                    o = try
+                        pop!(_pydecref_queue, o)
+                    finally
+                        unlock(_pydecref_queue_lock)
+                    end
+                    pydecref(o)
+                else
+                    yield()
```
stevengj, line 206: I'm not sure I like the thought of putting the CPU in a spinloop calling `yield()` over and over any time PyCall is loaded…

stevengj:
Hmm, I'm not sure we can factor the FFTW.jl solution that cleanly into a package, since `trylock(fftwlock)` probably needs to be replaced with the GIL lock (or we need to protect every call to Python with our own lock, which would be unfortunate).  Unfortunately, there doesn't seem to be any non-blocking analogue of `trylock` for the GIL?   So, we might still need to re-think the logic a bit.

src/pyinit.jl
stevengj: (The basic principle of the FFTW.jl solution is that you can flush the queue of objects to finalize the next time you have a finalization that succeeds in grabbing the lock.)

stevengj:
Reading the documentation some more, it seems like the solution may be simpler — we're not really allowed to call Python from anything other than the main thread, so if `Threads.threadid() != `` (we aren't in the main thread) in the finalizer, just push the PyPtr to a queue of objects to decref (protected by a lock as in FFTW), and otherwise both decref the objects and flush the queue.

marius311:
> I'm not sure I like the thought of putting the CPU in a spinloop
Yea youre right, I'm showing some of my ignorance on this stuff here. 

Re: the last comment, I think I see your point and maybe a simple way to do it, also basically amounts (I think) to option two from that Julia doc link instead of option 3. Will give it a go.

marius311:
Ok, this is quite simple and seems to work.

marius311:
Actually, sorry, I realize this doesn't actually solve the segfault I was seeing in https://github.com/JuliaPy/PyCall.jl/issues/881#issuecomment-785780082. I was getting confused since that only happens on this one system. So fwiw, I don't have an evironment where this PR fixes anything, either its not needed, or I get those segfaults regardless :shrug: granted it seems like a right thing to do.

mkitti:
Is it sufficient to guarantee that `Thread.threadid() == 1`? Perhaps we also need confirm that `Base.current_task() === Base.roottask` to ensure the stack is in a valid state for Python.

In JavaCall.jl, this is a necessary condition for some operations due to how Julia addresses the stack for each `Task` and how Java uses signalling. See the `JULIA_COPY_STACKS` environment variable and `ALWAYS_COPY_STACKS` option.
https://julialang.org/blog/2019/07/multithreading/#allocating_and_switching_task_stacks

src/PyCall.jl
```diff
@@ -76,8 +76,19 @@ mutable struct PyObject
...
+end
+
+# we can only call pydecref from the main thread. if we happen to hit
+# GC from a different thread, add the finalizer back to the finalization
+# queue to try again later when we'll eventually be on the main thread.
+function _pydecref_eventually(po)
+    if Threads.threadid()==1
+        return pydecref(po)
+    else
+        finalizer(_pydecref_eventually, po)
```
stevengj, line 90: Are we sure that this doesn't call the finalizer right away?  I'm not really sure what `finalizer` does under the hood.

src/PyCall.jl
```diff
@@ -76,8 +76,19 @@ mutable struct PyObject
...
+    end
+end
+
+# we can only call pydecref from the main thread. if we happen to hit
+# GC from a different thread, add the finalizer back to the finalization
+# queue to try again later when we'll eventually be on the main thread.
+function _pydecref_eventually(po)
+    if Threads.threadid()==1
+        return pydecref(po)
```
stevengj, line 88: Probably we could just return `nothing` from this function in all cases.

src/PyCall.jl
stevengj: Oh, I see that this is explicitly documented.  Good!

vtjnash: Yes, but note it is likely to be called again very soon, and probably on the same thread

marius311:
Fixed the return value. Up to you to decide what to do with this PR. Personally I'm far from sure on what the right solution is, if one is needed at the moment at all .

stevengj:
What does PythonCall.jl do, @cjdoris?

cjdoris:
> What does PythonCall.jl do, @cjdoris?

All the finalisers in PythonCall go via `enqueue` or `enqueue_all` [here](https://github.com/cjdoris/PythonCall.jl/blob/main/src/gc.jl).

These functions either immediately decref the pointer, or add it to a list of pointers to decref later. However, this is currently purely controlled by explicit `PythonCall.GC.disable()/enable()`, and like most other PythonCall functions should only be called from thread 1. It could be changed to check the thread and only immediately decref on thread 1 - which I think is the subject of the current PR, so I'll wait to see how you do it safely!

stevengj:
> So fwiw, I don't have an environment where this PR fixes anything, either its not needed, or I get those segfaults regardless

@marius311, is there any update on whether this PR observably fixes anything?

MilesCranmer:
I also think even though this PR might not be the optimal GC strategy, at least it's safe. We can always improve it later.

ericphanson:
btw I have a similar PR for PythonCall here: https://github.com/JuliaPy/PythonCall.jl/pull/520. My version is a bit more complicated as I'm using a dedicated task on thread 1 to process the GC queue (and some of the complication is just around testing that it's working properly, and since PythonCall has enable/disable functionality for it's GC).

PR #905, opened 2021-06-02
Assignment to `exc` in soft scope is ambiguous 

heetbeet:
Suppress warning:

```
┌ Warning: Assignment to `#6#exc` in soft scope is ambiguous because a global variable by the same name exists: `#6#exc` will be treated as a new local. Disambiguate by using `local #6#exc` to suppress this warning or `global #6#exc` to assign to the existing global variable.
└ @ C:\Users\simon\devel\empty_app\bin\julia\localdepot\packages\PyCall\BD546\src\PyCall.jl:654
```


stevengj:
This isn't the correct fix because it only works for `@pywith` statements in global scope.

stevengj:
The right fix is probably to wrap the whole `try .. finally` in a `let exc=true` block.

PR #910, opened 2021-06-12
Update numpy.jl

colbec:
Eliminate warning about `vendor()` being deprecated in favour of `:mkl` in `LinearAlgebra.BLAS.get_config().build_flags`

### DIFF
--- a/src/numpy.jl
+++ b/src/numpy.jl
@@ -64,7 +64,7 @@ function npyinitialize()
     numpy = pyimport("numpy")
 
     # emit a warning if both Julia and NumPy are linked with MKL (#433)
-    if LinearAlgebra.BLAS.vendor() === :mkl &&
+    if :mkl in LinearAlgebra.BLAS.get_config().build_flags &&
        LinearAlgebra.BLAS.BlasInt === Int64 && hasproperty(numpy, "__config__")
         config = numpy."__config__"
         if hasproperty(config, "blas_opt_info")


src/numpy.jl
```diff
@@ -64,7 +64,7 @@ function npyinitialize()
     numpy = pyimport("numpy")
 
     # emit a warning if both Julia and NumPy are linked with MKL (#433)
-    if LinearAlgebra.BLAS.vendor() === :mkl &&
+    if :mkl in LinearAlgebra.BLAS.get_config().build_flags &&
```
stevengj, line 67: Doesn't work with Julia 1.0, which we still declare compatibility with.

colbec: I have committed a revision to my colbec-patch-1 branch, trying to get it reflected in the PR

src/numpy.jl
```diff
@@ -64,7 +64,12 @@ function npyinitialize()
     numpy = pyimport("numpy")
 
     # emit a warning if both Julia and NumPy are linked with MKL (#433)
-    if :mkl in LinearAlgebra.BLAS.get_config().build_flags &&
+    if isless(VERSION,v"1.1")
```
stevengj, line 67: You can just do `VERSION < v"1.1"`

src/numpy.jl
```diff
@@ -64,7 +64,12 @@ function npyinitialize()
...
+        julia_compiled_with_mkl = occursin("MKL", LinearAlgebra.BLAS.openblas_get_config() )
```
stevengj, line 68: Why not just use the old code `LinearAlgebra.BLAS.vendor() === :mkl` in versions prior to `1.7`?

colbec: If that is the right way to go about it then fine; I don't have enough seniority to see results of checks to know what works in what version. What I proposed seems to work in 1.0.5 so that's what I went with. Regarding the `isless` I received errors when trying to compare to `VersionNumber` so tried to figure an alternative. I see that a simple < works with `VERSION`. It's all a learning process.

PR #927, opened 2021-10-17
Dict->IdDict in exception types

timholy:
Dicts use extensive specialization and this adds latency.
IdDicts don't and are the go-to associative container
for type-keys.

This one change reduces startup time by ~0.35s: in fresh sessions, with the package already precompiled, and `--startup-file=no`, `master` is

```julia
julia> tstart = time(); using PyCall; time() - tstart
1.188127040863037
```

whereas this branch is

```julia
julia> tstart = time(); using PyCall; time() - tstart
0.8567991256713867
```

### DIFF
--- a/src/exception.jl
+++ b/src/exception.jl
@@ -213,7 +213,7 @@ end
 
 function pyraise(e, bt = nothing)
     eT = typeof(e)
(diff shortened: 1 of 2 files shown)


timholy:
xref https://discourse.julialang.org/t/what-does-jl-init-restored-modules-do-why-does-loading-my-package-take-so-long/69906

PR #928, opened 2021-10-18
Test Python 3.10 (on Julia 1.6)

PallHaraldsson:


tkf:
I don't think the exception group needs an explicit support in PyCall?

FYI, we don't need to do this anymore after #932 and #933.

Issue #952, opened 2022-01-04
`] build PyCall` does not work on NixOS


samuela:
I'm running into some unexpected hiccups building PyCall.jl on NixOS. I have successfully used PyCall.jl on NixOS in the past, so I'm a bit stumped as to why it's borked now. I get the following failure:
```
(@v1.7) pkg> build PyCall
    Building Conda ─→ `~/.julia/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/6cdc8832ba11c7695f494c9d9a1c31e90959ce0f/build.log`
    Building PyCall → `~/.julia/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/71fd4022ecd0c6d20180e23ff1b3e05a143959c2/build.log`
ERROR: Error building `PyCall`:
PREFIX=/home/skainswo/.julia/conda/3
Unpacking payload ...
cannot execute /home/skainswo/.julia/conda/3/conda.exe: NIX_LD or NIX_LD_x86_64-linux is not set
cannot execute /home/skainswo/.julia/conda/3/conda.exe: NIX_LD or NIX_LD_x86_64-linux is not set
┌ Info: No system-wide Python was found; got the following error:
[...]

PR #955, opened 2022-01-11
Fix file path for file generated during the build process

samuela:
The built file is now `Pkg.dir("PyCall","deps","deps.jl")` as discovered in the investigation of https://github.com/JuliaPy/PyCall.jl/issues/952.


AtilaSaraiva:
Hey folks! Why is this PR not being reviewed?

PR #986, opened 2022-04-19
Python 3.10 is latest

PallHaraldsson:
I noticed Python 3.6 is EOLed, and was thinking you may not want say you support it (I didn't locate it), nor test it(?) to not waste resources (nor 2.7?). Maybe default to latest version, for Conda/Windows downloads, so people can profit from latest features?

Also there are security issues with all minor release of all 3.10, 3.9 down to 3.7 (documented) and likely 3.6 and 2.7 and older.

I'm not sure, but it seems 2.7 is still tested...

[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,9 +53,9 @@ to the path of the `python` (or `python3` etc.) executable and then re-running `
 In Julia:
 
     ENV["PYTHON"] = "... path of the python executable ..."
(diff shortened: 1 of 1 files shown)


stevengj:
Since these are just examples in the README, it's not a big deal, but might as well update.

PR #991, opened 2022-06-14
[WIP] Optionally store deps.jl in a different directory

jlapeyre:
This works together with CachePath.jl as [described here](https://discourse.julialang.org/t/cachepath-jl-allows-using-pycall-with-two-libpythons-and-no-recompilation/76325) to allow using different libpythons without rebuilding.

If this looks ok, I'll add documentation and an example, etc.


mkitti:
There is a test failure on system "Python / Test Julia nightly Python 3.x macos-latest x64".

Also it seems to me that the environment variables would only be checked during build or precompilation, rather than package module initialization. Is that correct?

Perhaps it might be best to use a [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl) mechanism to configure how PyCall finds libpython, and perhaps one of these modes could involve looking at environment variable.

stevengj:
Yes, I think the way to go is ideally to use Preferences.jl here.

MilesCranmer:
Is there an existing PR for using Preferences.jl?

PR #997, opened 2022-07-30
fix getproperty exceptions

### LINKED ISSUE #696: Wrong error reported when accessing an attribute
Hi! I have a class that does some checking before setting an attribute value, like this (pure Python):
```python
class A:
    def __setattr__(self, name, value):
        if value == 0:
            raise ValueError(value)
        # set value...

a = A()
a.a = 5  # OK
a.a = 0  # correct error message:
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-9-7600552ca637> in <module>()
----> 1 a.a = 0

<ipython-input-6-585623315894> in __setattr__(self, name, value)
      2     def __setattr__(self, name, value):
      3         if value == 0:
----> 4             raise ValueError(value)
      5         # set value...
      6

ValueError: 0
```

However, running the same code with PyCall.jl gives a very confusing error message:
```julia
py"""
class A:
    def __setattr__(self, name, value):
        if value == 0:
            raise ValueError(value)
        # set value...
"""
a = py"A"()
a.a = 5  # OK
a.a = 0  # completely wrong exception:
ERROR: KeyError: key :a not found
Stacktrace:
 [1] setproperty!(::PyObject, ::Symbol, ::Int64) at /users/aplavin/.julia/packages/PyCall/ttONZ/src/PyCall.jl:329
 [2] top-level scope at none:0
```

aplavin:
this throws KeyError as before when the attribute isn't there (AttributeError in python)
when the attribute is present, but its computation throws a python exception, this PR rethrows it in julia

fixes https://github.com/JuliaPy/PyCall.jl/issues/696

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -300,7 +300,7 @@ function _getproperty(o::PyObject, s::Union{AbstractString,Symbol})
     p = ccall((@pysym :PyObject_GetAttrString), PyPtr, (PyPtr, Cstring), o, s)
     if p == C_NULL && pyerr_occurred()
         e = pyerror("PyObject_GetAttrString")
-        if e.T != @pyglobalobjptr(:PyExc_AttributeError)
+        if PyPtr(e.T) != @pyglobalobjptr(:PyExc_AttributeError)
             throw(e)
         end
         pyerr_clear()
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -300,7 +300,7 @@ function _getproperty(o::PyObject, s::Union{AbstractString,Symbol})
     p = ccall((@pysym :PyObject_GetAttrString), PyPtr, (PyPtr, Cstring), o, s)
     if p == C_NULL && pyerr_occurred()
         e = pyerror("PyObject_GetAttrString")
-        if pyisinstance(e.T, @pyglobalobjptr(:PyExc_AttributeError))
+        if !pyisinstance(e.T, @pyglobalobjptr(:PyExc_AttributeError))
             throw(e)
         end
         pyerr_clear()
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -298,7 +298,13 @@ end
 function _getproperty(o::PyObject, s::Union{AbstractString,Symbol})
     ispynull(o) && throw(ArgumentError("ref of NULL PyObject"))
     p = ccall((@pysym :PyObject_GetAttrString), PyPtr, (PyPtr, Cstring), o, s)
-    p == C_NULL && pyerr_clear()
+    if p == C_NULL && pyerr_occurred()
+        e = pyerror("")
+        if e.T.__name__ != "AttributeError"
+            throw(e)
+        end
+        pyerr_clear()
+    end
     return p
 end
 
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -299,7 +299,7 @@ function _getproperty(o::PyObject, s::Union{AbstractString,Symbol})
     ispynull(o) && throw(ArgumentError("ref of NULL PyObject"))
     p = ccall((@pysym :PyObject_GetAttrString), PyPtr, (PyPtr, Cstring), o, s)
     if p == C_NULL && pyerr_occurred()
-        e = pyerror("")
+        e = pyerror("PyObject_GetAttrString")
         if e.T.__name__ != "AttributeError"
             throw(e)
         end
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -300,7 +300,7 @@ function _getproperty(o::PyObject, s::Union{AbstractString,Symbol})
     p = ccall((@pysym :PyObject_GetAttrString), PyPtr, (PyPtr, Cstring), o, s)
     if p == C_NULL && pyerr_occurred()
         e = pyerror("PyObject_GetAttrString")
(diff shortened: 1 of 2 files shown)


src/PyCall.jl
```diff
@@ -298,7 +298,13 @@ end
 function _getproperty(o::PyObject, s::Union{AbstractString,Symbol})
     ispynull(o) && throw(ArgumentError("ref of NULL PyObject"))
     p = ccall((@pysym :PyObject_GetAttrString), PyPtr, (PyPtr, Cstring), o, s)
-    p == C_NULL && pyerr_clear()
+    if p == C_NULL && pyerr_occurred()
+        e = pyerror("")
```
stevengj, line 302:
```suggestion
        e = pyerror("PyObject_GetAttrString")
```

src/PyCall.jl
```diff
@@ -298,7 +298,13 @@ end
...
+        if e.T.__name__ != "AttributeError"
```
stevengj, line 303:
```suggestion
        if pyisinstance(e.T, @pyglobalobjptr(:PyExc_AttributeError))
```

stevengj:
Looks good, modulo a couple of minor tweaks noted above.

stevengj:
What is `e.T` that `pyisinstance(e.T, @pyglobalobjptr(:PyExc_AttributeError)` fails to give the expected result?

stevengj:
What does `pyisinstance(e.T, @pyglobalobjptr(:PyExc_AttributeError))` return in those cases?

stevengj:
Oh, I see the problem.  `e.T` is not an *instance* of `AttributeError`, it is the type itself.  So the correct check is
```jl
if PyPtr(e.T) != @pyglobalobjptr(:PyExc_AttributeError)
```
(The goal here is to avoid the string conversion.)

src/PyCall.jl
```diff
@@ -298,7 +298,13 @@ end
 function _getproperty(o::PyObject, s::Union{AbstractString,Symbol})
     ispynull(o) && throw(ArgumentError("ref of NULL PyObject"))
     p = ccall((@pysym :PyObject_GetAttrString), PyPtr, (PyPtr, Cstring), o, s)
-    p == C_NULL && pyerr_clear()
+    if p == C_NULL && pyerr_occurred()
+        e = pyerror("PyObject_GetAttrString")
+        if !pyisinstance(e.T, @pyglobalobjptr(:PyExc_AttributeError))
```
stevengj, line 303:
```suggestion
        if e.T != @pyglobalobjptr(:PyExc_AttributeError)
```

src/PyCall.jl
```diff
@@ -327,15 +333,17 @@ setproperty!(o::PyObject, s::Symbol, v) = _setproperty!(o,s,v)
...
-    if ispynull(o)
-        throw(ArgumentError("assign of NULL PyObject"))
-    end
-    if -1 == ccall((@pysym :PyObject_SetAttrString), Cint,
-                   (PyPtr, Cstring, PyPtr), o, s, PyObject(v))
+    ispynull(o) && throw(ArgumentError("assign of NULL PyObject"))
+    p = ccall((@pysym :PyObject_SetAttrString), Cint, (PyPtr, Cstring, PyPtr), o, s, PyObject(v))
+    if p == -1 && pyerr_occurred()
+        e = pyerror("PyObject_SetAttrString")
+        if e.T.__name__ != "AttributeError"
```
stevengj, line 340:
```suggestion
        if e.T != @pyglobalobjptr(:PyExc_AttributeError)
```

src/PyCall.jl
```diff
@@ -298,7 +298,13 @@ end
 function _getproperty(o::PyObject, s::Union{AbstractString,Symbol})
     ispynull(o) && throw(ArgumentError("ref of NULL PyObject"))
     p = ccall((@pysym :PyObject_GetAttrString), PyPtr, (PyPtr, Cstring), o, s)
-    p == C_NULL && pyerr_clear()
+    if p == C_NULL && pyerr_occurred()
+        e = pyerror("PyObject_GetAttrString")
+        if e.T != @pyglobalobjptr(:PyExc_AttributeError)
```
stevengj, line 303:
```suggestion
        if PyPtr(e.T) != @pyglobalobjptr(:PyExc_AttributeError)
```

src/PyCall.jl
```diff
@@ -327,15 +333,17 @@ setproperty!(o::PyObject, s::Symbol, v) = _setproperty!(o,s,v)
...
-    if ispynull(o)
-        throw(ArgumentError("assign of NULL PyObject"))
-    end
-    if -1 == ccall((@pysym :PyObject_SetAttrString), Cint,
-                   (PyPtr, Cstring, PyPtr), o, s, PyObject(v))
+    ispynull(o) && throw(ArgumentError("assign of NULL PyObject"))
+    p = ccall((@pysym :PyObject_SetAttrString), Cint, (PyPtr, Cstring, PyPtr), o, s, PyObject(v))
+    if p == -1 && pyerr_occurred()
+        e = pyerror("PyObject_SetAttrString")
+        if e.T != @pyglobalobjptr(:PyExc_AttributeError)
```
stevengj, line 340:
```suggestion
        if PyPtr(e.T) != @pyglobalobjptr(:PyExc_AttributeError)
```

PR #1014, opened 2022-11-22
fix BoundsError of args[1] when args is empty

szcf-weiya:
if `args` on L101 is empty, then `args[1]` in `isexpr()` would throw `BoundsError: attempt to access 0-element Vector{Any} at index [1]`

```julia
julia> ex = :(f()::Float64)
:(f()::Float64)

julia> args = ex.args[1].args[2:end]
Any[]

julia> args[1]
ERROR: BoundsError: attempt to access 0-element Vector{Any} at index [1]
Stacktrace:
 [1] getindex(A::Vector{Any}, i1::Int64)
   @ Base ./array.jl:924
 [2] top-level scope
   @ REPL[118]:1

```

### DIFF
--- a/src/pyfncall.jl
+++ b/src/pyfncall.jl
@@ -99,10 +99,9 @@ macro pycall(ex)
     end
     func = ex.args[1].args[1]
     args, kwargs = ex.args[1].args[2:end], []
-    if length(args) > 0
-        if isexpr(args[1],:parameters)
-            kwargs, args = args[1], args[2:end]
-        end
+    if !isempty(args) && isexpr(args[1],:parameters)
+        kwargs, args = args[1], args[2:end]
+    end
     end
     T = ex.args[2]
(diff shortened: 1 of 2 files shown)


src/pyfncall.jl
```diff
@@ -99,8 +99,10 @@ macro pycall(ex)
     end
     func = ex.args[1].args[1]
     args, kwargs = ex.args[1].args[2:end], []
-    if isexpr(args[1],:parameters)
-        kwargs, args = args[1], args[2:end]
+    if length(args) > 0
+        if isexpr(args[1],:parameters)
+            kwargs, args = args[1], args[2:end]
+        end
```
stevengj, line 105:
```suggestion
    if !isempty(args) && isexpr(args[1],:parameters)
        kwargs, args = args[1], args[2:end]
    end
```
Might as well combine the `if` statements, and use `isempty`

Issue #757, opened 2020-03-26
`deepcopy(::PyObject)` can yield segfault


tkf:
I think we need to overload `deepcopy_internal(::PyObject, ...)` and increment the refcount in it?

```julia
julia> using PyCall

julia> a = pycall(pyimport("numpy").zeros, PyObject, 10000)
PyObject array([0., 0., 0., ..., 0., 0., 0.])

julia> pycall(pyimport("sys").getrefcount, Int, a)
2

julia> b = deepcopy(a)
PyObject array([0., 0., 0., ..., 0., 0., 0.])

julia> pycall(pyimport("sys").getrefcount, Int, a)
2

julia> pycall(pyimport("sys").getrefcount, Int, b)  # creating b did not incref
2

julia> pycall(pyimport("builtins").id, Int, a)
140671027744560

julia> pycall(pyimport("builtins").id, Int, b)  # just to show that b _is_ the same Python object
140671027744560

julia> a = nothing

julia> GC.gc(); GC.gc(); GC.gc(); GC.gc(); GC.gc(); GC.gc(); GC.gc()

julia> pycall(pyimport("builtins").id, Int, b)
140671027744560

julia> pycall(pyimport("sys").getrefcount, Int, b)  # now `PyPtr(b)` is pointing nowhere
140671027744561

julia> b  # sometime segfault doesn't happen immediately
PyObject ''uiltin_function_or_method'

julia> pyimport("gc").collect()

signal (11): Segmentation fault
in expression starting at REPL[14]:1
pymalloc_alloc at /tmp/python-build.20200119182925.141496/Python-3.7.6/Objects/obmalloc.c:1420 [inlined]
pymalloc_alloc at /tmp/python-build.20200119182925.141496/Python-3.7.6/Objects/obmalloc.c:1386 [inlined]
_PyObject_Malloc at /tmp/python-build.20200119182925.141496/Python-3.7.6/Objects/obmalloc.c:1578
_PyObject_GC_Alloc at /tmp/python-build.20200119182925.141496/Python-3.7.6/Modules/gcmodule.c:1693 [inlined]
_PyObject_GC_Malloc at /tmp/python-build.20200119182925.141496/Python-3.7.6/Modules/gcmodule.c:1715
_PyObject_GC_New at /tmp/python-build.20200119182925.141496/Python-3.7.6/Modules/gcmodule.c:1727
```

stevengj:
Sounds right.

PR #1039, opened 2023-06-09
enable deepcopy on PyObject #757

jw3126:


stevengj:
Why not just increment the refcount as proposed in #757?

src/PyCall.jl
```diff
@@ -979,6 +979,14 @@ include("serialize.jl")
 
 include("pyinit.jl")
 
+function Base.deepcopy_internal(obj::PyObject, stackdict::Base.IdDict)
+    msg = """
+    Deepcopy is disabled on PyObject. See
+    https://github.com/JuliaPy/PyCall.jl/issues/757
+    """
+    error(msg)
+end
```
stevengj, line 988:
```suggestion
const _deepcopy = PyNULL()

function Base.deepcopy_internal(obj::PyObject, stackdict::Base.IdDict)
    haskey(stackdict, obj) && stackdict[obj]
    ispynull(_deepcopy) && copy!(_deepcopy, pyimport("copy")["deepcopy"])
    return pycall(_deepcopy, PyObject, obj)
end
```

stevengj: Or alternatively the following implementation is more conservative:
```jl
function Base.deepcopy_internal(obj::PyObject, stackdict::Base.IdDict)
    haskey(stackdict, obj) && stackdict[obj]
    return pyincref(obj)
end
```

jw3126: Personally I expect that I can do whatever mutation I want with a deepcopy and it does never any effect on the original object. So from that perapective the refinc is a footgun. Too much copy is the lesser evil in my opinion. The third option is to throw an error and force the user to think how to best deal with the given case, which can be annoying. Probably your copy.deepcopy solution is the best option.

jw3126:
@stevengj Does this PR look good to you?

jw3126:
@stevengj LGTY?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "PyCall"
 uuid = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
 authors = ["Steven G. Johnson <stevenj@mit.edu>", "Yichao Yu <yyc1992@gmail.com>", "Takafumi Arakaki <aka.tkf@gmail.com>", "Simon Kornblith <simon@simonster.com>", "Páll Haraldsson <Pall.Haraldsson@gmail.com>", "Jon Malmaud <malmaud@gmail.com>", "Jake Bolewski <jakebolewski@gmail.com>", "Keno Fischer <keno@alumni.harvard.edu>", "Joel Mason <jobba1@hotmail.com>", "Jameson Nash <vtjnash@gmail.com>", "The JuliaPy development team"]
-version = "1.95.1"
+version = "1.95.2"
```
stevengj, line 4:
```suggestion
version = "1.96"
```
Probably should be a minor version bump since a functioning `deepcopy` is basically a new feature more than a bugfix.

test/runtests.jl
```diff
@@ -620,6 +620,44 @@ const PyInt = pyversion < v"3" ? Int : Clonglong
...
+    @test collect(l) == [1,2,3]
+
+    obj = py"""
+    class C757:
+        def __init__(self, a, b):
+            self.a = a
+            self.b = b
+    """
+    obj = py"C757(C757(1,2), C757(3,4))"o
+    obj2 = deepcopy(obj)
```
stevengj, line 639:
```suggestion
    obj2 = deepcopy(obj)
    @test PyPtr(obj) != PyPtr(obj2) # make sure a new Python object is created
```

stevengj: Would be good to test that it actually creates a new object, though I guess you did a related test with `l != l2` above.

jw3126:
Thanks, @stevengj can you tag a release?

stevengj:
Will do

PR #1040, opened 2023-06-12
Urgent: add option to disable numpy install / prevent corruption of conda-forge environments

MilesCranmer:
Right now PyCall.jl automatically installs numpy into a conda environment. This install cannot be disabled. If this occurs from within an existing conda environment, it can overwrite the numpy installed by a conda-forge build. This violates the assumptions of conda-forge, and can silently corrupt any conda env that depends on numpy.

This issue was noticed in https://github.com/conda-forge/pysr-feedstock/pull/85, manifesting in this issue: https://github.com/conda-forge/scipy-feedstock/issues/238. We believe this is caused by the line:

https://github.com/JuliaPy/PyCall.jl/blob/bcaba00d1e2c412b2f61d33343ef5a9ab1b9258a/deps/build.jl#L79

This PR enables the `PYCALL_INSTALL_NUMPY` environment flag which can be set to `false` to disable the `Conda.add("numpy")` line and prevent the corruption of a conda env. The default behavior is unchanged if this flag is not set, so this is backwards compatible.

cc @h-vetinari @mkitti @ngam

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
 VersionParsing = "81def892-9a0e-5fdd-b105-ffc91e053289"
 
 [compat]
-Conda = "1.0"
+Conda = "1.9"
 MacroTools = "0.4, 0.5"
 VersionParsing = "1.0"
 julia = "1.4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "PyCall"
 uuid = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
 authors = ["Steven G. Johnson <stevenj@mit.edu>", "Yichao Yu <yyc1992@gmail.com>", "Takafumi Arakaki <aka.tkf@gmail.com>", "Simon Kornblith <simon@simonster.com>", "Páll Haraldsson <Pall.Haraldsson@gmail.com>", "Jon Malmaud <malmaud@gmail.com>", "Jake Bolewski <jakebolewski@gmail.com>", "Keno Fischer <keno@alumni.harvard.edu>", "Joel Mason <jobba1@hotmail.com>", "Jameson Nash <vtjnash@gmail.com>", "The JuliaPy development team"]
-version = "1.95.1"
+version = "1.95.2"
 
 [deps]
 Conda = "8f4d0f93-b110-5947-807f-2305c1781a2d"
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -75,9 +75,8 @@ try # make sure deps.jl file is removed on error
     end
 
     use_conda = dirname(python) == abspath(Conda.PYTHONDIR)
-    enable_numpy_install = parse(Bool, get(ENV, "PYCALL_INSTALL_NUMPY", "true"))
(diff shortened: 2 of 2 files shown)


mkitti:
I'm not sure I'm following. All PyCall.jl wants to do is ensure that the conda environment it is operating in has `numpy` installed. Normally, this is a conda environment that is created specifically for a Julia install, but from a conda installed Julia this is an existing conda environment.

mkitti:
Could we consider adding `--freeze-installed` instead?

MilesCranmer:
https://github.com/JuliaPy/Conda.jl/blob/81fe00cb0388c8ea58f9244a0de0f9c5dd3534a1/src/Conda.jl#L224

I think this will reinstall, no?

MilesCranmer:
https://docs.conda.io/projects/conda/en/latest/commands/install.html

> Conda attempts to install the newest versions of the requested packages. To accomplish this, it may update some packages that are already installed, or install additional packages. To prevent existing packages from updating, use the --freeze-installed option.

(maybe that’s what you meant by `--freeze-installed`?)

MilesCranmer:
I think it’s safer to entirely turn off calls to `conda install` when inside a conda-forge build script though. Just feels a bit dangerous. What if even with `freeze-installed`, it installs some other packages which are not present in the env that conda-forge is building?

mkitti:
> What if even with `freeze-installed`, it installs some other packages which are not present in the env that conda-forge is building?

It should throw errors since you did not previously add all the needed dependencies.

MilesCranmer:
Okay I added the required functionality in https://github.com/JuliaPy/Conda.jl/pull/240

mkitti:
See https://github.com/JuliaPy/Conda.jl/pull/241

stevengj:
> Why would PyCall need numpy to begin with? That’s kinda odd and arbitrary (isn’t it supposed to be a generic python caller?)

In order to wrap Julia arrays with numpy arrays when calling Python.

stevengj:
> Why would PyCall need numpy to begin with? That’s kinda odd and arbitrary (isn’t it supposed to be a generic python caller?)

In order to wrap Julia arrays with numpy arrays when calling Python.

stevengj:
> Why would PyCall need numpy to begin with? That’s kinda odd and arbitrary (isn’t it supposed to be a generic python caller?)

In order to wrap Julia arrays with numpy arrays when calling Python.

mkitti (approved):
The revised build does not try to upgrade numpy if it is in the configured conda environment.

mkitti:
@stevengj could we merge and release this change?

MilesCranmer:
Just pinging this - could this be merged @stevengj?

mkitti:
I only have an auto-merge button here.

image

An chance we could get the other CI bits to pass? Python 2.7 should probably just be retired...

MilesCranmer:
It says:
```
Error: Version 2.7 with arch x64 not found
```

so we can't even test it. Also, pyjulia already dropped support.

MilesCranmer:
@stevengj sorry for the spam but we really need this fixed to solve some downstream failures. Any chance you could merge+release?

MilesCranmer:
Right, we bundle it with the PySR install as a special julia depot.

PR #1045, opened 2023-07-18
Try to resolve issues for x86 Windows

mkitti:
miniforge is no longer available for x86 Windows and does not have satisified solve

Fix https://github.com/JuliaPy/Conda.jl/pull/241#issuecomment-1636873725

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -76,7 +76,13 @@ try # make sure deps.jl file is removed on error
 
     use_conda = dirname(python) == abspath(Conda.PYTHONDIR)
     if use_conda
-        Conda.add("numpy"; satisfied_skip_solve=true)
+        if Sys.iswindows() && Sys.WORD_SIZE == 32
+            # Conda is no longer distributed for 32-bit Windows
+            # and does not have satisified_skip_solve
+            Conda.add("numpy")
+        else
+            Conda.add("numpy"; satisfied_skip_solve=true)
+        end
     end
 
     (libpython, libpy_name) = find_libpython(python)


deps/build.jl
```diff
@@ -76,7 +76,13 @@ try # make sure deps.jl file is removed on error
 
     use_conda = dirname(python) == abspath(Conda.PYTHONDIR)
     if use_conda
-        Conda.add("numpy"; satisfied_skip_solve=true)
+        if Sys.iswindows() && Sys.WORD_SIZE == 32
+            # Conda is no longer distributed for 32-bit Windows
+            # and does not have satisified_skip_solve
+            Conda.add("numpy")
+        else
+            Conda.add("numpy"; satisfied_skip_solve=true)
```
stevengj, line 84: Shouldn't Conda.jl just ignore this flag in that case?

mkitti: In most cases, I would want Conda.jl to error rather than silenting ignoring the keyword. The caller should either anticipate the error or handle the error.

Here, I'm just trying to maintain backwards compatability for a very specific situation, 32-bit Windows, which upstream conda no longer supports.

stevengj: I'm not 100% convinced that an error is the behavior `Conda.add` users would want.

The `satisfied_skip_solve` flag in practice mostly seems like an optional hint to speed things up, as it is used here, and it's a bit ugly that callers of Conda.jl will have to know about and implement the 32-bit windows exception themselves.

mkitti:
@davidanthoff any thoughts about this?

mkitti:
I'm trying to clarify conda's perspective on 32-bit Windows and why `--satisfied-skip-solve` may not be available on that platform.

This seems to be where conda-forge's support stopped:
https://github.com/conda-forge/staged-recipes/issues/5640

Scipy appears to have dropped support since October 2022:
https://discuss.python.org/t/dropping-32-bit-packages/5476/10

deps/build.jl
MilesCranmer: > Honestly, my preferred solution is to stop supporting conda on 32-bit Windows and Python 2.7 entirely due to lack of upstream support

I also see this as the best solution.

mkitti:
I created an alternative solution at https://github.com/JuliaPy/Conda.jl/pull/246 . There Conda.jl will issue a warning if someone tries to use `--satisified-skip-solve` on 32-bit Windows.

PR #1054, opened 2023-09-18
add 2-arg hash function

jverzani:
On discourse (https://discourse.julialang.org/t/sympy-sym-s-as-keys-in-dicts/103995) the issue of using tuples of symbolic keys let to the realization that the 2 argument form of `hash` resolved to a different code path than the 1-argument form. This consolidates the two.

If this PR is reasonable, it might also make sense to remove the precompile statement for `hashsalt` and the `hashsalt` function.

The approach taken in `PythonCall` might provide an alternative. (`Base.hash(x::Py, h::UInt) = reinterpret(UInt, Int(pyhash(x))) - 3h`)


src/PyCall.jl
```diff
@@ -273,20 +273,20 @@ end
 const pysalt = hash("PyCall.PyObject") # "salt" to mix in to PyObject hashes
 hashsalt(x) = hash(x, pysalt)
```
stevengj, line 274: Probably just delete this function (also from `src/precompile.jl`).

stevengj:
LGTM.  I agree about removing `hashsalt`.

I can't remember why I did it this way — may be from quite an old Julia version.

src/PyCall.jl
```diff
@@ -273,20 +273,20 @@ end
...
 
-function hash(o::PyObject)
+function hash(o::PyObject, salt::UInt64=pysalt)
```
stevengj, line 276:
```suggestion
function hash(o::PyObject, salt::UInt=zero(UInt))
    salt ⊻= pysalt
```

jverzani:
Thanks! Let me know if there is anything else to be done here.

stevengj (approved):

MilesCranmer:
@stevengj would it be possible to merge this? Thanks! - Miles

src/PyCall.jl
```diff
@@ -271,22 +271,22 @@ end
 # computing hashes of PyObjects
 
 const pysalt = hash("PyCall.PyObject") # "salt" to mix in to PyObject hashes
-hashsalt(x) = hash(x, pysalt)
 
-function hash(o::PyObject)
+function hash(o::PyObject, salt::UInt=zero(UInt))
```
stevengj, line 275:
```suggestion
function hash(o::PyObject, salt::UInt)
```
The default value is not needed here because Julia defines `hash(x::Any) = hash(x, zero(UInt))` for all types.

Issue #1052, opened 2023-08-29
notable startup latency


thautwarm:
It seems that the latency gets worse with newer Julia. Here are the startup tests on my local machine:

Julia v1.9.2:
```
julia> @time using PyCall
  0.427297 seconds (250.30 k allocations: 18.071 MiB, 1.88% compilation time)

julia> @time using PyCall
  1.162834 seconds (1.32 M allocations: 91.732 MiB, 5.27% gc time, 99.92% compilation time: 100% of which was recompilation)
```

Julia v1.7.3:
```
julia> @time using PyCall
  0.689246 seconds (1.16 M allocations: 81.998 MiB, 1.08% gc time, 42.07% compilation time)

julia> @time using PyCall
  0.630776 seconds (1.98 M allocations: 104.340 MiB, 2.96% gc time, 99.86% compilation time)
```

Julia v1.9 is quite slow at loading PyCall comparing to it used to be in v1.7. I wonder if there are any known reasons for the latency or possible approaches to address this?

MilesCranmer:
Hey @thautwarm,
I was wondering if this is still an issue on v1.10 with https://github.com/JuliaLang/julia/pull/49525 merged?
Cheers,
Miles

thautwarm:
It gets resolved since 1.10, but still applicable for earlier versions.

stevengj:
Closing as a Julia issue that is now fixed.

PR #1055, opened 2023-09-26
reduce method invalidations and latency by remove method of converting PyObject to nothing

songjhaha:
This PR is trying to solve the high latency when `using PyCall` in second time like https://github.com/JuliaPy/PyCall.jl/issues/1052. 

We found the method defination `Base.convert(::Type{Nothing}, ::MyType)` would create many method invalidations, which casues this latency (see https://github.com/JuliaLang/julia/issues/51389), so this PR remove the method `convert(::Type{Nothing}, ::PyObject)`, and some patches by @thautwarm solve the auto converting `convert(::Type{PyAny}, o::PyObject)`.

before this PR:
```julia
julia> @time using PyCall
  0.478829 seconds (255.49 k allocations: 18.250 MiB, 1.91% compilation time)

julia> @time using PyCall
  1.957530 seconds (1.29 M allocations: 90.161 MiB, 5.23% gc time, 99.97% compilation time: 100% of which was recompilation)
```

after this PR:
```julia
julia> @time using PyCall
  0.424794 seconds (261.29 k allocations: 18.610 MiB, 2.36% compilation time)

julia> @time using PyCall
  0.000566 seconds (309 allocations: 21.836 KiB)
```

This latency time is estimated in Julia1.9.3 and  win10 OS

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -61,7 +61,8 @@ convert(::Type{T}, po::PyObject) where {T<:Real} =
 convert(::Type{T}, po::PyObject) where T<:Complex =
     T(@pycheck ccall(@pysym(:PyComplex_AsCComplex), Complex{Cdouble}, (PyPtr,), po))
 
-convert(::Type{Nothing}, po::PyObject) = nothing
+# TODO: This method defination causes method invalidations and latency in Julia 1.9.3, https://github.com/JuliaLang/julia/issues/51389
+# convert(::Type{Nothing}, po::PyObject) = nothing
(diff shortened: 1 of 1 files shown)


MilesCranmer:
Hey @songjhaha,
Sorry this hasn't been reviewed yet. I just saw this PR while scrolling by and was wondering if this still has an impact on v1.10 or whether https://github.com/JuliaLang/julia/pull/49525 was enough to fix this issue?
Thanks!
Miles

PR #1059, opened 2023-10-30
A few fixes for julia master

yuyichao:
1. Fix use of `cconvert(Cwstring)`. This may count as a breakage from base julia but it's a bit of a grey area. It may be argued that the return type for `cconvert` isn't really guaranteed.
2. `Docs.doc` is defined in base but the methods are defined in `REPL`. With `REPL` moved out of the sysimg by default this needs to be loaded explicitly.


stevengj:
Thanks!

PR #1063, opened 2023-11-08
version stdlibs

stevengj:
this is now required for merging into the registry.  I'm not 100% sure that this works with Julia 1.4 (may only work with 1.6?) but we'll see what CI says.


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "PyCall"
 uuid = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
 authors = ["Steven G. Johnson <stevenj@mit.edu>", "Yichao Yu <yyc1992@gmail.com>", "Takafumi Arakaki <aka.tkf@gmail.com>", "Simon Kornblith <simon@simonster.com>", "Páll Haraldsson <Pall.Haraldsson@gmail.com>", "Jon Malmaud <malmaud@gmail.com>", "Jake Bolewski <jakebolewski@gmail.com>", "Keno Fischer <keno@alumni.harvard.edu>", "Joel Mason <jobba1@hotmail.com>", "Jameson Nash <vtjnash@gmail.com>", "The JuliaPy development team"]
-version = "1.96.2"
+version = "1.97.0"
```
stevengj, line 4:
```suggestion
version = "1.96.2"
```

simeonschaub:
Ah, nice! Was just looking into this. Thanks!

stevengj:
Looks like the package manager is able to resolve the new dependencies on Julia 1.4 and 1.5.  (It's failing on a Conda import of `pyzmq`, which maybe can be resolved by switching to the `six` package similar to https://github.com/JuliaPy/Conda.jl/pull/209.)

PR #1065, opened 2023-12-05
Fix the stdlib compat entries

### LINKED ISSUE #1064: PyCall 1.96.2 breaks tests at Imbalance.jl
I can reproduce the following CI fail on my mac under Julia 1.9.1 and julia 1.6.5 (errors are slightly different):

https://github.com/JuliaAI/Imbalance.jl/actions/runs/6949846973/job/18908791611

However, if I pin PyCall to 1.96.1 then tests pass locally on my machine.

DilumAluthge:
Fixes #1064 

Almost all stdlibs should just use a compat entry of `"< 0.0.0, 1"`. So e.g.

```toml
StdlibName = "< 0.0.1, 1"
```

The `SHA` stdlib is an exception, because it should be:
```toml
SHA = "< 0.0.1, 0.7, 1"
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -18,10 +18,10 @@ MacroTools = "0.4, 0.5"
 VersionParsing = "1.0"
 
 julia = "1.4"
-Dates = "1.4"
-Libdl = "1.4"
-LinearAlgebra = "1.4"
-Serialization = "1.4"
+Dates = "<0.0.1, 1"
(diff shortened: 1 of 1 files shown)


DilumAluthge:
cc: @ablaom

ablaom:
I can confirm that this fixes the issue at Imbalance.jl.

PR #1068, opened 2023-12-15
Rename `distutils.sysconfig` to `sysconfig`

MilesCranmer:
Required for Python 3.12 as distutils is removed.

@stevengj do you think you review and merge this when you get a chance? It's only a one-word change and is required for 3.12 compatibility in PyJulia.

cc @mkitti

Fixes https://github.com/JuliaPy/pyjulia/pull/538 (hopefully)

### DIFF
--- a/deps/buildutils.jl
+++ b/deps/buildutils.jl
@@ -5,7 +5,7 @@ import Conda, Libdl
 
 pyvar(python::AbstractString, mod::AbstractString, var::AbstractString) = chomp(read(pythonenv(`$python -c "import $mod; print($mod.$(var))"`), String))
 
-pyconfigvar(python::AbstractString, var::AbstractString) = pyvar(python, "distutils.sysconfig", "get_config_var('$(var)')")
+pyconfigvar(python::AbstractString, var::AbstractString) = pyvar(python, "sysconfig", "get_config_var('$(var)')")
 pyconfigvar(python, var, default) = let v = pyconfigvar(python, var)
     v == "None" ? default : v
 end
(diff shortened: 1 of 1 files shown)


stevengj:
Then it will break earlier versions?  Can you try both names in a try-catch block?

MilesCranmer:
This sysconfig module was introduced by itself in Python 3.2. distutils has basically been a symlink to it since 3.2, then deprecated in 3.10 and 3.11, and removed in 3.12. But sysconfig has been there since 3.2.

stevengj:
We still support older versions.

MilesCranmer:
Even before Python 3.2? I don’t think they are even supported from Python itself anymore:

> All official maintenance for Python 3.2, including security patches, has ended.

Python 2.7 is also retired and has no support from Python itself.

I could add a try catch if you want though?

MilesCranmer:
2.7 was retired in 2020 after a decade of maintenance patches:

> We have decided that January 1, 2020, was the day that we sunset Python 2. That means that we will not improve it anymore after that day, even if someone finds a security problem in it.

https://www.python.org/doc/sunset-python-2/

We dropped support for 2.7 in PyJulia because of this. Maybe could do the same in PyCall.jl?

MilesCranmer:
If someone really wants to use an unmaintained, 10+ old version of Python, with all of its security issues, I assume they would be technically skilled enough to also install an old version of PyCall that does support it, right?

And if they don’t have the technical knowledge to do that, I think it’s actually safer to throw an error about unsupported Python, than to let themselves put themselves at risk.

MilesCranmer:
@stevengj I realise this is likely something to discuss in a future PR though. So to help speed up this fix I have implemented a try-catch for earlier Python.

FYI the Python 2.7 CI failures look to be from 2.7 not being available.

stevengj:
Thanks, LGTM.  Indeed, I would prefer having a conversation about dropping older Python versions in a separate PR.

PR #1069, opened 2023-12-15
Add parsing for datetime.time

nateybear:
Using same patterns as elsewhere in this file (and avoiding timezones like in DateTime), add support for parsing times from the Python datetime package to Julia's Dates.Time struct. This came up as a need I had in a personal project and was a quick addition.


stevengj:
Looks good. Could you add a couple of tests?

nateybear:
> Looks good. Could you add a couple of tests?

I added a `rountripeq` test like for Date and DateTime

nateybear:
@stevengj bumping since I also forgot about this. Tests have been implemented, check failures seem unrelated to this change.

stevengj:
Looks good, thanks!

PR #1074, opened 2024-01-08
Thread-safe garbage collection v3 (ReentrantLock version)

MilesCranmer:
This is an alternative strategy to make the GC thread-safe compared with #1073 (#883).

@stevengj @mkitti @marius311 

My concern with #1073 is whether only allowing GC to run on a particular thread might cause an issue or lag in performing garbage collection. Instead this allows the GC to run on any thread but they must obtain a lock first. This is the logic:

```julia
const PYDECREF_LOCK = ReentrantLock()

function _pydecref_locked(po::PyObject)
    # If available, we lock and decref
    !islocked(PYDECREF_LOCK) &&
        trylock(() -> pydecref_(po), PYDECREF_LOCK) &&
        return nothing

    # Add back to queue to be decref'd later
    finalizer(_pydecref_locked, po)
    return nothing
end
```

---

One question I have is: does `islocked` work in a re-entrant context? (Or does it even matter, and I could just use a SpinLock?)

### DIFF
--- a/src/PyCall.jl
+++ b/src/PyCall.jl
@@ -81,12 +81,12 @@ mutable struct PyObject
     end
 end
 
-const PYDECREF_LOCK = ReentrantLock()
+const PYDECREF_PYOBJECT_LOCK = ReentrantLock()
 
 function pydecref_safe_(po::PyObject)
     # If available, we lock and decref
-    !islocked(PYDECREF_LOCK) &&
-        trylock(() -> (pydecref_unsafe_(po); true), PYDECREF_LOCK) &&
+    !islocked(PYDECREF_PYOBJECT_LOCK) &&
+        trylock(() -> (pydecref_unsafe_(po); true), PYDECREF_PYOBJECT_LOCK) &&
         return nothing
 
     # Add back to queue to be decref'd later
--- a/src/pybuffer.jl
+++ b/src/pybuffer.jl
@@ -32,9 +32,22 @@ mutable struct PyBuffer
         b = new(Py_buffer(C_NULL, PyPtr_NULL, 0, 0,
                           0, 0, C_NULL, C_NULL, C_NULL, C_NULL,
                           C_NULL, C_NULL, C_NULL))
-        finalizer(pydecref, b)
+        finalizer(pydecref_safe_, b)
         return b
     end
+end
+
+const PYDECREF_PYBUFFER_LOCK = ReentrantLock()
+
+function pydecref_safe_(b::PyBuffer)
+    # If available, we lock and decref
+    !islocked(PYDECREF_PYBUFFER_LOCK) &&
+        trylock(() -> (pydecref(b); true), PYDECREF_PYBUFFER_LOCK) &&
+        return nothing
+
+    # Add back to queue to be decref'd later
+    finalizer(pydecref_safe_, b)
+    return nothing
 end
 
 """


mkitti:
You might want to try a test-test-set pattern:
https://github.com/JuliaIO/HDF5.jl/blob/master/src/api/lock.jl

Basically `islocked` is cheap, so check that first. If it is not locked, then try to lock.

MilesCranmer:
Wait isn't that what this already does? Or you mean replacing the `lock` with `trylock`?

mkitti:
Yes, use `trylock`.

```
help?> trylock
search: trylock

  trylock(lock) -> Success (Boolean)

  Acquire the lock if it is available, and return true if successful. If the
  lock is already locked by a different task/thread, return false.

  Each successful trylock must be matched by an unlock.

  Function trylock combined with islocked can be used for writing the
  test-and-test-and-set or exponential backoff algorithms if it is supported
  by the typeof(lock) (read its documentation).
```

MilesCranmer:
Done!

MilesCranmer:
Do we need to make the `PyBuffer` deallocator thread-safe as well? Right now it's this:

```julia
mutable struct PyBuffer
    buf::Py_buffer
    PyBuffer() = begin
        b = new(Py_buffer(C_NULL, PyPtr_NULL, 0, 0,
                          0, 0, C_NULL, C_NULL, C_NULL, C_NULL,
                          C_NULL, C_NULL, C_NULL))
        finalizer(pydecref, b)
        return b
    end
end
```

which references the unsafe `pydecref`.

mkitti:
By the way, do you know what the threading advice actually is for the Python C API. Is it only one thread at a time, or only call from the same thread?

mkitti:
It might be time to drop Python 2.7 CI

MilesCranmer:
> It might be time to drop Python 2.7 CI

Done! #1075

MilesCranmer:
> By the way, do you know what the threading advice actually is for the Python C API. Is it only one thread at a time, or only call from the same thread?

I don't see anything related to multi-threading here https://docs.python.org/3/c-api/refcounting.html. Would they have a particular requirement?

mkitti:
https://docs.python.org/3/c-api/init.html#non-python-created-threads

> If you need to call Python code from these threads (often this will be part of a callback API provided by the aforementioned third-party library), you must first register these threads with the interpreter by creating a thread state data structure, then acquiring the GIL, and finally storing their thread state pointer, before you can start using the Python/C API. When you are done, you should reset the thread state pointer, release the GIL, and finally free the thread state data structure.
> 
> The [PyGILState_Ensure()](https://docs.python.org/3/c-api/init.html#c.PyGILState_Ensure) and [PyGILState_Release()](https://docs.python.org/3/c-api/init.html#c.PyGILState_Release) functions do all of the above automatically. The typical idiom for calling into Python from a C thread is:
> 
> PyGILState_STATE gstate;
> gstate = PyGILState_Ensure();
> 
> /* Perform Python actions here. */
> result = CallSomeFunction();
> /* evaluate result or handle exception */
> 
> /* Release the thread. No Python API allowed beyond this point. */
> PyGILState_Release(gstate);
> Note that the PyGILState_* functions assume there is only one global interpreter (created automatically by [Py_Initialize()](https://docs.python.org/3/c-api/init.html#c.Py_Initialize)). Python supports the creation of additional interpreters (using [Py_NewInterpreter()](https://docs.python.org/3/c-api/init.html#c.Py_NewInterpreter)), but mixing multiple interpreters and the PyGILState_* API is unsupported.

mkitti:
Ok, we're going to need a redesign here since Julia GC is now multithreaded. 

Now the question is do we really `PyGILState_STATE` per thread, or can we manage per `Task`.

MilesCranmer:
> Ok, we're going to need a redesign here since Julia GC is now multithreaded.
> 
> Now the question is do we really `PyGILState_STATE` per thread, or can we manage per `Task`.

I feel like it would be quite a bit of work to properly incorporate the Python GIL into Julia code. That means you would need to thread-lock every single part of PyCall.jl, no?

Thread-locking the garbage collection is the major issue IMO, but it also seems easier to solve (with just this PR)?

MilesCranmer:
Pinging this thread before my teaching starts (and I get buried in work) :)

mkitti:
@stevengj , have you had a chance to take a look?

PR #1075, opened 2024-01-08
Remove unavailable Python 2.7 from GitHub actions

MilesCranmer:
Python 2.7 is no longer available on any GitHub images: https://github.com/actions/python-versions/blob/main/versions-manifest.json, so this PR removes it from the GitHub actions tests.

### DIFF
--- a/.github/workflows/system.yml
+++ b/.github/workflows/system.yml
@@ -41,16 +41,6 @@ jobs:
             architecture: x64
             python-version: '3.x'
             julia-version: '1.4'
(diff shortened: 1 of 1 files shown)


MilesCranmer:
One of @stevengj @mkitti do you think you could also update the branch rules so these tests are no longer required?  Screenshot 2024-01-08 at 21 07 07

abhro:
This should probably be closed now that #1098 is merged, right?

